PR #83, opened 2021-11-04
chore: enable sonarcloud check

pbastia:


pbastia:
not needed since we're using CodeQL

PR #98, opened 2021-11-19
test: submodule in db style tests & add to gh workflow

dleard:


.github/actions/dev-env-setup/action.yml
```diff
@@ -37,3 +37,6 @@ runs:
           ~/.cache/Cypress
           ./app/node_modules
         key: ${{ runner.os }}-yarn-cache-${{ hashFiles('app/yarn.lock') }}
+    - name: Checkout Submodules
+      run: git submodule update --init
+      shell: bash
```
matthieu-foucault, line 42: should we do this only in the db style tests job?

matthieu-foucault (approved):

PR #111, opened 2021-11-25
feat: document upload round trip

AlexZorkin:


schema/deploy/tables/cif_attachment.sql
```diff
@@ -0,0 +1,42 @@
+-- Deploy cif:tables/cif_attachment to pg
+
+BEGIN;
+
+create table cif.cif_attachment
```
matthieu-foucault, line 5:
```suggestion
create table cif.attachment
```
I don't know if we need the `cif_` prefix on tables (other than `cif_user`), since it's already under a `cif` schema @pbastia @dleard what do we think? We'll want to be consistent here

schema/deploy/tables/cif_attachment.sql
```diff
@@ -0,0 +1,42 @@
...
+(
+  id integer primary key generated always as identity,
+  file uuid,
+  file_name varchar(1000),
+  s3_path varchar(1000)
```
matthieu-foucault, line 10: can the `s3_path` be inferred from the `uuid`, if so we don't need a column here. We can add a [computed column](https://www.graphile.org/postgraphile/computed-columns/) later to have an easily-accessible download link (but I think downloading should be a separate PR)

matthieu-foucault (commented):
Looks like your branch needs a rebase on `develop` as I see some changes from your previous PR in here

schema/deploy/tables/cif_attachment.sql
AlexZorkin: I think it makes sense to simplify. I have updated the table to be just attachment.

schema/deploy/tables/cif_attachment.sql
AlexZorkin: you are correct, I removed this field from the schema.

app/pages/admin/attachments.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <DefaultLayout session={query.session}>
+      <h2>Attachments List</h2>
+      <ul>
+        {query.allCifAttachments.edges.map(({ node }) => (
+          <li key={node.s3Path} onClick={() => console.log('downloading:', node.fileName)}>
+            {node.fileName}
+          </li>
+        ))}
+      </ul>
+        <input
```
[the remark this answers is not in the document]
AlexZorkin: updated the view to use the FilePicker component.

app/server/middleware/graphql/index.ts
```diff
@@ -52,4 +62,30 @@ const postgraphileMiddleware = () => {
...
+  const stream = createReadStream();
+
+  // Save tile to remote storage system
+  const { uuid } = await saveRemoteFile({stream, filename})
+
+  return uuid
+}
+
+async function saveRemoteFile({stream, filename}) {
+  const response = await fetch('http://localhost:8000/api/v1/attachments/upload', {
```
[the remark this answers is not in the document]
AlexZorkin: added

app/server/middleware/graphql/index.ts
```diff
@@ -52,4 +62,30 @@ const postgraphileMiddleware = () => {
...
+      method: "POST",
+      headers: {
+        'api-key': '_doNotUseThisApiKeyInProduction_',
```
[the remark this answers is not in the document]
AlexZorkin: added

app/lib/relay/client.ts
```diff
@@ -1,25 +1,37 @@
...
+    batchMiddleware({
+      batchUrl: async () => Promise.resolve("/graphql"),
+      batchTimeout: 10,
+      allowMutations: true,
+    }),
+    uploadMiddleware()
+  ])
 
-    const json = await response.text();
-    return JSON.parse(json, withHydrateDatetime);
```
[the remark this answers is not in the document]
AlexZorkin: Im not sure I understand this comment. Did you want to include these lines?

app/.postgraphilerc.js
```diff
@@ -3,10 +3,17 @@ module.exports = {
...
       "@graphile-contrib/pg-many-to-many",
+      "postgraphile-plugin-upload-field",
     ],
     graphileBuildOptions: {
       connectionFilterAllowNullInput: true,
       connectionFilterRelations: true,
+      uploadFieldDefinitions: [
+        {
+          match: ({ schema, table, column, tags }) =>
+            table === "attachment" && column == "file",
```
pbastia, line 14:
```suggestion
            table === "attachment" && column === "file",
```

app/server/middleware/graphql/index.ts
```diff
@@ -36,12 +42,48 @@ if (process.env.NODE_ENV === "production") {
...
 const postgraphileMiddleware = () => {
   return postgraphile(pgPool, process.env.DATABASE_SCHEMA || "cif", {
     ...postgraphileOptions,
     graphileBuildOptions: {
       connectionFilterAllowNullInput: true,
       connectionFilterRelations: true,
+      uploadFieldDefinitions: [
+        {
+          match: ({ table, column }) =>
+            table === "attachment" && column == "file",
```
pbastia, line 83:
```suggestion
            table === "attachment" && column === "file",
```

app/lib/relay/client.ts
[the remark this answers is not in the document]
matthieu-foucault: No we don't. I didn't really look at what these lines did when I first copied this code from the relay-nextjs example

dleard (approved):

pbastia (approved):

PR #112, opened 2021-11-25
Feat: create cif user upon login

pbastia:
Additional middleware calling the graphql endpoint to make sure the current user exists in the database


app/server/index.ts
```diff
@@ -46,6 +47,8 @@ app.prepare().then(async () => {
 
   server.use(await ssoMiddleware());
 
+  server.get("/auth-callback", createUserMiddleware("localhost", port));
```
pbastia, line 50: Refactoring idea: sso-express would export its configuration or default configuration, so we don't have to use a magic string here

matthieu-foucault (dismissed):
You can delete `app/mutations/User/createUserFromSession.ts` as it is now obsolete

matthieu-foucault (approved):

PR #117, opened 2021-11-26
fix: banner image should not use a `Link` to `/`

matthieu-foucault:
This will ensure that logged in users are redirected to the appropriate landing route

### DIFF
--- a/app/components/Layout/Navigation.tsx
+++ b/app/components/Layout/Navigation.tsx
@@ -1,4 +1,3 @@
-import Link from "next/link";
 import Image from "next/image";
 import Button from "@button-inc/bcgov-theme/Button";
 import { BaseNavigation } from "@button-inc/bcgov-theme/Navigation";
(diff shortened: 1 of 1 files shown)


app/components/Layout/Navigation.tsx
```diff
@@ -31,16 +30,19 @@ const Navigation: React.FC<Props> = ({
...
-                <Image
-                  src="/img/BCID_CleanBC_rev_tagline_colour.svg"
-                  alt="logo for Province of British Columbia CleanBC"
-                  height={50}
-                  width={300}
-                />
-              </a>
-            </Link>
+            {/*
+              We don't want a fron't end navigation here,
```
pbastia, line 34:
```suggestion
              We don't want a front end navigation here,
```

pbastia (changes requested):

pbastia (approved):

PR #144, opened 2021-12-03
Feat: select operator

dleard:


app/pages/_app.tsx
```diff
@@ -5,6 +5,7 @@ import { getClientEnvironment } from "../lib/relay/client";
 import "normalize.css";
 import BCGovTypography from "components/BCGovTypography";
 import SessionExpiryHandler from "components/SessionExpiryHandler";
+import "semantic-ui-css/semantic.min.css";
```
matthieu-foucault, line 8: that's a lot of CSS to import in every page for just one component

matthieu-foucault: Something like MUI's [Autocomplete](https://mui.com/components/autocomplete/) might be a better option if we want to bring in a single, customizable, component

pbastia: We might get away with just a few files: https://github.com/Semantic-Org/Semantic-UI-React/issues/2029#issuecomment-326455590

matthieu-foucault: I'm not super familiar with SemanticUI, but it looks much easier to make a theme for MUI than SemanticUI.
Also there's a [useAutocomplete hook](https://mui.com/components/autocomplete/#useautocomplete) if we want to not import any styles

dleard: I'll give MUI Autocomplete a look. I figured importing SemanticUI's CSS would just be a temporary solution so I could continue moving forward on developing this feature for now

matthieu-foucault: You can also start with a simple select, and we add a searchable one after

app/package.json
```diff
@@ -60,10 +60,14 @@
     "@bcgov-cas/sso-react": "^1.0.1",
     "@bcgov/bc-sans": "^1.0.1",
     "@button-inc/bcgov-theme": "^1.0.1",
+    "@emotion/react": "^11.7.0",
+    "@emotion/styled": "^11.6.0",
```
matthieu-foucault, line 64: where is this used? We already have two different styling libraries (`styled-jsx` and `styled-components`)

app/lib/theme/widgets/DisplayOnlyWidget.tsx
```diff
@@ -0,0 +1,18 @@
...
+const DisplayOnlyWidget: React.FC<WidgetProps> = ({ options }) => {
+  return (
+    <span className="paragraph-text">
+      <label>{options.title}</label>
+      <p>{options.text}</p>
+      <style jsx>{`
+        h3 {
+          font-size: 1.25rem;
+          margin-bottom: 0.5rem;
+        }
```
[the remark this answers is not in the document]
dleard: I'll remove it, copy paste from CIIP

app/package.json
dleard: They came with @mui I can remove them and see if it affects the Autocomplete

dleard: mui Autocomplete explodes without those dependencies

matthieu-foucault: I see emotion can be replaced by styled-components, but it's currently discouraged as there is an issue with styled-components, SSR and MUI: https://mui.com/guides/styled-engine/#how-to-switch-to-styled-components

matthieu-foucault (changes requested):

matthieu-foucault (approved):

Issue #69, opened 2021-10-28
As a project coordinator, I want to create a project, so that I can add high level project information into the system


LindsayMacfarlane:
#### Description:
As a project coordinator, I want to create a project, so that I can add high level project information into the system.
At this stage, we only want to add the minimal information to identify a project. We can add more fields (e.g. project type, naics code after that)

As per the wireframe below the project overview includes attributes that will be stored in the `project` table (e.g. rfp_id, operator_id, project summary, etc) and the `project_manager` table.

As we plan on using `react-jsonschema-form` to render form on the front-end, we will need to engineer a solution to link the front-end forms to the `form_change` table, taking into account that each record of the `form_change` table should result in a single record of a table being updated:

- a `project_revision` table should be added, and an optional foreign key to that table should be added to `form_change`
- committing a `project_revision` record means committing all of the `form_change` records underneath it
- the project creation page should be made of 3+ `react-jsonschema-form` components (one for each `form_change` record), and will update the pending changes directly (see #77).
- when clicking on the submit button (the discard button will be handled in #96), a mutation on the `project_revision` is called, affecting all of the `form_change`s

 

#### Acceptance Criteria:

Given I am on the project list page
When I click the 'create project' button
Then a new `project_revision` record is created, with `form_change` records for : `project` and `project_manager` tables [only one project manager to start]

Given I entered project data
When I click the "Submit" button
Then the project gets persisted to the database [project and project_manager tables]
  and the `project_revision` and  `form_change` records are marked as "committed"

#### Development Checklist:
- `project_revision` has:
    > a state
    > a generated ID (sequential)
    > foreign key to `project` table
    > timestamp/user columns
- `form_change` table has a foreign key to `project_revision` (allow null for now)
- there is a trigger on the project_revision table that propagates the commit to the individual `form_change` rows

On the project form
- [ ] Project Name
- [ ] Project Summary
- [ ] Project Manager selection [only 1]
- [ ] Add project button
- [ ] Meets the DOD

rajpersram:
Added a selection for Project Manager(s)

rajpersram:
![CIF Workflow Co-Design - Create or Edit Project - Overview tab.jpg](https://images.zenhubusercontent.com/5c2e4cfdab4f3464fb8df8d1/ece5f5b5-5f27-4ed1-b73e-1a503647af55)

rajpersram:
Sidebar and revision message:
[...]

PR #145, opened 2021-12-03
Feat: unique project identifier

pbastia:


matthieu-foucault (commented):
Partial review

app/components/Layout/FormBorder.tsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+
+interface Props {
+  title: string;
+}
+
+const FormBorder: React.FunctionComponent<Props> = (props) => {
+  return (
+    <>
+      <fieldset>
```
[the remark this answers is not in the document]
pbastia: We stick a lot closer to @rajpersram 's designs if we use the regular `<fieldset>`

schema/deploy/tables/project.sql
```diff
@@ -5,9 +5,9 @@ begin;
 
 create table cif.project(
   id integer primary key generated always as identity,
-  cif_identifier integer unique not null,
   description varchar(10000) not null,
-  funding_stream_id integer not null references cif.funding_stream(id)
+  funding_stream_id integer not null references cif.funding_stream(id),
+  unique_project_id varchar(1000) not null unique
```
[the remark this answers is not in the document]
pbastia: what about `rfp_number` ? this is what the business area calls it on internal documents

app/pages/cif/projects.tsx
```diff
@@ -81,14 +81,14 @@ export function Projects({ preloadedQuery }: RelayProps<{}, projectsQuery>) {
       <h1>Projects</h1>
       {query.allProjects.edges.length === 0 && <p>None</p>}
       {query.allProjects.edges.map(({ node }) => (
-        <Card title={node.cifIdentifier} key={node.id}>
+        <Card title={node.uniqueProjectId} key={node.id}>
```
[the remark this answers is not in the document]
pbastia: ...I never thought about this being a possibility! I'm not sure, I quite like knowing for sure if we're talking about Relay or database entities, although since we're describing database columns in the form data, that could become quite hairy to keep clean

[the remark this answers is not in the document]
dleard: I didn't think of that as a possibility either. Yes, I think we should turn it off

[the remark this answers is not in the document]
matthieu-foucault: Or, maybe better, we don't mess with postgraphile's name inflections, and use a camel-to-snake case conversion in postgres when applying the form changes, with something like: `regexp_replace(name, '([A-Z])','_\1', 'g'))`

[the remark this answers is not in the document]
dleard: wrapping that regex in a `lower()` would work

app/lib/theme/widgets/TextWidget.tsx
```diff
@@ -9,17 +9,27 @@ const TextWidget: React.FunctionComponent<WidgetProps> = ({
...
+      <Input
+        id={id}
+        onChange={(e) => onChange(e.target.value || undefined)}
+        placeholder={placeholder}
+        label={getRequiredLabel(label, required)}
+        value={value || ""}
+        size={(uiSchema && uiSchema["bcgov:size"]) || "medium"}
+        required={required}
+      ></Input>
+      <style jsx>
```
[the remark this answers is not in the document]
pbastia: The input styles have been overwritten in the `bcgov-theme` library, the styling is actually not exposed to us. Maybe something we can request as a feature?
For this specific one I opened an [issue](https://github.com/button-inc/service-development-toolkit/issues/350), it should actually fill the whole width by design

app/components/Project/FormBase.tsx
```diff
@@ -0,0 +1,66 @@
...
+const FormBase: React.FunctionComponent<Props> = ({
+  formData,
+  applyChangesFromComponent,
+  onFormErrors,
+  createInitialData,
+  schema,
+  uiSchema,
+}) => {
+  // We restrict the data to only the fields we care about
+  // It's important to default to `undefined` so we trigger the errors on rjsf
```
matthieu-foucault, line 22: I guess the second part is where [`omitExtraData`](https://react-jsonschema-form.readthedocs.io/en/latest/api-reference/form-props/#omitextradata) doesn't cut it?

app/pages/cif/create-project.tsx
```diff
@@ -62,24 +47,18 @@ export function CreateProject({
...
-
-  /* Uncomment this block when the applyChangeFromComponent function is ready to be used.
-   Currenly commented out so the linter won't complain about the unused function.
-
   // The applyChangeFromComponent function will require this page to be aware of the state of the newFormData object
   const formChangeData = query.formChange.newFormData;
 
   // A function to be called by individual components making changes to the overall form_change data
-  const applyChangeFromComponent = (changeObject: any)  => {
+  const applyChangesFromComponent = (changeObject: any) => {
```
matthieu-foucault, line 54: Is this needed if we have a 1-to-1 mapping between a jsonschema form and a form_change record? Or is that not the case?

app/components/Project/FormBase.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <FormBorder title="Background">
+      <Form
+        // @ts-ignore
+        ref={formRef}
+        schema={schema}
+        uiSchema={uiSchema}
+        formData={initialData}
+        onChange={(change) => {
+          applyChangesFromComponent(change.formData);
```
matthieu-foucault, line 57: does this callback mutate `change.formData`? From the name that's what it sounds like to me. If it's just a regular callback, perhaps this prop should be renamed to `onChange`?

app/components/Project/FormBase.tsx
pbastia: Yes, `omitExtraData` only works with `onSubmit`, not `onChange` unfortunately

app/pages/cif/create-project.tsx
matthieu-foucault: The idea put forward when doing the exploration was to have that one-to-one association between a form_change and a jsonschema form. I don't know how high fidelity those designs are, and whether we need to stick the same ordering of fields, especially if that means adding features in the code to support a different layout (unless that is an optimal layout that is supported by user research). If needed, I'd argue that the design needs to be updated to ensure that the form fields that will end up in the same form_change are located together

pbastia: I'm all up for that, happy to refactor - and remove code :) 
Since we maintain a theme for rjsf now, it'll be probably a lot easier to write a few custom object types and renderers to match the wireframes

dleard: Does this mean each `fieldset` in the wireframe is not a separate component & we will have one component that renders all those fields with one JsonSchemaForm?

pbastia: The <fieldset> can be part of the form schema (with an object template) and rendered by rjsf instead of wrapping an entire block

dleard: Gotcha. One "create-project" component, one form schema, many fieldsets rendered by rjsf

pbastia: one form schema per `form_change` database record, in this case, one for the project, one per contact, etc.
Everything will be wrapped in one revision when it's implemented as part of #69

matthieu-foucault (approved):

PR #147, opened 2021-12-08
chore: update pod resource limits

dleard:


app/.iyarc
```diff
@@ -0,0 +1,2 @@
+# Marking this as ignored until we update the major version of next
```
matthieu-foucault, line 1: a minor/patch update should do, we're already at v11

matthieu-foucault (approved):

PR #148, opened 2021-12-09
Feat/funding stream dropdown

AlexZorkin:
funding stream dropdown form added to the project create view


app/components/Project/ProjectFundingStreamForm.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { ProjectFundingStreamForm_query } from "__generated__/ProjectFundingStreamForm_query.graphql";
+import { graphql } from "react-relay/hooks";
+import { createFragmentContainer } from "react-relay";
+
+interface Props extends FormComponentProps {
+  query: ProjectFundingStreamForm_query;
+}
+
+const uiSchema = {
+  title: "Funding Stream",
```
matthieu-foucault, line 14: should this property be `ui:title`? Otherwise this implies that the form data has a `title` attribute

app/components/Form/FormBase.tsx
```diff
@@ -6,7 +6,7 @@ import FormComponentProps from "./FormComponentProps";
 
 interface Props extends FormComponentProps {
   schema: JSONSchema7;
-  uiSchema: object;
+  uiSchema: { title: string };
```
matthieu-foucault, line 9: There is a [UiSchema type exported](https://github.com/rjsf-team/react-jsonschema-form/blob/76ab08c5b06b1cbecd03fab0bdbca64b69b5c038/packages/core/index.d.ts#L67). Granted, it doesn't do much in terms of type checking

app/components/Project/ProjectFundingStreamForm.tsx
```diff
@@ -0,0 +1,65 @@
...
+  console.log(fundingStreamList);
+
+  var schema: JSONSchema7 = {
+    type: "object",
+    title: "Project Funding Stream",
+    properties: {
+      funding_stream_id: {
+        type: "string",
+        title: "Funding Stream ID",
+        enum: fundingStreamList,
```
matthieu-foucault, line 43: you may want an `enumNames` here. See the [ProductRowIdField](https://github.com/bcgov/cas-ciip-portal/blob/38e4c5778e1a0112c3eb6cb2def29dd6be67c947/app/containers/Forms/ProductRowIdField.tsx#L41) in CIIP as an example. Typescript might complain as `enumNames` is not part of `JSONSchema7`, so you might need a cast here.
The idea is that we want the id to be saved in the form, but you want to display the funding stream name

app/pages/cif/create-project.tsx
```diff
@@ -86,6 +88,16 @@ export function CreateProject({
             />
           </Grid.Col>
         </Grid.Row>
+        <Grid.Row>
+          <Grid.Col>
+            <ProjectFundingStreamForm
```
matthieu-foucault, line 93: Not sure if the funding stream selector needs to be a form in itself here, or just a form field that needs to be added to our theme

app/lib/theme/widgets/SelectWidget.tsx
```diff
@@ -0,0 +1,48 @@
...
+  title: string;
+  enum: number[];
+  value: number;
+}
+
+const SelectWidget: React.FunctionComponent<WidgetProps> = (props) => {
+  const { id, onChange, schema, placeholder, label, required, uiSchema } =
+    props;
+
+  const options = schema.anyOf as Array<Option>;
```
matthieu-foucault, line 16: if we're relying on `anyOf` being used rather than `enum`/`enumNames`, we should probably check that it exists and throw an meaningful error here

app/tests/unit/lib/theme/widgets/SelectWidget.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+        anyOf: [
+          { value: 1, enum: [1], type: "number", title: "Option 1" },
+          { value: 2, enum: [2], type: "number", title: "Option 2" },
+        ],
+      },
+    };
+
+    const componentUnderTest = render(<SelectWidget {...props} />);
+
+    expect(componentUnderTest.container).toMatchSnapshot();
```
matthieu-foucault, line 25: snapshot testing is a good smoke test, but we need an actual assertion here.

matthieu-foucault (changes requested):
A couple of minor improvements to make and that's good to go

app/components/Form/FormBase.tsx
```diff
@@ -39,10 +39,10 @@ const FormBase: React.FC<Props> = ({
       schema
     );
     onFormErrors(makeErrorsObject(errorSchema));
-  }, []);
+  }, [formData, makeErrorsObject, onFormErrors, formData]);
```
dleard, line 42: Adding these dependencies causes infinite re-renders. Despite the eslint warning, I think the empty array is correct. We want this useEffect() to essentially work like ComponentDidMount() and not depend on the makeErrorsObject / onFormErrors functions. 
https://stackoverflow.com/questions/58579426/in-useeffect-whats-the-difference-between-providing-no-dependency-array-and-an

matthieu-foucault: pretty sure we don't want errors on mount anyway, so this effect could be deleted?

app/lib/theme/widgets/SelectWidget.tsx
```diff
@@ -0,0 +1,48 @@
...
+        {getRequiredLabel(label, required)}
+      </label>
+      <Dropdown
+        id={id}
+        onChange={(e) => onChange(e.target.value || undefined)}
+        placeholder={placeholder}
+        size={(uiSchema && uiSchema["bcgov:size"]) || "large"}
+        required={required}
+      >
+        {options.map((opt) => {
```
dleard, line 29: Possibly a bug with the button-inc `Dropdown` component, but it does not include an empty option. It defaults to showing the first option in the options array in the dropdown. This means it will by _look_ like the first option (in this case EP), is selected by default where it is not. The placeholder will never be shown & "EP: will be displayed, but no description will show because it is not actually selected. I suggest adding an empty option above the mapping that shows the placeholder here to avoid confusion,
```suggestion
        <option key={-1} value={undefined}>{placeholder}</option>
        {options.map((opt) => {
```

dleard (changes requested):

matthieu-foucault:
Should a couple of funding streams be added to our dev data in `schema/data/dev` ?

app/components/Project/ProjectForm.tsx
```diff
@@ -37,9 +46,17 @@ const ProjectForm: React.FC<Props> = (props) => {
...
+  if (props.formData.fundingStreamId) {
+    selectedFundingStream = query.allFundingStreams.edges.find(
+      ({ node }) => node.rowId === props.formData.fundingStreamId
+    );
+  }
+
   const schema: JSONSchema7 = {
     type: "object",
-    required: ["rfpNumber", "description", "operatorId"],
+    required: ["rfpNumber", "description", "operatorId", "fundingStreamId"],
```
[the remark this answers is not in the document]
AlexZorkin: useMemo added around schema

app/components/Project/ProjectForm.tsx
```diff
@@ -92,6 +125,23 @@ const ProjectForm: React.FC<Props> = (props) => {
...
+      "ui:col-md": 12,
+      "ui:widget": "DisplayOnly",
+      "bcgov:size": "large",
+      "ui:options": {
+        text: `${
+          selectedFundingStream ? selectedFundingStream.node.description : ""
+        }`,
+        title: "Description",
+      },
+    },
```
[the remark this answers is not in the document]
AlexZorkin: useMemo added around uiSchema

app/lib/theme/widgets/SelectWidget.tsx
AlexZorkin: added a throw if anyOf is empty or undefined

app/tests/unit/lib/theme/widgets/SelectWidget.test.tsx
AlexZorkin: added a new test that tests the value of the actual dropdown

app/components/Form/FormBase.tsx
AlexZorkin: removed the error checking on form load so this is no longer an issue

app/lib/theme/widgets/SelectWidget.tsx
AlexZorkin: added a placeholder above

AlexZorkin:
> Should a couple of funding streams be added to our dev data in `schema/data/dev` ?

EP and IA rows are added in the funding_stream.sql deploy

dleard (approved):
:+1:

matthieu-foucault (approved):

PR #175, opened 2021-12-17
Feat: Funding Stream RFP

AlexZorkin:


schema/deploy/tables/funding_stream_rfp.sql
```diff
@@ -0,0 +1,44 @@
...
+
+comment on table cif.funding_stream_rfp is 'Table containing information about a CIF funding_stream_rfp';
+comment on column cif.funding_stream_rfp.id is 'Unique ID for the funding_stream_rfp';
+comment on column cif.funding_stream_rfp.year is 'The year this funding stream rfp is associated with';
+comment on column cif.funding_stream_rfp.funding_stream_id is 'The id of the funding_stream (cif.funding_stream.id) this rfp year is associated with';
+
+-- 2018 to 2022 is the MVP year range
+insert into cif.funding_stream_rfp (year, funding_stream_id) values
+(2018, 1), (2019, 1), (2020, 1), (2021, 1), (2022, 1),
+(2018, 2), (2019, 2), (2020, 2), (2021, 2), (2022, 2);
```
matthieu-foucault, line 42: I'm not sure about this. According to the [funded projects list](https://www2.gov.bc.ca/gov/content/environment/climate-change/industry/cleanbc-industry-fund/funded-projects), EP started in 2019. I think that IA started in 2021?

schema/deploy/tables/funding_stream_rfp.sql
```diff
@@ -0,0 +1,44 @@
...
+select cif_private.upsert_timestamp_columns('cif', 'funding_stream_rfp');
+
+do
+$grant$
+begin
+
+-- Grant cif_internal permissions
+perform cif_private.grant_permissions('select', 'funding_stream_rfp', 'cif_internal');
+perform cif_private.grant_permissions('insert', 'funding_stream_rfp', 'cif_internal');
+perform cif_private.grant_permissions('update', 'funding_stream_rfp', 'cif_internal');
```
matthieu-foucault, line 21: I think that only admins should be allowed to update this list?

schema/deploy/tables/project.sql
```diff
@@ -1,12 +1,14 @@
...
 -- requires: tables/funding_stream
+-- requires: tables/funding_stream_rfp
 
 begin;
 
 create table cif.project(
   id integer primary key generated always as identity,
   operator_id integer references cif.operator(id) not null,
   funding_stream_id integer not null references cif.funding_stream(id),
+  funding_stream_rfp integer not null references cif.funding_stream_rfp(id),
```
matthieu-foucault, line 11:
```suggestion
  funding_stream_rfp_id integer not null references cif.funding_stream_rfp(id),
```

matthieu-foucault: Do we need the `funding_stream_id` in this table if we have a foreign key to a `funding_stream_rfp`?

We might want to rethink the form component so that it allows you to the stream and rfp year, but only sets the  `fundingStreamRfpId` in the form.

matthieu-foucault (changes requested):

schema/deploy/tables/project.sql
matthieu-foucault: This still need a rename to `funding_stream_rfp_id` to be consistent with our naming conventions

app/components/Project/ProjectForm.tsx
```diff
@@ -99,38 +87,28 @@ const ProjectForm: React.FC<Props> = (props) => {
...
+            label: "Funding Stream"
+          },
+          child: {
+            entityType: "allFundingStreamRfps",
+            displayField: "year",
+            placeholder: "Select a Funding Stream RFP Year",
+            label: "Funding Stream RFP"
+          },
+          foreignKey: "fundingStreamId",
+          query: query,
```
matthieu-foucault, line 107: I'd rather keep our JSON schema a valid entity. This is the reason why we're using `anyOf` instead of `enumNames` for select components. Adding custom fields to our JSON schema will harm code readability (as one would assume that this is an actual JSON schema, but it isn't), and might prevent use of this schema with other libraries (e.g. if we wanted to so server-side validation of the change before committing)

app/components/Project/ProjectForm.tsx
```diff
@@ -159,31 +137,13 @@ const ProjectForm: React.FC<Props> = (props) => {
...
-          text: `${
-            selectedFundingStream ? selectedFundingStream.node.description : ""
-          }`,
-          title: "Description",
-        },
-      },
       fundingStreamRfpId: {
-        "ui:placeholder": "Select a Funding Stream RFP Year",
-        "ui:widget": "select",
+        "ui:widget": "SelectParentWidget",
```
matthieu-foucault, line 141: Instead of having a completely generic widget here and passing its options via the `schema`, id rather have a `SelectFundingStreamRfpWidget`.

The `SelectFundingStreamRfpWidget` should use `useFragment`, so that the relay fragment it uses are declared in the same file where they are used. Right now `eslint` should show you warnings about some parts of the relay query not being used in this file.

`SelectFundingStreamRfpWidget` can definitely use a generic `SelectParentWidget` component, and do things such as parsing the list of funding streams and RFPs, making a valid schema to use by `SelectParentWidget`.

The `query` that the fragment will need can be passed via the `formContext`.

app/components/Project/ProjectForm.tsx
```diff
@@ -99,38 +87,28 @@ const ProjectForm: React.FC<Props> = (props) => {
...
-            };
-          }),
+          parent: {
+            entityType: "allFundingStreams",
+            displayField: "name",
+            placeholder: "Select a Funding Stream",
+            label: "Funding Stream"
+          },
+          child: {
+            entityType: "allFundingStreamRfps",
```
matthieu-foucault, line 101: One problem here is that there is a relation between the parent and child (not all funding streams exist on a given year), but it is not expressed here

matthieu-foucault (changes requested):
Perhaps instead of having two `select` components with a dependency between them, we should have a single `select` displaying `Emission Performance - 2021`, `Innovation Accelerator - 2022`, etc?

app/lib/theme/widgets/SelectParentWidget.tsx
```diff
@@ -0,0 +1,109 @@
...
+  }, [query]);
+
+  let childOptions = query[child.entityType].edges.filter(
+    ({ node }) => {
+      // console.log("t0", node)
+      // console.log("t1", node[foreignKey])
+      // console.log("t2", selectedParentId)
+      return node[foreignKey] === selectedParentId
+    }
+  );
```
matthieu-foucault, line 51: Ah I see this is where the filtering is done for the child. We'll need a `useMemo` here

app/lib/theme/widgets/SelectRfpWidget.tsx
```diff
@@ -0,0 +1,81 @@
...
+      return {
+        rowId: rowId,
+        year: year,
+        fundingStreamId: fundingStreamId,
+      };
+    }),
+    displayField: "year",
+    placeholder: "Select a Funding Stream RFP Year",
+    label: "Funding Stream RFP",
+  } as EntitySchema;
```
matthieu-foucault, line 67: You'll want some `useMemo` hooks to avoid creating a new list of options at every render

app/lib/theme/widgets/SelectRfpWidget.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  return (
+    <>
```
matthieu-foucault, line 70: This react fragment is not needed

app/lib/theme/widgets/SelectParentWidget.tsx
```diff
@@ -0,0 +1,79 @@
...
+  ];
+  const [selectedParentId, setSelectedParentId] = useState(parentValue);
+
+  const onParentChange = (val) => {
+    setSelectedParentId(parseInt(val));
+  };
+
+  let childOptions = child.list.filter((opt) => {
+    return opt[foreignKey] === selectedParentId;
+  });
```
matthieu-foucault, line 21: use a `useMemo` hook here

app/lib/theme/widgets/SelectParentWidget.tsx
```diff
@@ -0,0 +1,79 @@
+import { WidgetProps } from "@rjsf/core";
+import Dropdown from "@button-inc/bcgov-theme/Dropdown";
+import getRequiredLabel from "../utils/getRequiredLabel";
+import { useState } from "react";
+
+const SelectParentWidget: React.FunctionComponent<WidgetProps> = (props) => {
+  const { id, onChange, required, uiSchema, value, parent, child, foreignKey } =
```
matthieu-foucault, line 7: you're going to need to extend `WidgetProps` here

app/lib/theme/widgets/SelectRfpWidget.tsx
```diff
@@ -0,0 +1,81 @@
+import { WidgetProps } from "@rjsf/core";
+import { graphql, useFragment } from "react-relay";
+import SelectParentWidget from "./SelectParentWidget";
+
+interface EntitySchema {
```
matthieu-foucault, line 5: This interface should be defined and exported in `SelectParentWidget`

app/lib/theme/widgets/SelectWidget.tsx
```diff
@@ -28,7 +28,7 @@ const SelectWidget: React.FunctionComponent<WidgetProps> = (props) => {
 
   return (
     <>
-      <label htmlFor={`funding-stream-dropdown-${props.id}`}>
+      <label htmlFor={`select-dropdown-${props.id}`}>
```
matthieu-foucault, line 31: should this just be `id` to match the `Dropdown` prop below?

schema/deploy/tables/project.sql
matthieu-foucault: :point_up:

schema/sqitch.plan
```diff
@@ -23,6 +23,7 @@ tables/cif_user 2021-10-21T16:10:32Z Matthieu Foucault <matthieu@button.is> # Cr
 trigger_functions/set_user_id 2021-10-21T16:10:32Z Matthieu Foucault <matthieu@button.is> # Create the cif_private.set_user_id trigger function
 mutations/create_user_from_session 2021-10-21T16:10:32Z Matthieu Foucault <matthieu@button.is> # Create the cif.create_user_from_session mutation function
 tables/funding_stream 2021-12-01T05:19:11Z Alex Zorkin <AizeeDev@BigThink> # adding funding_stream table to plan
+tables/funding_stream_rfp 2021-12-15T00:47:25Z Alex Zorkin <AizeeDev@BigThink> # add funding stream rfp table
```
matthieu-foucault, line 26: You might want to update the sqitch plan and your sqitch config to have an actual email address instead of `<AizeeDev@BigThink>`

app/tests/unit/components/Form/ProjectForm.test.tsx
```diff
@@ -87,6 +87,7 @@ describe("The Project Form", () => {
       description: "d",
       operatorId: 1,
       fundingStreamId: 1,
+      fundingStreamRfpId: 1,
```
matthieu-foucault, line 90: I don't think that you need `fundingStreamId` in any of these `formData` anymore, only `fundingStreamRfpId`

app/lib/theme/widgets/SelectRfpWidget.tsx
```diff
@@ -0,0 +1,81 @@
+import { WidgetProps } from "@rjsf/core";
```
pbastia, line 1: This is a specific widget for the application, not part of the theme (which will eventually be complete enough to be used in conjunction with the button service development toolkit in any project).
I think we should put it next to the ProjectForm file and add it to the list of widgets available directly to the form ([example](https://github.com/bcgov/cas-ciip-portal/blob/8eb4a75b695b1fb489d5acc51cb0b52d8e8eaa4c/app/containers/Facilities/AddFacilityModal.tsx#L98) in ciip)

app/lib/theme/widgets/SelectRfpWidget.tsx
```diff
@@ -0,0 +1,81 @@
...
+              }
+            }
+          }
+        }
+      }
+    `,
+    props.formContext.query
+  );
+
+  const parent = {
```
pbastia, line 41:
```suggestion
  const parent: EntitySchema = {
  ...
  };
```

matthieu-foucault (changes requested):
Looks good. 

- We might want some tests for `SelectParentWidget`, and perhaps for `SelectRfpWidget` (although those fragments are a bit of a pain to test)

- Some database tests need updating, as we now need a `fundingStreamRfpId` in the json data instead of `fundingStreamId`

app/lib/theme/widgets/SelectRfpWidget.tsx
```diff
@@ -0,0 +1,81 @@
...
+        name: name,
+        description: description,
+      };
+    }),
+    displayField: "name",
+    placeholder: "Select a Funding Stream",
+    label: "Funding Stream",
+  } as EntitySchema;
+
+  const child = {
```
pbastia, line 55: same
```suggestion
  const child: EntitySchema = { ... };
```

app/lib/theme/widgets/SelectParentWidget.tsx
```diff
@@ -0,0 +1,79 @@
+import { WidgetProps } from "@rjsf/core";
+import Dropdown from "@button-inc/bcgov-theme/Dropdown";
+import getRequiredLabel from "../utils/getRequiredLabel";
+import { useState } from "react";
+
+const SelectParentWidget: React.FunctionComponent<WidgetProps> = (props) => {
```
pbastia, line 6: You could create an interface that extends `WidgetProps` so you can have static typing on `parent`, `child` and `foreignKey` 

```typescript
interface Props extends WidgetProps {
  parent: EntitySchema,
  child: EntitySchema,
  foreignKey: string
}
```

pbastia (changes requested):

app/lib/theme/widgets/SelectRfpWidget.tsx
AlexZorkin: I moved the application specific component into the Form directory and added it to the 'widgets' object on the Form

matthieu-foucault (approved):
Could be merged as-is, the comments below are just a couple of minor code improvements

app/tests/unit/lib/theme/widgets/SelectParentWidget.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+  },
+  foreignKey: "fundingStreamId",
+};
+
+describe("The SelectParentWidget Widget", () => {
+  it("selects the first option from parent", async () => {
+    const selectParentWidget = render(<SelectParentWidget {...props} />);
+    const parentDropdown = selectParentWidget.container.querySelector(
+      "#select-parent-dropdown-test-id"
+    );
```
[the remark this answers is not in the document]
AlexZorkin: either way works. The id comes from the actual component so using querySelector gives you slightly more code coverage.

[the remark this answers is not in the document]
matthieu-foucault: I think it comes down to the philosophy behind `testing-library`. Their recommendation is to only test based on things that the user can see, i.e. in this case accessing the input via its label instead of the id.

matthieu-foucault (approved):

app/lib/theme/widgets/SelectParentWidget.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const { id, onChange, required, uiSchema, value, parent, child, foreignKey } =
+    props;
+
+  const parentValue = parent.list.find((opt) => opt.rowId == parseInt(value))?.[
+    foreignKey
+  ];
+  const [selectedParentId, setSelectedParentId] = useState(parentValue);
+
+  const onParentChange = (val) => {
+    setSelectedParentId(parseInt(val));
```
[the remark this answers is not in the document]
AlexZorkin: no effect it seems on the dropdown. If you have a better wrapping for it let me know.

matthieu-foucault (approved):

PR #185, opened 2021-12-22
Feat/pages refactor

matthieu-foucault:


app/pages/cif/projects.tsx
```diff
@@ -53,21 +54,13 @@ export function Projects({ preloadedQuery }: RelayProps<{}, projectsQuery>) {
...
       input: {},
     });
-    await router.push({
-      pathname: "/cif/create-project",
-      query: {
-        id: response.createProject.projectRevision.id,
-      },
-    });
+    await router.push(
+      getProjectRevisionPageRoute(response.createProject.projectRevision.id)
```
pbastia, line 58: do we need to handle the case where `createProject` fails?

pbastia (approved):
Just a minor comment, not 100% sure if it's necessary to handle it in this PR, so I'll approve it :)

app/pages/cif/projects.tsx
matthieu-foucault: Probably, but I'd like to switch to using the `useMutation` hook everywhere first

PR #186, opened 2021-12-23
Feat/resume draft project

matthieu-foucault:


app/tests/unit/pages/projects.test.tsx
```diff
@@ -3,56 +3,85 @@ import { Projects, ProjectsQuery } from "../../../pages/cif/projects";
...
+          },
+          pendingNewProjectRevision: {
+            id: "123",
+          },
+        };
+      },
+    });
+    renderProjects();
+
+    expect(screen.getByText(/resume project draft/i)).toBeInTheDocument();
```
pbastia, line 81: We probably want to make sure that the "Add a Project" text is not on the page, since we're not looking at snapshots

pbastia (changes requested):
just one small testing thing, looks great!

pbastia (approved):

PR #188, opened 2021-12-24
Landing page

matthieu-foucault:
- merged `/cif` and `/admin` in a single landing page


app/components/Dashboard.tsx
```diff
@@ -0,0 +1,159 @@
...
+          getProjectRevisionPageRoute(response.createProject.projectRevision.id)
+        );
+      },
+    });
+  }, [createProject, isProjectCreating, router]);
+
+  const isAdmin = useMemo(
+    () => ADMIN_ROLES.some((role) => session?.userGroups?.includes(role)),
+    [session?.userGroups]
+  );
```
pbastia, line 57: This feels like logic that could be refactored somewhere else

app/components/Dashboard.tsx
```diff
@@ -0,0 +1,159 @@
...
+          color: #1a5a96;
+          background: transparent;
+          border: none;
+          cursor: pointer;
+          text-decoration: underline;
+          text-align: left;
+        }
+        :global(button:hover) {
+          text-decoration: none;
+          color: blue;
```
pbastia, line 152: the BC design system uses #0056b3 for the hover color...but I couldn't find this actually written anywhere

pbastia (changes requested):

app/components/Dashboard.tsx
matthieu-foucault: This button is displayed as a link, so it's blue for the hover: https://developer.gov.bc.ca/Design-System/Links

pbastia (approved):

PR #194, opened 2022-01-06
feat: add projects table

matthieu-foucault:
Adds a simple table displaying the list of project, using a generic component. Pagination, filtering and ordering will be added in a subsequent PR


dleard (approved):
:shipit:

PR #199, opened 2022-01-06
Feat: deleted records are immutable

pbastia:
A trigger making sure that deleted records can't be updated.
It gets automatically added with the `deleted_at` column

### DIFF
--- a/schema/deploy/util_functions/upsert_timestamp_columns.sql
+++ b/schema/deploy/util_functions/upsert_timestamp_columns.sql
@@ -79,6 +79,22 @@ begin
         'comment on column ', table_schema_name, '.', table_name, '.deleted_at is ''deleted at timestamp'';'
       );
       execute(comment_string);
+
+      -- Adding the deleted_records_are_immutable trigger only with the deleted_at column
+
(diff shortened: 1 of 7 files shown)


schema/deploy/util_functions/upsert_timestamp_columns.sql
```diff
@@ -93,12 +94,28 @@ begin
     execute(trigger_string);
   end if;
 
+  if not exists (select *
```
dleard, line 97: Is it a worthwhile optimization to move the creation of this trigger inside the `if add delete = true...` block above on line 64? This way the trigger will not be created if the deleted_at column does not exist and the function will not called unnecessarily on every update for tables without a `deleted_at` column. It would also remove the need for the `if to_jsonb(old) ? 'deleted_at' then..` check in `deleted_records_are_immutable()`.

pbastia: I don't know how expensive a `to_jsonb(...) ? ...` operation is, but I like removing code 👌

dleard (approved):

PR #202, opened 2022-01-07
Feat: committed changes are immutable

pbastia:
Making sure that committed `project_revision` and `form_change` records can't be altered

### DIFF
--- a/schema/deploy/tables/form_change.sql
+++ b/schema/deploy/tables/form_change.sql
@@ -22,6 +22,11 @@ create trigger commit_form_change
     when (new.change_status = 'committed')
     execute procedure cif_private.commit_form_change();
 
+create trigger committed_changes_are_immutable
+    before update on cif.form_change
+    for each row
+    execute procedure cif_private.committed_changes_are_immutable();
+
 do
 $grant$
 begin
--- a/schema/deploy/tables/form_change.sql
+++ b/schema/deploy/tables/form_change.sql
@@ -16,16 +16,18 @@ create table cif.form_change (
 
 select cif_private.upsert_timestamp_columns('cif', 'form_change');
 
+-- We want the immutable trigger to run first to avoid doing unnecessary work
+create trigger _100_committed_changes_are_immutable
+    before update on cif.form_change
+    for each row
+    execute procedure cif_private.committed_changes_are_immutable();
+
 create trigger commit_form_change
-    after insert or update of change_status on cif.form_change
+    before insert or update of change_status on cif.form_change
     for each row
     when (new.change_status = 'committed')
     execute procedure cif_private.commit_form_change();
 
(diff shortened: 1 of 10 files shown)


schema/deploy/tables/form_change.sql
```diff
@@ -16,12 +16,19 @@ create table cif.form_change (
...
 
+-- We want the immutable trigger to run first to avoid doing unnecessary work
+create trigger _100_committed_changes_are_immutable
+    before update on cif.form_change
+    for each row
+    execute procedure cif_private.committed_changes_are_immutable();
+
 create trigger commit_form_change
-    after insert or update of change_status on cif.form_change
+    before insert or update of change_status on cif.form_change
```
pbastia, line 26: I couldn't remember why we wanted this to happen `after` the insert/update... ? It makes the flow simpler to have the triggers all run `before`

dleard: I think in this particular case we may want to leave this as an `after` trigger since it is updating a different table than the table the trigger is called on, though it looks like `before` is more efficient and should be our default in most other cases.

`Typically, row-level BEFORE triggers are used for checking or modifying the data that will be inserted or updated. For example, a BEFORE trigger might be used to insert the current time into a timestamp column, or to check that two elements of the row are consistent. Row-level AFTER triggers are most sensibly used to propagate the updates to other tables, or make consistency checks against other tables. The reason for this division of labor is that an AFTER trigger can be certain it is seeing the final value of the row, while a BEFORE trigger cannot; there might be other BEFORE triggers firing after it. If you have no specific reason to make a trigger BEFORE or AFTER, the BEFORE case is more efficient, since the information about the operation doesn't have to be saved until end of statement.`

https://www.postgresql.org/docs/current/trigger-definition.html

pbastia: I think `before` makes sense here too, otherwise we're bound to violate the exact rule we want to implement: the `project_revision` record needs to be updated with the newly created project_id (and we'd need some more logic to allow this specific case)
If we leave it as an `after` trigger only for the `form_change` record, the `project` record isn't created yet and the project_revision_id foreign key relation can't be satisfied.

dleard: hm, indeed. Ok, makes sense. Should we append something like `zz_` to the `commit_form_change` trigger to ensure it fires last in the `before` trigger execution order? To cover ourselves in the case we end up adding another `before` trigger later on that potentially makes changes to the record being committed?

dleard: Or am I over-engineering here? :P

pbastia: I think since this trigger is added manually we don't need to? I'd leave that decision to the developer who has to make it, if it ever happens

dleard (approved):

PR #218, opened 2022-01-13
Feat/project pagination filtering

matthieu-foucault:
This currently suffers from some "loading" text showing between renders, and will need to be updated in a future PR in conjunction with a feature for `relay-nextjs` allowing it to not render a `Suspense`, so that we can change the `fallback` to be the previous component when we're fetching the next query.


app/pages/cif/projects.tsx
```diff
@@ -20,7 +33,19 @@ export const ProjectsQuery = graphql`
...
+    allProjects(
+      first: $pageSize
+      offset: $offset
+      filter: {
+        projectName: { includesInsensitive: $projectName }
+        operatorByOperatorId: {
+          tradeName: { includesInsensitive: $operatorTradeName }
+        }
+        rfpNumber: { includesInsensitive: $rfpNumber }
+      }
```
dleard, line 45: Project Status column is not searchable
```suggestion
      filter: {
        projectName: { includesInsensitive: $projectName }
        operatorByOperatorId: {
          tradeName: { includesInsensitive: $operatorTradeName }
        }
        rfpNumber: { includesInsensitive: $rfpNumber }
        projectStatusByProjectStatusId: {
          name: { includesInsensitive: $status }
        }
      }
```

app/pages/cif/projects.tsx
```diff
@@ -9,9 +9,22 @@ import { useRouter } from "next/router";
...
 export const ProjectsQuery = graphql`
-  query projectsQuery {
+  query projectsQuery(
+    $projectName: String
+    $operatorTradeName: String
+    $rfpNumber: String
+    $offset: Int
+    $pageSize: Int
+    $orderBy: [ProjectsOrderBy!]
+  ) {
```
dleard, line 27: Will need to add the $status variable here to make the Project Status column searchable
```suggestion
  query projectsQuery(
    $projectName: String
    $operatorTradeName: String
    $rfpNumber: String
    $status: String
    $offset: Int
    $pageSize: Int
    $orderBy: [ProjectsOrderBy!]
  ) {
```

dleard (changes requested):

dleard (approved):

PR #220, opened 2022-01-13
Feat: cancel project changes

pbastia:


app/pages/cif/project-revision/[projectRevision]/index.tsx
```diff
@@ -163,7 +182,7 @@ export function ProjectRevision({
           >
             Submit
           </Button>
-          <Button size="medium" variant="secondary">
+          <Button size="medium" variant="secondary" onClick={discardRevision}>
```
Comment (matthieu-foucault) [inline app/pages/cif/project-revision/[projectRevision]/index.tsx:185]: These buttons should be disabled if either the submit or discard mutations are in flight. 

I wouldn't disable them if `updatingFormChange` is true though, as it would lead to some distracting flickering, with the buttons changing state as you type. There is no risk to leaving them enabled in that case as the order of mutations would be preserved.

app/mutations/useDiscardMutation.ts
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * const [discardMyEntityMutation, isInFlight] = useDiscardMutation("myEntity", mutation);
+ *
+ */
+
+export default function useDiscardMutation<
+  TMutation extends MutationParameters
+>(
+  relayNodeName: string,
+  mutation: GraphQLTaggedNode
```
matthieu-foucault, line 34: It's a bit annoying that you have to explicitely pass the `relayNodeName`, but unfortunately it's not available in the `mutation` object...

app/mutations/FormChange/discardFormChange.ts
```diff
@@ -0,0 +1,9 @@
+import { graphql } from "react-relay";
+
+export const mutation = graphql`
+  mutation discardFormChangeMutation($input: UpdateFormChangeInput!) {
```
matthieu-foucault, line 4: Should this be removed for now? We can leave it if we're sure it will be used, but I don't want to introduce dead code.

matthieu-foucault (dismissed):
Looks great! A couple of minor improvements possible, nothing substantial

app/mutations/useDiscardMutation.ts
pbastia: unfortunately yes :( alternatively, as a refactor for later, we could write our own object that contains that information, a `NamedGraphQLTaggedNode` that holds the name and the grapgql string?

app/mutations/FormChange/discardFormChange.ts
pbastia: It's part of the AC of [the card](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/96) so I left it there. It's not needed right now, but we'll most likely need it after today's discussion

app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx
```diff
@@ -267,4 +271,34 @@ describe("The Create Project page", () => {
...
+
+    jest
+      .spyOn(require("react-relay"), "useMutation")
+      .mockImplementation(() => [jest.fn(), true]);
+
+    render(
+      <RelayEnvironmentProvider environment={environment}>
+        <ProjectRevision
+          data-testid="2"
+          CSN={true}
```
Comment (matthieu-foucault) [inline app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:292]: FYI I figured out what that CSN prop is. It's supposed to be whether or now we're on the client, and will determine whether a `Suspense` is rendered or not

matthieu-foucault (approved):

PR #244, opened 2022-01-25
feat: operators list view with filtering

AlexZorkin:


matthieu-foucault (changes requested):
Looks good. Just needs

- a `prettier` run
- rebase once #243 is merged to fix the dependency audit
- [optional]: a cypress test for the operators page (see `app/cypress/integration/cif/projects.spec.js` for an example)

matthieu-foucault (approved):

PR #246, opened 2022-01-26
Feat: validation on both server and client

pbastia:
Single source of JSON schema for server- and client-side validation:

- [x] json-schema in .json form
- [x] server-side postgraphile plugin for validation
- [x] client-side rjsf feature to validate multiple forms on the click of one button
- [x] tests


app/data/jsonSchemaForm/projectManagerSchema.ts
```diff
@@ -0,0 +1,16 @@
...
+  $schema: "http://json-schema.org/draft-07/schema",
+  type: "object",
+  title: "Project Manager",
+  required: ["cifUserId"],
+  properties: {
+    cifUserId: {
+      type: "number",
+      title: "Project Manager",
+      default: undefined,
+      anyOf: undefined,
```
pbastia, line 11: This is where we need to include `anyOf`, otherwise rjsf's [deepEqual](https://github.com/rjsf-team/react-jsonschema-form/blob/8335ad1966c7e249225d9dce50766750785d8e77/packages/core/src/components/Form.js#L38) implementation trips and triggers a change event on form load - which in turn triggers an update mutation.
Ajv doesn't care when validating on the server side.

app/pages/cif/project-revision/[projectRevision]/index.tsx
```diff
@@ -89,14 +91,32 @@ export function ProjectRevision({
...
+  const triggerFormSelfValidation = (formObject: any): [] => {
+    formObject.onSubmit({
+      preventDefault: () => {},
+      persist: () => {},
+    });
 
-  const onFormErrors = (formId: string, incomingErrors: Array<any>) => {
-    setErrors({ ...errors, [formId]: incomingErrors });
+    // Effectively validating the form a second time to retrieve the errors
+    const validationResult = formObject.validate(formObject.state.formData);
```
Comment (pbastia) [inline app/pages/cif/project-revision/[projectRevision]/index.tsx:101]: Unfortunately rjsf's [onSubmit](https://github.com/rjsf-team/react-jsonschema-form/blob/8335ad1966c7e249225d9dce50766750785d8e77/packages/core/src/components/Form.js#L307) (which we need to trigger internal error display) doesn't return a set of errors, so we need to re-validate to grab them

Comment (matthieu-foucault) [reply app/pages/cif/project-revision/[projectRevision]/index.tsx:101]: I wish there was a better way to do this, other than using refs...
The only thing I can think of is having a single form, and having an adapter layer to merge the various form_change into a single form data, but that comes with its own complexity, and may not be better

schema/deploy/tables/form_change.sql
```diff
@@ -11,7 +11,9 @@ create table cif.form_change (
   form_data_record_id integer,
   project_revision_id integer references cif.project_revision(id),
   change_status varchar(1000) default 'pending' references cif.change_status,
-  change_reason varchar(10000)
+  change_reason varchar(10000),
+  json_schema_name varchar(1000),
```
matthieu-foucault, line 15:
```suggestion
  json_schema_name varchar(1000) not null,
```

matthieu-foucault (approved):

matthieu-foucault:
@pbastia just a couple of small optional improvements, and a question re. using forwardRef

app/pages/cif/project-revision/[projectRevision]/index.tsx
Comment (pbastia) [reply app/pages/cif/project-revision/[projectRevision]/index.tsx:101]: I was looking at this...rjsf has methods to merge schemas and forms, but that would be a large amount of work for something that we are likely to refactor at a later point

app/components/Form/FormBase.tsx
```diff
@@ -11,22 +11,23 @@ interface Props extends FormComponentProps {
 const FormBase: React.FC<Props> = ({
   formData,
   onChange,
-  onFormErrors,
   schema,
   uiSchema,
   formContext,
   widgets,
+  setRef,
```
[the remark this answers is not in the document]
pbastia: Awesome. Rjsf seems to like it a lot better and reports fewer errors on the dropdowns 👌

matthieu-foucault:
@pbastia I see some failing unit tests, looks good otherwise

matthieu-foucault (approved):

PR #249, opened 2022-01-29
Feat: add contacts to project form

pbastia:


app/pages/cif/project-revision/[projectRevision]/index.tsx
```diff
@@ -91,25 +96,15 @@ export function ProjectRevision({
...
   /**
    *  Function: approve staged change, trigger an insert on the project
    *  table & redirect to the project page
    */
   const commitProject = async () => {
     const errors = [
-      ...triggerFormSelfValidation(projectFormRef.current),
-      ...triggerFormSelfValidation(projectManagerFormRef.current),
+      ...validateFormWithErrors(projectFormRef.current),
+      ...validateFormWithErrors(projectManagerFormRef.current),
```
Comment (pbastia) [inline app/pages/cif/project-revision/[projectRevision]/index.tsx:106]: @AlexZorkin this is what I had in mind for refactoring, making these form also self-validating and implementing the `ISupportExternalValidation` interface

app/lib/theme/EmptyObjectFieldTemplate.tsx
```diff
@@ -0,0 +1,9 @@
+import { ObjectFieldTemplateProps } from "@rjsf/core";
+
+const EmptyObjectFieldTemplate: React.FC<ObjectFieldTemplateProps> = (
```
matthieu-foucault, line 3: Not quite sure I understand what this template does, it might need some comments

app/lib/theme/utils/generateLabelComponent.tsx
```diff
@@ -0,0 +1,18 @@
+import { UiSchema } from "@rjsf/core";
+import getRequiredLabel from "./getRequiredLabel";
+
+export const generateLabelComponent = (
```
matthieu-foucault, line 4: should this be a react component?

schema/deploy/mutations/add_contact_to_revision.sql
```diff
@@ -0,0 +1,35 @@
...
+      (select form_data_record_id from cif.form_change where form_data_schema_name='cif' and form_data_table_name='project' and project_revision_id=$1),
+      $2
+    )::jsonb,
+    'INSERT',
+    'cif',
+    'project_contact',
+    nextval(pg_get_serial_sequence('cif.project_contact', 'id')),
+    $1,
+    'pending',
+    'Creating new project: secondary project_contact record',
```
matthieu-foucault, line 30: given that we have change reasons on the `project_revision`, I'm wondering if we need to have that column mandatory on the `form_change` table

app/components/Form/ProjectContactForm.tsx
```diff
@@ -0,0 +1,302 @@
...
+        },
+        connections: [
+          query.projectRevision.formChangesByProjectRevisionId.__id,
+        ],
+      },
+    });
+  };
+
+  const [discardFormChange] = useMutation(discardFormChangeMutation);
+  const deleteContact = (formChangeId: string) => {
```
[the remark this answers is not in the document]
pbastia: isn't this out of scope for this card? This is only including the project creation, so all the form_change records will be new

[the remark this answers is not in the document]
matthieu-foucault: Let's add that as an acceptance criteria of the project amendment user story then

app/lib/theme/utils/generateLabelComponent.tsx
pbastia: yes

app/components/Form/FormComponentProps.tsx
```diff
@@ -1,6 +1,8 @@
-import type { AjvError } from "@rjsf/core";
+import type { AjvError, ObjectFieldTemplateProps } from "@rjsf/core";
 export default interface FormComponentProps {
```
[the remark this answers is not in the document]
pbastia: rjsf enforces `schema` to be there, so we can't use the `FormProps` as a base...this small interface is just here to group what our different forms might need since they implement rjsf in turn

schema/deploy/mutations/add_contact_to_revision.sql
pbastia: I'm not sure...But because we keep that model for contacts or operators (form_change records without revision) it might be helpful to enforce a trace of where each form_change came from and in which context it was created

app/components/Form/ProjectContactForm.tsx
```diff
@@ -0,0 +1,278 @@
...
+        <Grid.Row>
+          <Grid.Col span={10}>
+            <FormBorder title="Project Contacts">
+              <Grid.Row>
+                <Grid.Col span={10} className="right-aligned-column">
+                  <Button>View Contact List</Button>
+                </Grid.Col>
+              </Grid.Row>
+              <Grid.Row>
+                <label>Primary Contact</label>
```
matthieu-foucault, line 184: needs an `htmlFor`

app/lib/theme/widgets/FieldLabel.tsx
```diff
@@ -0,0 +1,27 @@
...
+  uiSchema,
+}) => {
+  if (
+    uiSchema &&
+    uiSchema["ui:options"] &&
+    uiSchema["ui:options"].label === false
+  ) {
+    return null;
+  }
+  return <label htmlFor={htmlFor}>{getRequiredLabel(label, required)}</label>;
```
matthieu-foucault, line 24: should forward all of the other props to `label`, not just `htmlFor`

app/lib/theme/widgets/SearchDropdownWidget.tsx
```diff
@@ -3,17 +3,18 @@ import { WidgetProps } from "@rjsf/core";
...
+  const { onChange, schema, placeholder, readonly, label, required, uiSchema } =
+    props;
 
   const handleChange = (e: React.ChangeEvent<{}>, option: any) => {
     onChange(option?.value);
   };
 
   const getSelected = useCallback(() => {
-    if (props.value === null || props.value === undefined) return undefined;
+    if (props.value === null || props.value === undefined) return "";
```
matthieu-foucault, line 17: not sure about this. An empty string will not trigger a `not null` constraint

matthieu-foucault (approved):
There's a few things that we'll want to discuss, and the styles can be improved, but since this needs to be merged to avoid further conflicts let's merge this now

PR #250, opened 2022-01-31
feat: `commit_form_change` can generate `form_data_record_id`

matthieu-foucault:
- upon commit of a `form_change`, `form_data_record_id` will be added to the record if not specified
- added some `not null` constraints to the `form_change` table


schema/deploy/trigger_functions/commit_form_change.sql
```diff
@@ -38,16 +35,28 @@ begin
   vals := (select array_to_string(array(select quote_nullable(value) from jsonb_each_text(new.new_form_data)), ','));
 
   if new.operation = 'INSERT' then
```
pbastia, line 37: Do we need to add support for lowercase? It's the team's coding style, and I see we have it in a few tests below too

matthieu-foucault: the `operation` column should really be an enum or foreign key to a table, not a varchar, but that's a different PR

schema/test/unit/trigger_functions/commit_form_change_test.sql
```diff
@@ -163,6 +163,29 @@ select lives_ok(
...
+select is(
+  (select count(*) from mock_schema.mock_table where text_col='test text'),
+  1::bigint,
+  'A record should be inserted when a pending change is committed'
+);
+
+select ok(
+  ((select form_data_record_id from mock_schema.mock_form_change) is not null),
+  'form_data_record_id should be set when committing a change without specifying the record id'
+);
```
pbastia, line 188: should we check that the `form_data_record_id` is set to the current sequence value? Or maybe just to the ID that was just inserted? I guess they'll just be `1` all around for the sake of this test

```suggestion
select is(
  (select form_data_record_id from mock_schema.mock_form_change),
  (select currval(pg_get_serial_sequence('mock_schema.mock_table', 'id'))),
  'form_data_record_id should be set when committing a change without specifying the record id'
);
```

pbastia (approved):
Just a small thing about a test, not necessary

schema/test/unit/trigger_functions/commit_form_change_test.sql
matthieu-foucault: I can change that. I think it's 4 actually, because other records were inserted in the previous tests

pbastia (approved):

PR #252, opened 2022-02-02
Create/Edit contact form

matthieu-foucault:


app/pages/cif/contact/form/[contactForm].tsx
```diff
@@ -0,0 +1,127 @@
...
+      ...DefaultLayout_session
+    }
+    formChange(id: $contactForm) {
+      id
+      newFormData
+      formDataRecordId
+      updatedAt
+    }
+  }
+`;
```
Comment (pbastia) [inline app/pages/cif/contact/form/[contactForm].tsx:26]: This component can probably be made 100% generic in the future

schema/deploy/computed_columns/contact_pending_form_change.sql
```diff
@@ -0,0 +1,15 @@
...
+begin;
+
+create function cif.contact_pending_form_change(cif.contact) returns cif.form_change as $$
+  select * from cif.form_change
+  where form_data_schema_name = 'cif'
+  and form_data_table_name = 'contact'
+  and form_data_record_id = $1.id
+  and change_status = 'pending';
+$$ language sql stable;
+
```
pbastia, line 12: this will probably blow up once we allow concurrent editing, we'll have to filter by user in session

schema/deploy/mutations/create_form_change.sql
```diff
@@ -0,0 +1,56 @@
...
+    coalesce(new_form_data, (
+      select new_form_data
+      from cif.form_change
+      where form_data_record_id = create_form_change.form_data_record_id
+      and form_data_schema_name = create_form_change.form_data_schema_name
+      and form_data_table_name = create_form_change.form_data_table_name
+      and change_status = 'committed'
+      order by updated_at desc, id desc
+      limit 1
+    )),
```
pbastia, line 38: Do we want to merge the found `new_form_data` in the previous form_change record, with what was passed in? In case there are hidden fields like a project_id that were only inserted in the json on the form_change creation, like [here](https://github.com/bcgov/cas-cif/blob/43b25015d77b9fa8dacc549807a3d900b4c2a0dc/schema/deploy/mutations/create_project.sql#L46).
I'm thinking something like
```psql
coalesce((select ...)::jsonb, {}::jsonb) || coalesce(new_form_data::jsonb, {}::jsonb)
```

schema/verify/mutations/create_form_change.sql
```diff
@@ -0,0 +1,14 @@
...
+begin;
+
+select pg_get_functiondef(
+  (
+    select
+        oid FROM pg_proc
+    where
+        proname = 'create_form_change'
+  )
+);
```
pbastia, line 12: Oooh I like this way of verifying instead of passing a string 👌

pbastia (approved):

app/data/jsonSchemaForm/contactSchema.ts
```diff
@@ -0,0 +1,37 @@
...
+      type: "string",
+      title: "Family Name",
+    },
+    email: {
+      type: "string",
+      title: "Email",
+    },
+    phone: {
+      type: "string",
+      title: "Phone",
```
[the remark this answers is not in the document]
matthieu-foucault: We want server side validation, yes. Will add in a subsequent PR

schema/deploy/mutations/create_form_change.sql
matthieu-foucault: something like the project id would be copied over, so I'm not sure what we'd need to merge here

PR #254, opened 2022-02-02
feat: form_change and project_revision can be deleted

matthieu-foucault:
- [x] rename `deleted_by` to `archived_by`
- [x] rename `deleted_at` to `archived_at`
- [x] `form_change` table should not have `archived_at`, `archived_by` column
- [x] `project_revision` table should not have `archived_at`, `archived_by` column
- [x] pending `form_change` and `project_revision` records should be deleted when discarded
- [x] `form_change.operation` should be a `form_change_operation` enum, allowing `create`, `update`, `archive`


schema/deploy/util_functions/upsert_timestamp_columns.sql
```diff
@@ -116,21 +116,21 @@ $$ language plpgsql;
...
   );
   select cif_private.upsert_timestamp_columns(
-  table_schema_name := 'some_schema',
-  table_name := 'some_table',
-  add_create := true,
-  add_update := true,
-  add_delete := true);
+  table_schema_name => 'some_schema',
+  table_name => 'some_table',
+  add_create => true,
```
dleard, line 131: Just curiosity, what is the difference with the arrow here as opposed to `:=`?

dleard (approved):
LGTM

schema/deploy/util_functions/upsert_timestamp_columns.sql
matthieu-foucault: None, `:=` is just an older syntax: https://www.postgresql.org/docs/14/sql-syntax-calling-funcs.html

PR #267, opened 2022-02-11
Fix post-auth race condition

matthieu-foucault:


app/server/middleware/createUser.ts
```diff
@@ -1,8 +1,9 @@
-import type { Request, Response, NextFunction } from "express";
+import type { Request } from "express";
+import { getUserGroups } from "../helpers/userGroupAuthentication";
+import { performQuery } from "./graphql";
```
pbastia, line 3: We typically haven't used the syntax `folder_name/index.ts` to refer to exports in modules across our codebase, is it something we'd like to start?

pbastia (approved):
one very minor code style comment, looks great!

PR #269, opened 2022-02-11
Feat: import operators

dleard:
Uses postgres_fdw to import operator data from ggircs database in a separate openshift namespace

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ drop_foreign_test_db:
 	@$(PSQL) -d postgres -tc "SELECT count(*) FROM pg_database WHERE datname = 'foreign_test_db'" | \
 		grep -q 0 || \
 		$(PSQL) -d postgres -c "DROP DATABASE foreign_test_db" && \
-		$(PSQL) -d postgres -c "DROP USER foreign_user";
+		$(PSQL) -d postgres -c "DROP USER IF EXISTS foreign_user";
 
 .PHONY: create_foreign_test_db
 create_foreign_test_db: ## Ensure that the $(DB_NAME)_test database exists
--- a/Makefile
+++ b/Makefile
@@ -209,16 +209,22 @@ lint_chart:
 install: ## Installs the helm chart on the OpenShift cluster
 install: GIT_SHA1=$(shell git rev-parse HEAD)
 install: NAMESPACE=$(CIF_NAMESPACE_PREFIX)-$(ENVIRONMENT)
+install: GGIRCS_NAMESPACE=$(GGIRCS_NAMESPACE_PREFIX)-$(ENVIRONMENT)
 install: CHART_DIR=./chart/cas-cif
 install: CHART_INSTANCE=cas-cif
 install: HELM_OPTS=--atomic --wait-for-jobs --timeout 2400s --namespace $(NAMESPACE) \
 										--set defaultImageTag=$(GIT_SHA1) \
 	  								--set download-dags.dagConfiguration="$$dagConfig" \
+										--set ggircs.namespace=$(GGIRCS_NAMESPACE) \
 										--values $(CHART_DIR)/values-$(ENVIRONMENT).yaml
 install:
 	@set -euo pipefail; \
 	if [ -z '$(CIF_NAMESPACE_PREFIX)' ]; then \
 		echo "CIF_NAMESPACE_PREFIX is not set"; \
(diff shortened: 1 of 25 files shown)


Makefile
```diff
@@ -103,6 +103,22 @@ create_test_db:
 		$(PSQL) -d postgres -c "CREATE DATABASE $(DB_NAME)_test" &&\
 		$(PSQL) -d $(DB_NAME)_test -c "create extension if not exists pgtap";
 
+.PHONY: drop_foregin_test_db
```
matthieu-foucault, line 106:
```suggestion
.PHONY: drop_foreign_test_db
```

chart/cas-cif/templates/secret.yaml
```diff
@@ -10,6 +13,15 @@
 {{- $dbOwnerPassword = index $secret.data "database-owner-password" }} {{/* # pragma: allowlist secret */}}
 {{- end -}}
 
+{{- $ggircsSecret := (lookup "v1" "Secret" "9212c9-dev" "cas-ggircs" ) }}
```
matthieu-foucault, line 16: "9212c9-dev" should not be hardcoded here

chart/cas-cif/templates/secret.yaml
```diff
@@ -10,6 +13,15 @@
...
+{{- if $ggircsSecret }}
+{{- $dbReadonlyUser = index $ggircsSecret.data "database-readonly-user" }} {{/* # pragma: allowlist secret */}}
```
matthieu-foucault, line 18: I believe all of these #pragma directives can be removed now that we're not using detect-secrets anymore

chart/cas-cif/values-dev.yaml
```diff
@@ -38,3 +38,8 @@ deploy-db:
 
 download-dags:
   airflowEndpoint: https://cas-airflow-dev.apps.silver.devops.gov.bc.ca
+
+ggircs:
+  namespace: 9212c9-dev
+  prefix: 9212c9
+  environment: dev
```
matthieu-foucault, line 45: this shouldn't be needed here

chart/cas-cif/values.yaml
```diff
@@ -98,3 +98,10 @@ download-dags:
   dagId: fetch_and_save_dag_from_github
   helm:
     hook: "pre-install,pre-upgrade"
+
+ggircs:
+  namespace: ~
+  prefix: ~
+  environment: ~
```
matthieu-foucault, line 105: you shouldn't need all three variables here

schema/data/test_setup/external_database_setup.sql
```diff
@@ -0,0 +1,40 @@
+do $$
+  begin
+    RAISE NOTICE '*** BEGIN EXTERNAL TEST DATABASE SETUP ***';
+  end
+$$;
```
matthieu-foucault, line 5: Is this needed? maybe use `debug` instead of `notice` to avoid making our tests output too verbose?

matthieu-foucault (changes requested):

chart/cas-cif/values-dev.yaml
dleard: I'll need namespace, but not the prefix or environment

schema/data/test_setup/external_database_setup.sql
dleard: I was using it when configuring the makefile to set up the foreign database & left it in, but it's not really needed. I'll just remove it.

chart/cas-cif/values-dev.yaml
```diff
@@ -38,3 +38,6 @@ deploy-db:
 
 download-dags:
   airflowEndpoint: https://cas-airflow-dev.apps.silver.devops.gov.bc.ca
+
+ggircs:
+  namespace: 9212c9-dev
```
matthieu-foucault, line 43: this doesn't need to be hardcoded: it can be set in the makefile, using something like:
```
install: GGIRCS_NAMESPACE=$(CIF_NAMESPACE_PREFIX)-$(ENVIRONMENT)
```
and setting ggircs.namespace in the `HELM_OPTS` variable

matthieu-foucault (changes requested):
Just a small change an this looks good to go

matthieu-foucault (approved):

PR #280, opened 2022-02-16
feat: project amendment

matthieu-foucault:
Most of the work was related to the Contacts, and ensuring that the forms are handled and displayed properly when deleting contacts


dleard (approved):
Looks good! Nice reusable discard hook.

We may want to let anyone who is testing things on dev that the projects created in the dev data are not editable. Since they are just inserts on the project table they have no form_change data.

We probably want to look into creating the dev data properly with form_change records at some point so that we don't get falsy error reports once we get more into user testing.

app/cypress/integration/cif/project-revision/index.spec.js
```diff
@@ -64,4 +64,93 @@ describe("the new project page", () => {
     // Renders the default error message for a required field
     cy.get(".error-detail").last().should("contain", "Please enter a value");
   });
+
```
dleard, line 67: Line 26 of this file just needs a selector fix:
cy.get('#primaryContactForm_contactId').click()

dleard (changes requested):

dleard (approved):

app/mutations/FormChange/updateFormChange.ts
```diff
@@ -13,6 +13,11 @@ const mutation = graphql`
       formChange {
         id
         newFormData
+        operation
+        projectRevisionByProjectRevisionId {
+          ...ProjectContactForm_projectRevision
+          ...ProjectManagerFormGroup_revision
```
pbastia, line 19: This mutation is also used by the contact and operator forms...they don't need any of these fragments in the return type. Should we create a separate `updateFormChangeMutation` with different return queries, depending on the use case?

PR #295, opened 2022-02-23
Form changes computed columns support multiple revisions

matthieu-foucault:


matthieu-foucault:
This approach to `project_revision` records only containing a delta of the `form_change` that are modified is leading to greater code complexity, which will harm this project's maintainability. I'll start a new branch where the project_revision will have a complete snapshot of a project's `form_change` records.

PR #297, opened 2022-02-24
Feat: import operators with history

pbastia:
Using the `form_change` mechanism to insert or update operators from swrs


schema/data/test_setup/external_database_setup.sql
```diff
@@ -26,3 +26,4 @@ values
 
 grant usage on schema swrs to foreign_user;
 grant select on all tables in schema swrs to foreign_user;
+grant insert on all tables in schema swrs to foreign_user;
```
dleard, line 29: is this grant necessary? We aren't inserting anything in the swrs schema are we?

dleard (changes requested):
just one grant question

schema/data/test_setup/external_database_setup.sql
pbastia: We are in the test, to test the update of a SWRS name

dleard (approved):

PR #298, opened 2022-02-25
Chore: simplified managers computed column

dleard:
simplify the computed column to reflect the pattern - form changes are copied across revisions

I have temporarily hardcoded a projectManagerLabelId value in create_project() to unblock project creation. This code block will be removed in the front-end focused part of this PR since creat_project() should not create a form_change for managers by default.

### DIFF
--- a/schema/deploy/computed_columns/project_revision_project_manager_form_changes_by_label.sql
+++ b/schema/deploy/computed_columns/project_revision_project_manager_form_changes_by_label.sql
@@ -9,26 +9,14 @@ returns setof cif.manager_form_changes_by_label_composite_return
 as
 $computed_column$
 
-  with project_form_change_history as (
-    select fc.*
-      from cif.form_change fc
-      join cif.project_manager_label pml on cast(new_form_data->>'projectManagerLabelId' as integer) = pml.id
-        and (fc.updated_at, fc.id) = (select max(updated_at), max(id) from cif.form_change where cast(new_form_data->>'projectManagerLabelId' as integer) = pml.id)
-      where project_revision_id = $1.id
-        and form_data_schema_name='cif'
-        and form_data_table_name='project_manager'
-    union
-    select fc.*
-      from cif.form_change fc
-      join cif.project_manager pm on
-        fc.form_data_record_id = pm.id
-        and pm.project_id = $1.project_id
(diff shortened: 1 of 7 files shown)


schema/deploy/computed_columns/project_revision_project_manager_form_changes_by_label.sql
```diff
@@ -9,26 +9,14 @@ returns setof cif.manager_form_changes_by_label_composite_return
...
-      from cif.form_change fc
-      join cif.project_manager pm on
-        fc.form_data_record_id = pm.id
-        and pm.project_id = $1.project_id
-        and (fc.updated_at, fc.id) = (select max(updated_at), max(id) from cif.form_change where form_data_record_id = pm.id)
-  )
-  select label, row(project_form_change_history.*) as form_change
-    from project_form_change_history
+  with latest_changes as (
+    select distinct on (new_form_data->'projectManagerLabelId') * from cif.form_change
```
matthieu-foucault, line 13: The same `projectManagerLabelId` shouldn't appear more than once for a given `project_revision_id`, so the `distinct on` isn't needed.

schema/test/unit/computed_columns/project_revision_project_manager_form_changes_by_label_test.sql
```diff
@@ -1,7 +1,7 @@
 begin;
 SET client_min_messages TO WARNING; -- don't show all the truncate messages
 
-select plan(6);
+select plan(7);
```
matthieu-foucault, line 4: those tests are probably more complex than they should be now, but I guess that's fine

matthieu-foucault (changes requested):

schema/deploy/computed_columns/project_revision_project_manager_form_changes_by_label.sql
dleard: unless 2 users are concurrently updating the same revision?

matthieu-foucault: If 2 users were able to do concurrent updates, they would still be updating the same `form_change` record

dleard: ah indeed!

matthieu-foucault (approved):

PR #302, opened 2022-02-28
Update DefaultLayout.tsx

gurjmatharu:
- added margin:auto to main

### DIFF
--- a/app/components/Layout/DefaultLayout.tsx
+++ b/app/components/Layout/DefaultLayout.tsx
@@ -50,11 +50,11 @@ const DefaultLayout: React.FC<Props> = ({
             min-height: 100vh;
             display: flex;
             flex-direction: column;
+            background-color: #fafafc;
           }
           main {
             padding: 30px 40px;
             flex-grow: 1;
-            background-color: #fafafc;
             margin: auto;
           }
         `}
--- a/app/components/Layout/DefaultLayout.tsx
+++ b/app/components/Layout/DefaultLayout.tsx
@@ -51,11 +51,11 @@ const DefaultLayout: React.FC<Props> = ({
             display: flex;
             flex-direction: column;
           }
-
           main {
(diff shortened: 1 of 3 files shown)


app/components/Layout/DefaultLayout.tsx
```diff
@@ -51,11 +51,11 @@ const DefaultLayout: React.FC<Props> = ({
             display: flex;
             flex-direction: column;
           }
-
           main {
             padding: 30px 40px;
             flex-grow: 1;
             background-color: #fafafc;
+            margin: auto;
```
matthieu-foucault, line 58: - The `background-color` rule may need to be moved to the `#page-wrap` rule set, as the side margins are white
- The "Contact Form" page ends up being too narrow, see screenshot below. Maybe a margin to the right of the "New Contact"/"Edit Contact" heading will fix it? 
image

matthieu-foucault (changes requested):
Not sure why your e2e test failed, but your branch does need to be updated.
We prefer `git rebase` vs merging the base branch into your branch, to keep the git history readable.

app/pages/cif/contact/form/[contactForm].tsx
```diff
@@ -120,6 +120,9 @@ function ContactFormPage({ preloadedQuery }: RelayProps<{}, ContactFormQuery>) {
           justify-content: space-between;
           align-items: start;
         }
+        header h2 {
+          padding-right: 10px;
+        }
```
Comment (matthieu-foucault) [inline app/pages/cif/contact/form/[contactForm].tsx:125]: Let's add the same thing to the contact view page (`app/pages/cif/contact/[contact].tsx`)

Comment (gurjmatharu) [reply app/pages/cif/contact/form/[contactForm].tsx:125]: Added - see update.

matthieu-foucault (dismissed):

matthieu-foucault (approved):

PR #304, opened 2022-02-28
Feat: multiple project managers

dleard:
Allows adding multiple managers to a project.

`addManagerToRevision` and `deleteManagerFromRevision` mutations have a query in the mutation return. The record returned from `projectManagerFormChangesByLabel` is a composite return value that doesn't represent an actual single database record. This means that we can't use the connection directive to update the data. We may want to look into [subscriptions](https://relay.dev/docs/guided-tour/updating-data/graphql-subscriptions/) at some point. I think they may end up being necessary to handle concurrent editing.

We could also look into using [refetchable fragments](https://relay.dev/docs/guided-tour/refetching/refreshing-fragments/) to refresh the ProjectManagerForm fragment, though it kind of sounds like this is essentially the same as adding a query to the mutation return.


app/components/Form/ProjectManagerForm.tsx
```diff
@@ -38,27 +69,168 @@ const ProjecManagerForm: React.ForwardRefRenderFunction<any, Props> = (
...
+  return (
+    <>
+      <Grid cols={10} align="center">
+        <Grid.Row>
+          <Grid.Col span={10}>
+            <FormBorder title="Project Managers">
+              {projectRevision.managerFormChanges.edges.map(({ node }) => (
+                <React.Fragment key={node.projectManagerLabel.id}>
+                  <Grid.Row>
+                    <label>{node.projectManagerLabel.label}</label>
```
matthieu-foucault, line 182:
```suggestion
                    <label htmlFor={your_select_id}>{node.projectManagerLabel.label} (optional)</label>
```
In the current behaviour, all managers are optional, so this should be reflected in the label. 
We do have a `FieldLabel` component, which could be reused here. At least the `getRequiredLabel` function could be reused.

app/components/Form/ProjectManagerForm.tsx
```diff
@@ -38,27 +69,168 @@ const ProjecManagerForm: React.ForwardRefRenderFunction<any, Props> = (
...
-  return <FormBase {...props} ref={ref} schema={schema} uiSchema={uiSchema} />;
+  // Add a manager to the project revision
+  const [applyAddManagerToRevision] = useAddManagerToRevisionMutation();
+  const addManager = (data: any) => {
+    applyAddManagerToRevision({
+      variables: {
+        projectRevision: projectRevision.id,
+        projectRevisionId: projectRevision.rowId,
+        newFormData: data,
+      },
```
matthieu-foucault, line 101: can we have an optimistic response here? The component flashes with an empty value and then the value changes when the network request succeeds

app/components/Form/ProjectManagerForm.tsx
```diff
@@ -38,27 +69,168 @@ const ProjecManagerForm: React.ForwardRefRenderFunction<any, Props> = (
...
+                        ObjectFieldTemplate={EmptyObjectFieldTemplate}
+                      />
+                    </Grid.Col>
+                    <Grid.Col span={4}>
+                      <Button
+                        variant="secondary"
+                        size="small"
+                        onClick={() => deleteManager(node.formChange?.id)}
+                      >
+                        Clear
```
matthieu-foucault, line 211: should be disabled if there is no manager selected, or if the mutation is already in flight for this manager. You may need to refactor the react fragment into a separate component to track individual mutations

matthieu-foucault (changes requested):

app/next-env.d.ts
```diff
@@ -1,4 +1,5 @@
 /// <reference types="next" />
+/// <reference types="next/types/global" />
```
matthieu-foucault, line 2: you need to update your local yarn packages. This was removed in the latest nextjs updage

app/components/Form/ProjectManagerForm.tsx
pbastia: We could also make the clear button a noop if the mutation is in flight or if there is no manger selectd, I don't think that would harm the UX, things would be cleared again from the user's perspective

app/tests/unit/components/Form/ProjectManagerForm.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+    expect(mutationSpy).toHaveBeenCalledWith({
+      variables: {
+        input: {
+          id: "Change 2 ID",
+        },
+        projectRevision: "Test Revision ID",
+      },
+    });
+  });
+
```
pbastia, line 173: We're missing a test that makes sure the `updateFormChangeMutation` is called when we update a dropdown field without clearing it

pbastia (changes requested):
Nice! 👌  Just missing a small test

app/mutations/Manager/addManagerToRevision.ts
```diff
@@ -0,0 +1,54 @@
...
+                  id
+                  rowId
+                  label
+                }
+                formChange {
+                  id
+                  newFormData
+                }
+              }
+            }
```
matthieu-foucault, line 37: https://bcgov-cas.promyze.app/#/CAS%20GGIRCS/home?practiceId=61e1eb156e4cd70fe286ec9f
You could reuse `ProjectManagerForm_query` here, or, if that fragment was split in two, `ProjectManagerForm_projectRevision`

app/components/Form/ProjectManagerForm.tsx
```diff
@@ -38,27 +69,168 @@ const ProjecManagerForm: React.ForwardRefRenderFunction<any, Props> = (
...
+        projectRevisionId: projectRevision.rowId,
+        newFormData: data,
+      },
+    });
+  };
+
+  // Delete a manager from the project revision
+  const [discardFormChange] = useDeleteManagerFromRevisionMutation();
+  const deleteManager = (id: string) => {
+    discardFormChange({
```
matthieu-foucault, line 108: The action here needs to depend on the `form_change` `operation`:

- If `operation === "CREATE"`, then you can discard it, as the manager was added in the same revision
- If `operation === "UPDATE"`, then the `form_change.operation`  should be updated to `ARCHIVE`

matthieu-foucault (changes requested):

matthieu-foucault (changes requested):
Haven't done a pass on the code yet, but there's an issue with the screenshots:
image

"required" validation should not show if this is optional

app/components/Form/ProjectManagerForm.tsx
```diff
@@ -1,64 +1,262 @@
...
+        debounceKey: formChangeId,
+      });
+      // If a form_change does not exist, and the payload contains a cifUserId create a form_change record
+    } else if (formChange?.cifUserId) {
+      addManager(data);
+    }
+    // If a form_change exists, and the payload does not contain a cifUserId delete it
+    else if (formChangeId && !formChange.cifUserId && !discardInFlight) {
+      deleteManager(formChangeId);
+    }
```
pbastia, line 211: I feel this entire pattern of creating / updating / deleting form changes depending on whether we have a certain field or not will be reused over and over, just a noting a thought for later on how we could reuse that logic

matthieu-foucault: Yes, this may be refactored into a hook. We can do that after we have two examples of this pattern (there will be a second one in #280 )

matthieu-foucault (approved):
-- nvm -- reviewing too many PRs in parallel

app/pages/cif/project-revision/[projectRevision]/index.tsx
```diff
@@ -171,18 +168,13 @@ export function ProjectRevision({
...
-                handleChange(
-                  query.projectRevision.projectManagerFormChange,
-                  change.formData
-                )
+            <ProjecManagerFormGroup
+              query={query}
+              revision={query.projectRevision}
+              projectManagerFormRef={projectManagerFormRef}
+              setValidatingForm={(validator) =>
+                (projectManagerFormRef.current = validator)
```
Comment (pbastia) [inline app/pages/cif/project-revision/[projectRevision]/index.tsx:176]: I'd really like to find better wording for this function / callback if you have ideas, I'm quite unhappy with what I chose when I first implemented it.
This is just confusing

pbastia (commented):
looks good! I'll let @matthieu-foucault approve if he has more comments

pbastia (approved):

app/data/jsonSchemaForm/projectManagerSchema.ts
```diff
@@ -2,7 +2,6 @@ const projectManagerSchema = {
   $schema: "http://json-schema.org/draft-07/schema",
   type: "object",
   title: "Project Manager",
-  required: ["cifUserId"],
```
matthieu-foucault, line 5: This is an interesting one, as this property is actually required in the database. Can we leave a comment below to explain why we don't want it to be required in the JSON schema?

matthieu-foucault (changes requested):

app/components/Form/ProjectManagerForm.tsx
```diff
@@ -38,27 +59,214 @@ const ProjecManagerForm: React.ForwardRefRenderFunction<any, Props> = (
...
+  };
+
+  return (
+    <>
+      <React.Fragment key={change.projectManagerLabel.id}>
```
matthieu-foucault, line 226: Don't need those two fragments

matthieu-foucault (approved):

matthieu-foucault (approved):

PR #313, opened 2022-03-01
Feat: display validation errors

dleard:
colour for error messages / icon taken from:
http://web-accessibility.carnegiemuseums.org/design/color/

### DIFF
--- a/app/components/Form/FormBase.tsx
+++ b/app/components/Form/FormBase.tsx
@@ -2,14 +2,17 @@ import Form from "lib/theme/FormWithTheme";
 import { forwardRef } from "react";
 import { FormProps, AjvError } from "@rjsf/core";
 import { customTransformErrors } from "lib/theme/customTransformErrors";
-import { customFormats, customFormatsErrorMessages } from "data/jsonSchemaForm/customFormats";
+import {
+  customFormats,
+  customFormatsErrorMessages,
+} from "data/jsonSchemaForm/customFormats";
 
 const FormBase: React.ForwardRefRenderFunction<any, FormProps<any>> = (
   props,
   ref
 ) => {
   console.log(props);
-  console.log(ref)
+  console.log(ref);
 
   const transformErrors = (errors: AjvError[]) => {
     return customTransformErrors(errors, customFormatsErrorMessages);
--- a/app/components/Form/FormBase.tsx
+++ b/app/components/Form/FormBase.tsx
@@ -1,23 +1,48 @@
 import Form from "lib/theme/FormWithTheme";
 import { forwardRef } from "react";
-import { FormProps } from "@rjsf/core";
+import { FormProps, AjvError } from "@rjsf/core";
+import { customTransformErrors } from "lib/theme/customTransformErrors";
+import { customFormats, customFormatsErrorMessages } from "data/jsonSchemaForm/customFormats";
 
 const FormBase: React.ForwardRefRenderFunction<any, FormProps<any>> = (
   props,
   ref
 ) => {
+  console.log(props);
+  console.log(ref)
+
+  const transformErrors = (errors: AjvError[]) => {
+    return customTransformErrors(errors, customFormatsErrorMessages);
+  };
   return (
     <>
       <Form
         {...props}
         // @ts-ignore
         ref={ref}
+        customFormats={customFormats}
+        transformErrors={transformErrors}
         noHtml5Validate
         omitExtraData
+        showErrorList={false}
       ></Form>
       <style jsx>{`
         :global(label) {
           font-weight: bold;
+        }
+        .errors {
+          margin: 1em 0 0 0;
+          padding: 20px;
+          background: #ce5c5c;
+          color: white;
+          font-size: 20px;
+        }
+        :global(ul.error-detail) {
+          padding: 0;
+          list-style: none;
+        }
+        :global(li.text-danger) {
+          color: #cd2026 !important;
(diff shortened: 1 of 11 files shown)


app/components/Form/FormBase.tsx
```diff
@@ -1,24 +1,49 @@
...
       <style jsx>{`
         :global(label) {
           font-weight: bold;
         }
+        .errors {
+          margin: 1em 0 0 0;
+          padding: 20px;
+          background: #ce5c5c;
+          color: white;
+          font-size: 20px;
```
matthieu-foucault, line 38: I don't think that is does anything, or is the style we want for errors

app/lib/theme/customTransformErrors.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const customTransformErrors = (
+  errors: AjvError[],
+  customFormatsErrorMessages: { [key: string]: string }
+) => {
+  // Ignore oneOf errors https://github.com/rjsf-team/react-jsonschema-form/issues/1263
+  return errors
+    .filter((error) => error.name !== "oneOf")
+    .map((error) => {
+      if (error.name !== "format") return error;
```
matthieu-foucault, line 11: Can we update `is a required property` (`error.name == "required"`) to a more actionable phrasing?
e.g. "Please enter a value.". For dropdown and radio buttons, "Please select a value." would be more appropriate, but I'm not sure if we can easily find out the input type here

app/lib/theme/FieldTemplate.tsx
```diff
@@ -1,15 +1,38 @@
...
       </div>
       <style jsx>{`
         div:not(:last-child) {
           margin-bottom: 1em;
         }
+        :global(.error-icon) {
+          color: #cd2026;
+        }
+        :global(.error-div) {
+          height: 5%;
```
matthieu-foucault, line 32: should this height be in `rem` instead? A height in `%` will be relative to the parent component's height, no?

app/components/Form/FormBase.tsx
```diff
@@ -1,24 +1,49 @@
...
+        }
+        :global(ul.error-detail) {
+          padding: 0;
+          list-style: none;
+        }
+        :global(li.text-danger) {
+          color: #cd2026 !important;
+        }
```
matthieu-foucault, line 46: Are those needed? I see styles in the `FieldTemplate` as well

app/lib/theme/FieldTemplate.tsx
```diff
@@ -1,15 +1,38 @@
...
+import { faExclamationTriangle } from "@fortawesome/free-solid-svg-icons";
+
+const FieldTemplate = ({ children, errors, help, rawErrors }) => {
   return (
     <>
       <div>
         {children}
-        {errors}
+        {rawErrors ? (
+          <div className={"error-div"}>
```
matthieu-foucault, line 10:
```suggestion
          <div className="error-div">
```

app/lib/theme/FieldTemplate.tsx
```diff
@@ -1,15 +1,38 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faExclamationTriangle } from "@fortawesome/free-solid-svg-icons";
+
+const FieldTemplate = ({ children, errors, help, rawErrors }) => {
   return (
     <>
       <div>
         {children}
-        {errors}
+        {rawErrors ? (
```
matthieu-foucault, line 9: this condition can be inside of the `error-div`

app/lib/theme/FieldTemplate.tsx
```diff
@@ -1,15 +1,38 @@
...
+          <div className={"error-div"}>
+            <FontAwesomeIcon
+              className="error-icon"
+              icon={faExclamationTriangle}
+            />
+            {errors}
```
matthieu-foucault, line 15: There's a bit too much margin between the triangle and the error message

app/lib/theme/FieldTemplate.tsx
```diff
@@ -1,15 +1,38 @@
...
+          </div>
+        ) : (
+          <div className={"error-div"}>
+            <div>&nbsp;</div>
```
matthieu-foucault, line 19: is this needed?

app/data/jsonSchemaForm/customFormats.ts
```diff
@@ -0,0 +1,7 @@
+export const customFormats = {
+  rfpDigits: "d{3,4}",
+};
+
+export const customFormatsErrorMessages = {
+  rfpDigits: "random RFP digits should be 3 or 4 digits",
```
matthieu-foucault, line 6:
```suggestion
  rfpDigits: "Please enter 3 or 4 digits for the random RFP digits",
```

Nicer phrasing :)

matthieu-foucault (changes requested):

app/lib/theme/customTransformErrors.ts
dleard: I fiddled with getting the input type into the error object, but didn't find a quick solution. Leaving it as "Please enter a value" for all "required" error messages.

matthieu-foucault (changes requested):
Looks like you just need an update of your e2e and unit tests to match the latest changes

matthieu-foucault (approved):

matthieu-foucault (approved):

PR #324, opened 2022-03-08
Feat/form controls should have consistent styling

gurjmatharu:
Please see comments on issue: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/272

### DIFF
--- a/app/components/Form/SelectProjectStatusWidget.tsx
+++ b/app/components/Form/SelectProjectStatusWidget.tsx
@@ -45,7 +45,7 @@ const SelectProjectStatus: React.FunctionComponent<WidgetProps> = (props) => {
   }, [allFundingStreamRfpProjectStatuses, fundingStreamRfpId]);
 
   return (
-    <>
+    <div>
       <label htmlFor={id}>{label}</label>
       <Dropdown
         id={id}
@@ -68,12 +68,17 @@ const SelectProjectStatus: React.FunctionComponent<WidgetProps> = (props) => {
       </Dropdown>
       <style jsx>
         {`
-          :global(input) {
+          div :global(input) {
             width: 100%;
+          }
+          div :global(.pg-select-wrapper) {
+            padding: 9px 20px 9px 2px;
+            margin-bottom: 5px;
+            margin-top: 2px;
           }
         `}
       </style>
-    </>
+    </div>
   );
 };
 
--- a/app/lib/theme/widgets/SearchDropdownWidget.tsx
+++ b/app/lib/theme/widgets/SearchDropdownWidget.tsx
@@ -75,6 +75,7 @@ const SearchDropdownWidget: React.FC<WidgetProps> = (props) => {
             outline-style: solid;
             outline-width: 4px;
             outline-color: #3B99FC;
+            outline-offset: 1px;
           }
         `}
       </style>
--- a/app/lib/theme/widgets/SearchDropdownWidget.tsx
+++ b/app/lib/theme/widgets/SearchDropdownWidget.tsx
@@ -72,7 +72,7 @@ const SearchDropdownWidget: React.FC<WidgetProps> = (props) => {
         }}
       />
 
-      <style jsx global>
+      <style jsx>
         {`
           .money:focus {
             outline-style: solid;
--- a/app/lib/theme/widgets/SearchDropdownWidget.tsx
+++ b/app/lib/theme/widgets/SearchDropdownWidget.tsx
@@ -4,7 +4,7 @@ import Widgets from "@rjsf/core/dist/cjs/components/widgets";
 import TextField from "@mui/material/TextField";
 import Autocomplete from "@mui/material/Autocomplete";
 import FieldLabel from "./FieldLabel";
-import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown';
+import KeyboardArrowDownIcon from "@mui/icons-material/KeyboardArrowDown";
 
 const SearchDropdownWidget: React.FC<WidgetProps> = (props) => {
   const { onChange, schema, placeholder, readonly, label, required, uiSchema } =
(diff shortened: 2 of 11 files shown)


matthieu-foucault (commented):
To do before this PR can be reviewed
- install the `pre-commit` and `commit-msg` hooks using [`pre-commit`](https://pre-commit.com/), and update formatting
- update e2e test
- request PR review via our Teams channel, so that I'm not the only team member notified

app/lib/theme/widgets/SearchDropdownWidget.tsx
```diff
@@ -43,11 +44,41 @@ const SearchDropdownWidget: React.FC<WidgetProps> = (props) => {
...
+<style jsx global>
+        {`
+          .money:focus {
+            outline-style: solid;
+            outline-width: 4px;
+            outline-color: #3B99FC;
+            outline-offset: 1px;
+          }
+        `}
+      </style>
```
matthieu-foucault, line 81: Delete this? This is also failing the formatting check

app/lib/theme/widgets/SelectParentWidget.tsx
```diff
@@ -84,11 +84,19 @@ const SelectParentWidget: React.FunctionComponent<
...
+          .pg-select-wrapper{
+            padding: 9px 2px 9px 2px;
+            margin-bottom: 5px;
+            margin-top:2px;
+          }
+          .pg-select-input {
+            color:#8e8e8e;
+          }
         `}
       </style>
```
matthieu-foucault, line 101: We want to avoid `style jsx global` as it removes the scoping from these styles, meaning that they might have side-effects and affect other components on the page. Now, I don't think we've been using [`:global()`](https://github.com/vercel/styled-jsx#one-off-global-selectors) properly either: I think that we should have an outer `div` (instead of a fragment) and have the following styles:

```suggestion
      <style jsx>
        {`
          div :global(input) {
            width: 100%;
          }
          div :global(.pg-select-wrapper) {
            padding: 9px 2px 9px 2px;
            margin-bottom: 5px;
            margin-top:2px;
          }
          div :global(.pg-select-input) {
            color:#8e8e8e;
          }
        `}
      </style>
```

The `div` part of the selector will be scoped to this specific component, and `:global` allows to select classes that are rendered by children components

matthieu-foucault (changes requested):
I see some e2e tests are failing because of contrast issues, so I can't fully review this yet

app/lib/theme/widgets/SelectParentWidget.tsx
```diff
@@ -86,12 +86,20 @@ const SelectParentWidget: React.FunctionComponent<
...
         {`
-          :global(input) {
+          div :global(input) {
             width: 100%;
           }
+          div :global(.pg-select-wrapper) {
+            padding: 9px 2px 9px 2px;
+            margin-bottom: 5px;
+            margin-top: 2px;
+          }
```
matthieu-foucault, line 96: Those should be applied to the `SelectWidget` as well

app/lib/theme/widgets/SearchDropdownWidget.tsx
```diff
@@ -43,11 +44,44 @@ const SearchDropdownWidget: React.FC<WidgetProps> = (props) => {
...
+          borderRadius: "0.25em",
+          marginTop: "0.2em",
+          "&.Mui-focused": {
+            outlineStyle: "solid",
+            outlineWidth: "4px",
+            outlineColor: "#3B99FC",
+            outlineOffset: "1px",
+          },
+        }}
+        popupIcon={<KeyboardArrowDownIcon />}
```
matthieu-foucault, line 58: you can set this icon color to `black` to get closer to the `SelectWidget` one.

app/lib/theme/widgets/SearchDropdownWidget.tsx
```diff
@@ -43,11 +44,44 @@ const SearchDropdownWidget: React.FC<WidgetProps> = (props) => {
...
-          return <TextField {...params} placeholder={placeholder} />;
+          return (
+            <TextField
+              {...params}
+              placeholder={placeholder}
+              variant="standard"
+              InputProps={{ ...params.InputProps, disableUnderline: true }}
+              sx={{
+                padding: "5px",
+                marginLeft: "2px",
```
matthieu-foucault, line 68: background color should be `white`

matthieu-foucault (changes requested):
A couple more tweaks, and I think it will be consistent enough and not needing another sub-task/PR.
The style on hover on the search dropdown has somehow been taken care of, and if you set the search dropdown arrow to black, it will look similar enough to the select widget.

app/lib/theme/widgets/SelectParentWidget.tsx
```diff
@@ -86,12 +86,20 @@ const SelectParentWidget: React.FunctionComponent<
...
+          div :global(.pg-select-input) {
+            color: #585555;
+          }
```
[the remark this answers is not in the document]
gurjmatharu: Reverting to black - perhaps we should discuss a consistent component for all dropdowns on a different issue.

app/lib/theme/widgets/SelectParentWidget.tsx
gurjmatharu: Not sure if you mean `SelectWidget` or `SelectProjectStatusWidget.tsx` or both.  Applied css to both for now.

matthieu-foucault: I guess both, but there's probably some refactoring needed there (in another PR). I would expect `SelectProjectStatusWidget` to use `SelectWidget`

app/components/Form/SelectProjectStatusWidget.tsx
```diff
@@ -68,12 +68,17 @@ const SelectProjectStatus: React.FunctionComponent<WidgetProps> = (props) => {
       </Dropdown>
       <style jsx>
         {`
-          :global(input) {
+          div :global(input) {
             width: 100%;
           }
```
matthieu-foucault, line 73: Looks like this did have a side effect: It needs to be added in `MoneyWidget` as well now, as it's not using the whole width anymore

matthieu-foucault (changes requested):
Just one more change

matthieu-foucault (approved):

PR #349, opened 2022-03-14
Remove blank loading page when data is loading

matthieu-foucault:
- Added a `LoadingFallback` for transitions between pages
- `Table` component pre-fetches the data before updating the router, following the [behaviour described in the relay docs](https://relay.dev/docs/guided-tour/refetching/refetching-queries-with-different-data/#if-you-need-to-avoid-suspense-1)

### DIFF
--- a/app/components/Table/index.tsx
+++ b/app/components/Table/index.tsx
@@ -7,6 +7,7 @@ import {
   GraphQLTaggedNode,
   useRelayEnvironment,
 } from "react-relay";
+import Sentry from "@sentry/react";
 import FilterRow from "./FilterRow";
 import { FilterArgs, PageArgs, TableFilter } from "./Filters";
 import Pagination from "./Pagination";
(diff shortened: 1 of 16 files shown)


dleard (dismissed):
Looks good. Table still flashes a little bit, I think probably because we're setting 'disabled' briefly during fetch? Not a blocker. I'm going to add a tech debt card to implement this same thing in cas-ggircs.

matthieu-foucault:
>  Table still flashes a little bit

Are you referring to the headers getting lighter?
This might be solved by a different "disabled" style. We can discuss this with our designers if that's something we want to improve later

dleard:
> > Table still flashes a little bit
> 
> Are you referring to the headers getting lighter? This might be solved by a different "disabled" style. We can discuss this with our designers if that's something we want to improve later

Yes, that's what I was seeing. It's pretty minor, but yeah, sounds like something to run by designers.

app/components/Table/index.tsx
```diff
@@ -31,6 +47,44 @@ const Table: React.FC<Props> = ({
...
+      // See https://github.com/facebook/relay/blob/b8e78ca0fbbfe05f34b4854484df574d91ba2113/website/docs/guided-tour/refetching/refetching-queries-with-different-data.md#if-you-need-to-avoid-suspense
+      fetchQuery(
+        environment,
+        pageQuery,
+        withRelayOptions.variablesFromContext(url)
+      ).subscribe({
+        complete: afterFetch,
+        // if the query fails, we still want to update the route,
+        // which will retry the query and let a 500 page be rendered if it fails again
+        error: afterFetch,
```
pbastia, line 82: I wonder if that will trigger the Sentry boundary before retrying the query? Or is it one of the special cases where we have to call `Sentry.captureException(...)`

matthieu-foucault: This isn't an uncaught exception, so Sentry won't be automatically called. I can add `Sentry.captureException` here

dleard (approved):

PR #350, opened 2022-03-14
chart `db.preInitCommand` is run in a pre-upgrade hook

matthieu-foucault:


chart/cas-cif/templates/cronJobs/db-init.yaml
```diff
@@ -66,10 +66,75 @@ spec:
...
                   psql -d $(CIF_DATABASE) -v "ON_ERROR_STOP=1"<<EOF
                     create extension if not exists postgres_fdw;
                     grant usage on foreign data wrapper postgres_fdw to $(CIF_USER);
                   EOF
+---
+{{- if .Values.db.preInitCommand }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ template "cas-cif.fullname" . }}-db-pre-upgrade
```
pbastia, line 80: should we call it `{{ fullname }}-db-pre-init` and have this template in its own file?

pbastia (approved):
looks great! thanks for looking into this

pbastia (approved):
🍾

Issue #332, opened 2022-03-09
Add support for local development using docker & mini shift


gurjmatharu:
Currently the way for developers to get up and running is through the make file.  This becomes difficult once the team has various development machines (Mac m1, ubuntu, windows, etc).

We should support a local docker version with support for both `docker-compose` and [minishift](https://github.com/minishift/minishift)

Some resources:
- https://youtu.be/7_go1R7J8fM
- https://youtu.be/d5JNUsQ7k2Q

matthieu-foucault:
#### `docker-compose`

What components of the application do you want to run with docker-compose? 

- **PostgreSQL**: I agree that setting up postgres with asdf is unwieldy, and does not reflect what is used in production (a Spilo container). Connecting to postgres as a privileged user locally has also led to issues in the past with permissions, and a docker container would be closer to our prod environment
- **Database Migrations (sqitch)**: Perl is also a PITA to setup locally. We could run something similar to the `schema` container (or this container with a custom command) to deploy the sqitch changes. Additionally, this container could watch the deploy scripts and redeploy when they don't match the deployed changes. I wrote the [`sqitch check`](https://sqitch.org/docs/manual/sqitch-check/) command a couple of years ago to help with that, or we could just use watchman and drop and redeploy the db anytime the scripts change. I do recall this being very resource-intensive, but a couple of our build tools have migrated to Rust since and are way more efficient, so this might not be as bad nowadays.
- **Express server**: unless node itself is difficult to setup on certain machines, I'd say this is unnecessary.

#### `minishift`

The minishift repo description says 
>Run OpenShift 3.x locally

This might be problematic as we're deploying to 4.x.

I'd like more details as to what benefits running minishift would bring here, and what the resources requirements are.

gurjmatharu:
docker-compose
I don't have an opinion on which components, the major value in this would be to get a developer up and running as fast as possible.  Volume mapping the necessary parts allowing you to start hacking.


For open shift 4 the project would be this one 👉🏽  https://github.com/code-ready/crc/
If we have docker-compose running this would be redundant, except for working on helm chart work and deployments.  So unsure on how much of a need this one would be.

PR #355, opened 2022-03-15
Feat/docker compose setup

gurjmatharu:
#332


app/Dockerfile.dev
```diff
@@ -0,0 +1,27 @@
...
+RUN yarn install
+
+# Copy the rest of our Next.js folder into /app
+COPY . /app
+
+# Ensure port 3000 is accessible to our system
+EXPOSE 3000
+
+# Run yarn dev, as we would via the command line
+CMD ["yarn", "dev"]
```
matthieu-foucault, line 27: Are you using this file? I don't see it in your docker compose.
If that's useful in your environment, happy to leave it.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    build:
+      context: .
+      dockerfile: dev/db/Dockerfile
+    environment:
+      - POSTGRES_NAME=postgres
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_DB=cif
+    volumes:
+      - "./dev/db/data:/var/lib/postgresql/data"
```
matthieu-foucault, line 14: `./dev/db/data` should be added to our `.gitignore`

matthieu-foucault (changes requested):

app/Dockerfile.dev
gurjmatharu: I am, I could also just keep it locally either way I'm okay.

matthieu-foucault (approved):

PR #365, opened 2022-03-16
chore: metabase knp

dleard:
adds a KNP to allow metabase ingress to read the cif database

### DIFF
--- a/chart/cas-cif/templates/networkPolicies/metabase-ingress.yaml
+++ b/chart/cas-cif/templates/networkPolicies/metabase-ingress.yaml
@@ -1,4 +1,3 @@
-{{- if .Values.networkPolicies.enable }}
 kind: NetworkPolicy
 apiVersion: networking.k8s.io/v1
 metadata:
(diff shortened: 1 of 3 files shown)


chart/cas-cif/templates/networkPolicies/metabase-ingress.yaml
```diff
@@ -0,0 +1,18 @@
...
+spec:
+  podSelector:
+    matchLabels:
+      release: {{ .Release.Name }}
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          name: {{ .Values.ggircs.prefix }}
+          environment: {{ .Values.ggircs.environment }}
```
matthieu-foucault, line 15: You can match the `kubernetes.io/metadata.name` label, so that you don't need to break down the namespace into prefix and environment, and don't need to add more values to the chart (you can revert changes to the Makefile)

matthieu-foucault (changes requested):

matthieu-foucault (approved):

Issue #275, opened 2022-02-15
Current date should be mocked in e2e tests


matthieu-foucault:
#### Describe the task

This is leading to spurious diffs with happo.
When running the application with the `ENABLE_DB_MOCKS` flag, the `mocks.mocked_timestamp` cookie allows to set the current date.

The following commands can be added to Cypress, in `app/cypress/support/commands.js`.

```js
Cypress.Commands.add("useMockedTime", (dateTime) => {
  cy.setCookie(
    "mocks.mocked_timestamp",
    Math.round(dateTime.getTime() / 1000).toString()
  );
});
Cypress.Commands.add("clearMockedTime", () => {
  cy.clearCookie("mocks.mocked_timestamp");
});
```

then, tests can mock the current date using 

```js
cy.useMockedTime(new Date("June 10, 2020 09:00:00"));
```

#### Additional context

- Additionally, tests should ensure that the saving indicator says "Changes Saved" before triggering happo snapshots, to further prevent spurious diffs

PR #377, opened 2022-03-17
Test/e2e dates

BCerki:
Issue #275 

I've looked over the happo shots and I'm happy with them, but I didn't click the approve button so the reviewer will have a chance to look too if desired


pbastia (approved):
Very nice! No more happo diffs because of the dates 🎉

PR #389, opened 2022-03-18
Fix: refresh data on clear

dleard:
uses a custom dynamic delete mutation to handle the refreshing of data.

### DIFF
--- a/app/components/Form/ProjectManagerForm.tsx
+++ b/app/components/Form/ProjectManagerForm.tsx
@@ -140,6 +140,7 @@ const ProjectManagerForm: React.FC<Props> = (props) => {
   const handleClear = () => {
     let variables;
     let optimisticResponse;
+    // If the form_change has a CREATE operation then we want to delete the form_change from the database
     if (change.formChange?.operation === "CREATE")
       variables = {
         input: {
@@ -147,6 +148,7 @@ const ProjectManagerForm: React.FC<Props> = (props) => {
         },
         projectRevision: projectRevisionId,
       };
+    // If the form_change has an UPDATE operation then we want to update the form_change in the database with operation = 'ARCHIVE'
     else {
       variables = {
         input: {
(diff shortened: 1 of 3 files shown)


app/components/Form/ProjectManagerForm.tsx
```diff
@@ -135,13 +135,43 @@ const ProjectManagerForm: React.FC<Props> = (props) => {
...
+            id: change.formChange.id,
+            operation: "ARCHIVE",
+            newFormData: {},
+          },
+        },
+      };
+    }
+    deleteManager({
+      variables,
+      optimisticResponse,
```
matthieu-foucault, line 174: Can we update `useDiscardFormChange` instead to have an archive and delete mutation as an option, so that the branching logic on `operation === "CREATE"` isn't duplicated?

app/mutations/Manager/deleteManagerFromRevision.ts
```diff
@@ -32,16 +53,116 @@ export const mutation = graphql`
...
+
+export const useDeleteManagerFromRevisionMutation = (): [
+  (options: DeleteManagerOptions) => void,
+  boolean
+] => {
+  const [deleteManager, isDeleting] = useDeleteMutation();
+  const [archiveManager, isArchiving] = useArchiveMutation();
+  const deleteOrArchiveManager = (options: DeleteManagerOptions) => {
+    const { context, onCompleted, onError } = options;
+    if (context.operation === "CREATE") {
```
matthieu-foucault, line 128: This logic was moved away from the component, but is still a duplicate of `app/hooks/useDiscardFormChange.ts`. Can't this be factorized?

dleard: Possibly? I was trying to ensure the nice generic hook stays generic & handle this specific case where the mutation requires a query in the return separately. I think `deleteManagerFromRevision` could maybe call `useDiscardFormChange` if `useDiscardFormChange` was refactored to take custom delete/archive mutations and a projectRevisionId as parameters. But then we'd need to add logic specific to the deletion of managers to the generic hook's code, which I was trying to avoid. This does cause duplication though.

matthieu-foucault: Let's leave it as is and see if we want to refactor later

matthieu-foucault (approved):

Issue #325, opened 2022-03-09
As a CIF user, I want to be redirected to the 404 page when accessing a invalid URL, so that I don't trigger an error


matthieu-foucault:
#### Description:

Brief description of the work involved, including any user research context or key decisions.

#### Acceptance Criteria:

Given
When
Then

#### Development Checklist:

- [ ] Checklist item
- [ ] Checklist item
- [ ] Checklist item
- [ ] Meets the DOD

#### Notes:

-

matthieu-foucault:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/325?planning-poker) with ZenHub @BCerki @dleard @gurjmatharu @pbastia

LindsayMacfarlane:
Thanks @BCerki ! Tested in dev by entering invalid cif urls and received the following error message: Page not found. Sorry, we couldn't find the page you were looking forward.

The Return Home button brought me back to the login page.

PR #394, opened 2022-03-21
Feat/404 redirect

BCerki:
Issue #325

### DIFF
--- a/app/hooks/useRedirectTo404IfFalsy.tsx
+++ b/app/hooks/useRedirectTo404IfFalsy.tsx
@@ -1,6 +1,6 @@
 import { useEffect } from "react";
 import { useRouter } from "next/router";
-/** 
+/**
  *  Hook that returns true if the router will redirect, false otherwise.
  *  - Example usage:
  *    const isRedirecting = useRedirectTo404IfFalsy(object);


app/pages/cif/project-revision/[projectRevision]/index.tsx
```diff
@@ -58,11 +59,23 @@ export function ProjectRevision({
...
+  };
+
+  const updatedAtStatic = updatedAt();
+
   const lastEditedDate = useMemo(
-    () => new Date(query.projectRevision.updatedAt),
-    [query.projectRevision.updatedAt]
+    () => (updatedAtStatic ? new Date(updatedAtStatic) : null),
+    [updatedAtStatic]
   );
```
Comment (BCerki) [inline app/pages/cif/project-revision/[projectRevision]/index.tsx:74]: Not sure about this part--I needed to check if the project revision existed before trying to `useMemo`, but I can't use `isRedirecting` to do that because it can't come before `useMemo` per hook rules.

Comment (BCerki) [reply app/pages/cif/project-revision/[projectRevision]/index.tsx:74]: Also, my unit tests don't cover this part so I only noticed I needed it when I manually checked. Should I do any other tests (integration?) to cover it? This work is unrelated to the new hook, but it is related to whether or not you successfully get the 404 page.

app/tests/unit/hooks/useRedirectTo404IfFalsy.test.ts
```diff
@@ -0,0 +1,20 @@
+import useRedirectTo404IfFalsy from "hooks/useRedirectTo404IfFalsy";
```
BCerki, line 1: Not sure I've done this well

app/pages/cif/project-revision/[projectRevision]/index.tsx
```diff
@@ -59,10 +60,16 @@ export function ProjectRevision({
...
+      query.projectRevision?.updatedAt
+        ? new Date(query.projectRevision.updatedAt)
+        : null,
+    [query.projectRevision?.updatedAt]
   );
 
+  const isRedirecting = useRedirectTo404IfFalsy(query.projectRevision);
+  if (isRedirecting) return null;
+
   if (!query.projectRevision.id) return null;
```
Comment (matthieu-foucault) [inline app/pages/cif/project-revision/[projectRevision]/index.tsx:73]: This isn't needed. If `query.projectRevision` exists, it will have an id

matthieu-foucault (changes requested):
Looks good! just one redundant condition that can be removed

app/tests/unit/hooks/useRedirectTo404IfFalsy.test.ts
```diff
@@ -0,0 +1,20 @@
...
+import { mocked } from "jest-mock";
+jest.mock("next/router");
+mocked(useRouter).mockReturnValue({
+  replace: jest.fn(),
+} as any);
+
+describe("the useRedirectTo404IfFalsy hook", () => {
+  it("returns true if the record doesn't exist and the page should redirect", () => {
+    const { result } = renderHook(() => useRedirectTo404IfFalsy(null));
+    expect(result.current).toBe(true);
```
matthieu-foucault, line 13: you should check that `replace` was called here.

app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx
```diff
@@ -238,4 +239,24 @@ describe("The Create Project page", () => {
       true
     );
   });
+
+  it("redirects to the 404 page when a revision doesn't exist", async () => {
```
Comment (matthieu-foucault) [inline app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:243]: You're only checking that `null` is returned, not that the page was redirected

matthieu-foucault (changes requested):

matthieu-foucault:
> My unit tests don't cover this part (using a query before the redirect hook) so I only noticed the original code threw an error via manually checking. Should I do any other tests (integration?) to cover these situations? This work is unrelated to the new hook, but it is related to whether or not you successfully get the 404 page.

Somehow I can't reply to the comment itself, so replying here. 

This is in the user story's description as

> any usage of that variable prior to the return null should take that into account, e.g. using optional chaining.

To enforce this, an option would be to enable Typescript's `strictNullChecks`, but this would have to be done for the entire code base. I'm unsure why your unit tests wouldn't throw an exception though.

app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx
```diff
@@ -238,4 +239,24 @@ describe("The Create Project page", () => {
...
+    const spy = jest
+      .spyOn(require("app/hooks/useRedirectTo404IfFalsy"), "default")
+      .mockImplementation(() => {
+        return true;
+      });
```
Comment (matthieu-foucault) [inline app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:248]: why does this need to be mocked? Is the revision not actually `undefined`? If that is the case that would explain why you were not catching issues with using the undefined variable, and using a mock here is harmful to your test, as you're making your function take a path that it wouldn't otherwise.

app/hooks/useRedirectTo404IfFalsy.tsx
```diff
@@ -0,0 +1,13 @@
+import { useEffect } from "react";
+import { useRouter } from "next/router";
+
```
pbastia, line 3: I would add just a comment or 2 about how to use this, otherwise the semantics are a double-negative that isn't super easy to read.

```suggestion
/** 
 *  Hook that returns true if the router will redirect, false otherwise.
 *  - Example usage:
 *    const isRedirecting = useRedirectTo404IfFalsy(object);
 *    if(isRedirecting) return null;
 */
```

app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx
Comment (pbastia) [reply app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:248]: From what I saw the relay testing framework puts mock values wherever it's not defined in the generator configuration that we pass. 
I think if we want relay to return `undefined` values, we'll need to pass `null` somewhere  in the call to `MockPayloadGenerator.generate(...)`, and grab the result in another `const nullInitialQueryRef = loadQuery<ProjectRevisionQuery>(...)` object that we can use.

@BCerki happy to pair on that if you want, or if what I'm saying makes no sense

pbastia (changes requested):

app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx
Comment (BCerki) [reply app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:248]: Here I'm only trying to test that the component renders null (sorry for the misleading test name, forgot to change it) if the revision doesn't exist, same as I'm doing in `app/tests/unit/pages/operator/[operator].test.tsx`, etc. I'm trying to test the redirect part in the hook test.

Comment (BCerki) [reply app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:248]: Sure, that would be great, thanks

matthieu-foucault (approved):

app/tests/unit/pages/contact/[contact].test.tsx
```diff
@@ -1,53 +1,94 @@
...
 } from "__generated__/ContactViewQuery.graphql";
 import { loadQuery, RelayEnvironmentProvider } from "react-relay";
+import { MockResolvers } from "relay-test-utils/lib/RelayMockPayloadGenerator";
+import { useRouter } from "next/router";
+import { mocked } from "jest-mock";
 
-let environment;
+jest.mock("next/router");
 
-const loadContactData = (partialContact = {}) => {
```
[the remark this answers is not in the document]
Comment (matthieu-foucault) [reply app/tests/unit/pages/contact/[contact].test.tsx:11]: Yeah I think that was used in some test but isn't anymore

PR #414, opened 2022-03-23
Feat/error alert

BCerki:
Issue https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/309

This is 1/2 PRs that replaces `useMutation` with `useErrorMutation`. This PR replaces it in:
- Dashboard component
- ProjectContactForm component
- createEditContactFormChange mutation
- createNewContactFormChange mutation
- createProjectRevision mutation
- useDebouncedMutation (and all mutations/components that use `useDebouncedMutations)

The next PR will replace it in all remaining places.

Notes:
- I just put "an error occurred" for the user-friendly error messages so they're easier to find and replace after we get designer input
- I'm not confident I've got 100% unit test coverage--useDebouncedMutation is used in a lot of other functions--so would love some feedback on that.


app/components/Layout/GlobalAlert.tsx
```diff
@@ -0,0 +1,10 @@
+import Alert from "@button-inc/bcgov-theme/Alert";
+interface Props {
+  error: string;
+}
+
+const GlobalAlert = (props: Props) => {
+  return <Alert variant="danger">Error: {props.error}</Alert>;
```
matthieu-foucault, line 7: should the alert be `closable`?

app/mutations/useErrorMutation.ts
```diff
@@ -0,0 +1,52 @@
...
+  GraphQLTaggedNode,
+  IEnvironment,
+  MutationConfig,
+  MutationParameters,
+} from "relay-runtime";
+import * as Sentry from "@sentry/nextjs";
+
+export default function useErrorMutation<TMutation extends MutationParameters>(
+  mutation: GraphQLTaggedNode,
+  errorMessage: string
```
matthieu-foucault, line 27: we might want a `getErrorMessage: (relayError: Error) => string`, to allow different error messages depending on the actual error message. The documentation should say something about this function returning a user-friendly error message.

app/mutations/useErrorMutation.ts
```diff
@@ -0,0 +1,52 @@
...
+import {
+  Disposable,
+  GraphQLTaggedNode,
+  IEnvironment,
+  MutationConfig,
+  MutationParameters,
+} from "relay-runtime";
+import * as Sentry from "@sentry/nextjs";
+
+export default function useErrorMutation<TMutation extends MutationParameters>(
```
matthieu-foucault, line 25: A bit more verbose, but I think it's clearer.
```suggestion
export default function useMutationWithErrorMessage<TMutation extends MutationParameters>(
```

app/pages/_app.tsx
```diff
@@ -20,6 +21,9 @@ const initialPreloadedQuery = getInitialPreloadedQuery({
 });
 
 function MyApp({ Component, pageProps }: AppProps) {
+  const [error, setError] = useState(null);
+  const value = { error, setError };
+
```
matthieu-foucault, line 26: you can have a `useEffect` resetting the error to `null` when `Component` changes

app/components/Layout/GlobalAlert.tsx
matthieu-foucault: I don't know if we need `Error: ` in the text

app/mutations/useErrorMutation.ts
BCerki: Would this be a parameter in addition to `errorMessage`, or a replacement for it? Also, is `getErrorMessage` a function that comes with relay, or am I supposed to write it? I couldn't import it or find it in the docs (I found a bit about [user-friendly messages in schemas](https://relay.dev/docs/guided-tour/rendering/error-states/#accessing-errors-in-graphql-responses), but I don't think that's the same thing).

matthieu-foucault: A replacement for `errorMessage`, which you would have to write. This would allow to display different alerts for a given mutation, depending on the original error. With a string `errorMessage`, your mutation will always end up displaying the same error if it fails.

This is more of a future need, since we haven't discussed the actual error messages yet. I would imagine that our first `getErrorMessage` functions will be something like `() => "An error happened while saving the form"`, ignoring the actual error.

app/tests/unit/components/Form/ProjectContactForm.test.tsx
```diff
@@ -235,6 +269,24 @@ describe("The ProjectContactForm", () => {
     });
   });
 
+  it("calls useMutationWithErrorMessage and returns expected message when the Clear button is pressed", () => {
+    //Warning: Expected `optimisticResponse` to match structure of server response for mutation `updateFormChangeMutation`
```
BCerki, line 273: This test passes but there were some console warnings, and I'm not sure if they're related to `useMutationWithErrorMessage`, the test itself, or something that was in the app before I did any work on this section. I've made comments about other errors too.

matthieu-foucault (changes requested):
Our error messages should be a bit more precise and different for each mutation, so that we can easily identify which mutation triggered and error that would be reported by a user, and so that users can understand which action caused the error

app/tests/unit/components/Form/ProjectManagerForm.test.tsx
```diff
@@ -161,6 +161,7 @@ describe("The ProjectManagerForm", () => {
   });
 
   it("Calls the addManagerToRevision mutation when a new selection is made in the Manager dropdown", () => {
+    // Warning: Invalid value for prop `disabled` on <button> tag. Either remove it from the element, or pass a string or number value to keep it in the DOM. For details, see https://reactjs.org/link/attribute-behavior
```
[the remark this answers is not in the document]
BCerki: Would you like me to take out all the warning comments?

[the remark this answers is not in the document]
matthieu-foucault: Nah, they can stay. Having the unit tests run without `console.error` messages being printed would be great

matthieu-foucault (dismissed):
just needs a rebase

PR #427, opened 2022-03-29
Fix: Login-redirect

pbastia:


dleard (approved):
:100: Nice! It redirects properly now.

PR #435, opened 2022-03-29
docs: document form_changes intent and behaviour

matthieu-foucault:


mikevespi:
Looks great! One question, what happens if someone starts a form change and realize / decide they don't need it? Does the entry in `form_change` still get kept just stuck in `pending`?

docs/dbRecordsHistory.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Approach
+
+One approach often seen to track history of updates is to use and audit table, that would records all of the `insert`, `update` or `delete` operation on a table, using triggers. We also want to track changes **before** they are applied, e.g. when they are pending review, or when they are being auto-saved. Rather than having a separate table to track those changes, we decided to use a single `form_change` table, which allows us to track a change to a given database record. To avoid creating a table for every table we want to track, the `form_change` table supports tracking changes of every table, on the one condition that the tracked table has an `id` primary key of type `integer`.
+
+### Example: tracking changes to a `contact`
+
+In this example, we assume that our `contact` table only has the following columns: `id`, `given_name` and `family_name`.
+When the user opens the page to create a new contact in the system, the following database record will be created (we use a JavaScript object literal here, omitting some field, for an easier representation of the data):
```
pbastia, line 22:
```suggestion
When the user opens the page to create a new contact in the system, the following database record will be created (we use a JavaScript object literal here, omitting some fields, for an easier representation of the data):
```

docs/dbRecordsHistory.md
```diff
@@ -0,0 +1,74 @@
...
+  id: 123, // auto-generated
+  new_form_data: '{"givenName": "Bob", "familyName": "Loblaw"}', // a JSON, camelCase representation of the data, for easy use with front-end form libraries
+  operation: 'create', // a new contact record will be created
+  change_status: 'pending', // a pending change corresponds to a form that the user has not submitted yet
+  form_data_table_name: 'contact', // the table in which the data should be inserted
+  form_data_record_id: null // once the form is submitted and the contact is created, this will be populated
+}
+```
+
+When the user click the form's `submit` button, the `change_status` is updated to `committed`. A database trigger will then apply the change. In this case, the following SQL query would be performed:
```
pbastia, line 35:
```suggestion
When the user clicks the form's `submit` button, the `change_status` is updated to `committed`. A database trigger will then apply the change. In this case, the following SQL query will be performed:
```

docs/dbRecordsHistory.md
```diff
@@ -0,0 +1,74 @@
...
+  id: 123,
+  new_form_data: '{"givenName": "Bob", "familyName": "Loblaw"}',
+  operation: 'create',
+  change_status: 'committed',
+  form_data_table_name: 'contact',
+  form_data_record_id: 42
+}
+```
+
+Now that the `form_change` is committed, it cannot be modified (triggers are ensuring that the record is immutable), providing an audit trail. If a user were to open the form to edit the contact and submit it, a new `form_change` record, would be created, like so:
```
pbastia, line 54:
```suggestion
Now that the `form_change` is committed, it cannot be modified (triggers are ensuring that the record is immutable), providing an audit trail. If a user were to open the form to edit the contact and submit it, a new `form_change` record would be created:
```

pbastia (changes requested):
Very minor edits, thanks for writing this up!

dleard:
Is it worth adding a block about how on creating a project_revision, new form_change records are created from the previous revision's form_change data?

pbastia (approved):

PR #441, opened 2022-03-30
Tech debt: Enforce conventional commits

BCerki:
Issue: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/363

Notes:
- Install the commit-msg hook to get local enforcement: `pre-commit install --hook-type commit-msg`
- I tested the github action with an unconventional commit [here](https://github.com/bcgov/cas-cif/runs/5761212800?check_suite_focus=true).
- I'm using commitlint for the github action and gitlint for local checks because that's how CIIP does it, but if we want to standardize, happy to. I'd prefer commitlint because the github action requires less config than the gitlint one, and commitlint can integrate with commitizen

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,6 +86,14 @@ jobs:
           key: pre-commit-${{ env.PY }}-${{ hashFiles('.pre-commit-config.yaml') }}-v2
       - run: pre-commit run --all-files
 
+  commitlint:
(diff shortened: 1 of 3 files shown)


mikevespi (approved):
Looks good to me!

mikevespi (approved):
:shipit: !

PR #442, opened 2022-03-31
Feat/error alert part 2/2

BCerki:
Issue: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/309

### DIFF
--- a/app/mutations/useArchiveMutation.ts
+++ b/app/mutations/useArchiveMutation.ts
@@ -1,13 +1,14 @@
-import { useMutation, UseMutationConfig } from "react-relay";
+import { UseMutationConfig } from "react-relay";
 import {
   Disposable,
   GraphQLTaggedNode,
   MutationParameters,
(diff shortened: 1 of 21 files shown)


app/mutations/useArchiveMutation.ts
```diff
@@ -39,7 +40,10 @@ export default function useArchiveMutation<
   ) => Disposable,
   boolean
 ] {
-  const [updateMutation, isInFlight] = useMutation(mutation);
+  const [updateMutation, isInFlight] = useMutationWithErrorMessage(
+    mutation,
+    () => "An error occurred"
```
BCerki, line 45: I wasn't sure about the error message here

matthieu-foucault: Yeah this hook is generic. Let's leave it as-is for now

Issue #301, opened 2022-02-25
Development data should include form_change records


matthieu-foucault:
#### Describe the task

We are creating a number of entities in our development data to avoid having to recreate contacts, operators, projects, etc every time. Those records were however directly added to the `contact`, `operator` and `project` tables, respectively, which does not match how our application will create database records: `form_change` records should be created and their status updated to `committed`. For projects, `project_revision` records should be used.

Not having `form_change` records means that when editing one of these generated entities, the forms do not get initialized properly

LindsayMacfarlane:
Thanks @gurjmatharu! I'm assuming that there is nothing for me to test on this one?

gurjmatharu:
> Thanks @gurjmatharu! I'm assuming that there is nothing for me to test on this one?  

Just the projects are listed, editable, and have connected contacts. That's about it!

Issue #368, opened 2022-03-16
As a Project Coordinator, I want to see the status of every form in the task list, so that I know what my progress is


matthieu-foucault:
#### Description:

When looking at the task list, created in #366, the user should be able to tell whether each of the tasks is `Not Started`, `Complete`, `Incomplete` or has `Attention Required`

This needs to take into account pages that render multiple forms, e.g. contacts or managers

#### Acceptance Criteria:

Given that I am viewing a project revision task list
  - When none of a task's form(s) were updated since the previous revision
  Then the status in the task list should be `Not Started`

  - When some of a given task's form(s) was updated since the previous revision, their status is `staged`
  And the other forms were not updated (e.g. one contact was updated, the others were not)
  Then the status in the task list should be `Complete`

  - When some of a task's forms were updated since the previous revision
  And their status is `pending`
  Then the status in the task list should be `Incomplete`

  - When some of a task's forms were updated since the previous revision
  And their status is `staged`
  And they have validation errors
  Then the status in the task list should be `Attention Required`

#### Development Checklist:

- [ ] The task list should have a fixed width, to avoid page layout changes when the status of a task changes
- [ ] An `isPristine` computed column should be added to easily find out if a `form_change` was updated. Looking at `created_at` vs `updated_at` won't work since changes can be undone, this needs to compare the `new_form_data` against the previous form, using the column created in #374
- [ ] The `Attention Required` status should be bold
- [ ] Meets the DOD

#### Notes:

![image.png](https://images.zenhubusercontent.com/575f1a78bf93871ccd9c8bcd/4d600fe6-6dd5-4ad5-9659-1d09652b22be)

matthieu-foucault:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/368?planning-poker) with ZenHub @BCerki @dleard @gurjmatharu @pbastia

gurjmatharu:
I may be mistaken but for the acceptance criteria point:
```
When some of a task's forms were updated since the previous revision
And their status is staged
And they have validation errors
Then the status in the task list should be Attention Required
```
The status should be pending, no?  I don't think we can stage forms with validation errors.

gurjmatharu:
Also it looks like we're creating a project_contact form in status `pending` when we create a new project.  Is this expected behaviour?

matthieu-foucault:
[...]

PR #443, opened 2022-03-31
feat: fix dev data, use forms to populate

gurjmatharu:
#301


schema/data/dev/003_cif_project.sql
```diff
@@ -1,21 +1,29 @@
...
-        'test project ' || lpad(project_id::text, 3, '0'),
-        project_id * 1000
-        );
+        json_build_object(
+          'operatorId', '2',
+          'fundingStreamRfpId', '1',
+          'projectStatusId', '1',
+          'proposalReference', lpad(index::text, 3, '0'),
+          'summary', 'lorem ipsum dolor sit amet adipiscing eli',
+          'projectName', 'TEST-PROJECT-' || lpad(index::text, 3, '0'),
```
pbastia, line 22: having spaces in the project name might be more representative of the real data
```suggestion
          'projectName', 'Test Project ' || lpad(index::text, 3, '0'),
```

pbastia (changes requested):
Very small change on the project name, looks great otherwise! 👍

pbastia (approved):

pbastia (approved):

pbastia (approved):
just needs a rebase

schema/data/dev/003_cif_project.sql
```diff
@@ -1,21 +1,29 @@
 begin;
 
-
 do $$
   begin
-    for project_id in 1..50 loop
-      insert into cif.project(operator_id, funding_stream_rfp_id, project_status_id, proposal_reference, summary, project_name, total_funding_request) values
+    for index in 1..50 loop
+      insert into cif.form_change(
```
matthieu-foucault, line 6: This is closer to how projects are created when interacting with the application, but not quite the same. If this allows to edit existing projects it's good enough, but this may cause issues in the future. Using `cif.create_project()` would probably be the way to mirror the application's behaviour

gurjmatharu: Leaving as is for now - we can open up a tech debt card if needed.  Editing projects seems to work fine.

pbastia (approved):

pbastia (dismissed):

pbastia (commented):
just needs rebase!

schema/data/dev/005_cif_project_forms.sql
```diff
@@ -0,0 +1,48 @@
...
+        project_revision_id
+      )
+      values
+      (
+        json_build_object(
+          'contactId', index::text,
+          'projectId', index::text,
+          'contactIndex', '1'
+          ),
+        'create', 'cif', 'project_contact', 'committed', 'project_contact','1');
```
matthieu-foucault, line 26:
```suggestion
        'create', 'cif', 'project_contact', 'committed', 'project_contact', index);
```

schema/data/dev/005_cif_project_forms.sql
```diff
@@ -0,0 +1,48 @@
...
+        change_status,
+        json_schema_name,
+        project_revision_id
+      )
+      values
+      (
+        json_build_object(
+          'contactId', index::text,
+          'projectId', index::text,
+          'contactIndex', '1'
```
matthieu-foucault, line 24:
```suggestion
          'contactId', index,
          'projectId', index,
          'contactIndex', 1
```

schema/data/dev/005_cif_project_forms.sql
```diff
@@ -0,0 +1,48 @@
...
+          ),
+        'create', 'cif', 'project_contact', 'committed', 'project_contact','1');
+      insert into cif.form_change(
+        new_form_data,
+        operation,
+        form_data_schema_name,
+        form_data_table_name,
+        change_status,
+        json_schema_name
```
matthieu-foucault, line 33:
```suggestion
        json_schema_name,
        project_revision_id
```

schema/data/dev/005_cif_project_forms.sql
```diff
@@ -0,0 +1,48 @@
...
+      )
+      values
+      (
+        json_build_object(
+          'cifUserId', least(index, 6)::text,
+          'projectId', index::text,
+          'projectManagerLabelId', '1'
+          ),
+        'create', 'cif', 'project_manager', 'committed', 'project_manager');
```
matthieu-foucault, line 42:
```suggestion
        'create', 'cif', 'project_manager', 'committed', 'project_manager', index);
```

schema/data/dev/005_cif_project_forms.sql
```diff
@@ -0,0 +1,48 @@
...
+        form_data_table_name,
+        change_status,
+        json_schema_name
+      )
+      values
+      (
+        json_build_object(
+          'cifUserId', least(index, 6)::text,
+          'projectId', index::text,
+          'projectManagerLabelId', '1'
```
matthieu-foucault, line 40:
```suggestion
          'cifUserId', least(index, 6),
          'projectId', index,
          'projectManagerLabelId', 1
```

matthieu-foucault (changes requested):
With this current code, the project `form_change` won't have a `project_revision_id`. This will most likely lead to unexpected behaviour when we work on #368 , so we might want to change this after all to reduce confusion in the dev environment.

This means that
 - projects should be created using the `cif.create_project()` function. This will create a pending `form_change` for `project` and `project_contact`, which should later be updated with the correct `new_form_data`
 - projects should be created after contacts are created, i.e. `004_cif_contact` should become `003_cif_contact` and `003_cif_project` -> `004_cif_project`
 - the `project_manager` `form_change` should be created with the `pending` status, in `004_cif_project`
 - after creating/updating the pending form changes, all of the form revisions can be updated to `committed`

PR #447, opened 2022-04-02
chore: pull docker images through Artifactory

matthieu-foucault:
Todo: `lookup` doesn't return anything when doing a `helm template` (by design), so the helper needs to handle that to avoid throwing an exception when linting the chart

### DIFF
--- a/chart/cas-cif/templates/_helpers.tpl
+++ b/chart/cas-cif/templates/_helpers.tpl
@@ -74,3 +74,10 @@ Gets the suffix of the namespace. (-dev, -tools, ... )
 {{- define "cas-cif.namespaceSuffix" }}
 {{- (split "-" .Release.Namespace)._1 | trim -}}
 {{- end }}
+
+
+{{- define "cas-cif.imagePullSecrets" }}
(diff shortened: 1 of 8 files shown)


tpls.yaml
```diff
@@ -0,0 +1,18 @@
...
+...Successfully got an update from the "stable" chart repository
+...Successfully got an update from the "bitnami" chart repository
+Update Complete. ⎈Happy Helming!⎈
+Saving 5 charts
+Downloading cas-postgres from repo https://bcgov.github.io/cas-postgres/
+Downloading cas-airflow-dag-trigger from repo https://bcgov.github.io/cas-airflow
+Already downloaded cas-airflow-dag-trigger from repo https://bcgov.github.io/cas-airflow
+Already downloaded cas-airflow-dag-trigger from repo https://bcgov.github.io/cas-airflow
+Downloading nginx-sidecar from repo https://bcgov.github.io/cas-template-app
+Deleting outdated charts
```
pbastia, line 18: looks like this file made it in by accident?

Issue #369, opened 2022-03-16
As a Project Coordinator, I want to see which task in the task list is currently opened, so that understand which task I am looking at


matthieu-foucault:
#### Description:

This short user story merely adds highlighting to the `TaskList` component, based on the current route.
It simply need to use `useRouter()` to find the current route, and highlight the correct link depending on that.

#### Acceptance Criteria:

Given that I am viewing a project revision page
When I navigate to a form page, or the "check answers" page
Then the corresponding item in the task list is highlighted

#### Development Checklist:

- [ ] Meets the DOD

#### Notes:

![image.png](https://images.zenhubusercontent.com/575f1a78bf93871ccd9c8bcd/2232f898-7c36-4293-b4e7-146672823775)

matthieu-foucault:
@suhafa I tried some colors by looking at existing menus in BC government websites, but I'm not sure if that's the best option.
The example above uses an `#e5e5e5` background for the task list, and a `#fafafc` background for highlighted element (same as our page's background color). Does this look okay to you?

suhafa:
@matthieu-foucault  I experimented a bit with the layout of the task list with those colors and tried to find a way to incorporate them. The challenge is to keep it clean, and make the elements sit right (as opposed to looking like they are hovering). I attached a screenshot of a solution, which includes adding borders around the task list items, and spacing them out a bit more vertically. With this solution, I find it flows well with the colors, and that the 'grey-ing' out of inactive list items help visually by staying in the background. What do you think? 
<img width="460" alt="image" src="https://user-images.githubusercontent.com/99216375/159544164-c4ba55ff-b727-4f7f-9550-27083c2775da.png">

matthieu-foucault:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/369?planning-poker) with ZenHub @BCerki @dleard @gurjmatharu @pbastia @mikevespi

LindsayMacfarlane:
Looks awesome @BCerki ! Tested in dev by adding a new project and editing an existing project - confirmed that the project task list was highlighted depending on what step of the application I was on. Just confirming based on the attached wireframe - the status of each step was not part of this ticket (e.g., action required, not started, etc)?

BCerki:
Yes, the status is in this  other ticket: https://github.com/bcgov/cas-cif/issues/439

PR #448, opened 2022-04-04
Feat: highlight open task in task list

BCerki:
Issue https://github.com/bcgov/cas-cif/issues/369

I looked at the happo diffs and they look good to me, but held off approving so the reviewer can check too

### DIFF
--- a/app/components/TaskList.tsx
+++ b/app/components/TaskList.tsx
@@ -1,4 +1,5 @@
 import Link from "next/link";
+import { useRouter } from "next/router";
 import {
   getProjectRevisionContactsFormPageRoute,
   getProjectRevisionManagersFormPageRoute,
@@ -21,6 +22,24 @@ const TaskList: React.FC<Props> = ({ projectRevision }) => {
     `,
     projectRevision
   );
+  const router = useRouter();
+
+  const determineStepHighlight = (step) => {
+    if (!router || !router.pathname) return null;
+    if (router.pathname.includes(step)) {
+      return "highlight";
+    }
+    return null;
+  };
+
+  const determineSummaryHighlight = () => {
+    if (!router || !router.pathname) return null;
+    return `${
+      router.pathname === "/cif/project-revision/[projectRevision]"
+        ? "highlight"
+        : null
+    } bordered`;
+  };
 
   return (
     <div>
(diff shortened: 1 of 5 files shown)


app/components/TaskList.tsx
```diff
@@ -21,14 +22,32 @@ const TaskList: React.FC<Props> = ({ projectRevision }) => {
...
+  };
+
+  const determineSummaryHighlight = () => {
+    if (!router || !router.pathname) return null;
+    return `${
+      router.pathname === "/cif/project-revision/[projectRevision]"
+        ? "highlight"
+        : null
+    } bordered`;
+  };
```
matthieu-foucault, line 42:
```suggestion
  const currentStep = useMemo(() => {
    if (!router || !router.pathname) return null;
    if (router.pathname === "/cif/project-revision/[projectRevision]") return "summary";
    const routeParts = router.pathname.split("/");
    return routeParts[routeParts.length - 1]
  }, [router.pathname])
```

Then below you can simply check `currentStep === "overview"`, `currentStep === "managers"`, etc.
This will have the advantage to only be called once per page.

Using `getProjectRevisionPageRoute(id)` instead of hardcoding `"/cif/project-revision/[projectRevision]"` above would also be better, in which case you'd want to use `router.asPath` instead of `router.pathname`

app/components/TaskList.tsx
```diff
@@ -88,10 +122,17 @@ const TaskList: React.FC<Props> = ({ projectRevision }) => {
...
+        div :global(.highlight) {
+          background-color: #fafafc;
+        }
+        div :global(.bordered) {
+          border-bottom: 1px solid #d1d1d1;
+          padding: 10px 0 10px 0;
+        }
+
         div {
-          border: 2px solid black;
```
matthieu-foucault, line 92: I find it a bit weird to not have any border around the task list, @suhafa what do you think?

matthieu-foucault (changes requested):

app/components/TaskList.tsx
suhafa: Looks fine to me, the grey fill color acts like a block imo, so we don't really need to emphasize it with a border.

pbastia (approved):
nice!

matthieu-foucault (dismissed):
just needs a rebase

matthieu-foucault (approved):

PR #485, opened 2022-04-07
chore: use the crunchydata postgres operator

matthieu-foucault:


app/server/db/index.ts
```diff
@@ -14,7 +14,7 @@ export const getDatabaseUrl = () => {
 
   databaseURL += PGUSER;
   if (process.env.PGPASSWORD) {
-    databaseURL += `:${process.env.PGPASSWORD}`;
+    databaseURL += `:${encodeURIComponent(process.env.PGPASSWORD)}`;
```
matthieu-foucault, line 17: the passwords generated by the postgres operator may have special characters in them

chart/cas-cif/values.yaml
```diff
@@ -49,7 +30,7 @@ app:
 
 db:
   name: cif
-  appUser: cif_app
+  appUser: cifapp
```
matthieu-foucault, line 33: crunchydb doesn't like underscores in its user names

chart/cas-cif/templates/cronJobs/db-init.yaml
```diff
@@ -14,62 +12,35 @@ spec:
...
               image: {{ .Values.image.psql.repository }}:{{ default .Values.defaultImageTag .Values.image.psql.tag }}
               imagePullPolicy: {{ default .Values.defaultImagePullPolicy .Values.image.psql.pullPolicy }}
               resources:
                 limits:
-                  cpu: 1000m
+                  cpu: 50m
                   memory: 128Mi
                 requests:
-                  cpu: 100m
+                  cpu: 10m
```
pbastia, line 35: any risk in getting this so low?

pbastia (approved):
looks amazing 👌

chart/cas-cif/templates/cronJobs/db-init.yaml
matthieu-foucault: nah this just running psql and doesn't need any CPU

app/server/db/index.ts
pbastia: weren't we able to pull that URL from the secret generated by the operator? that might be a tech debt item to keep it compatible with local dev

pbastia (approved):

PR #504, opened 2022-04-14
chore: allow ingress from crunchy-monitor

dleard:
From the [how-to](https://github.com/bcgov/how-to-workshops/tree/master/crunchydb/monitoring)


matthieu-foucault (commented):
Should we do this in the `cas-pipeline` repo instead, and use an empty pod selector for the network policy? The only change that would be needed in this chart is adding the pgmonitor exporter

dleard:
> Should we do this in the `cas-pipeline` repo instead, and use an empty pod selector for the network policy? The only change that would be needed in this chart is adding the pgmonitor exporter

I could add the cluster name to the .values file and use that in the labels & pod selector

matthieu-foucault (approved):

matthieu-foucault:
@dleard this needs a rebase

matthieu-foucault (approved):

PR #505, opened 2022-04-18
chore: bump helm chart nginx-sidecar version

matthieu-foucault:


pbastia (approved):
We still need to merge that :) just needs a rebase

matthieu-foucault:
@pbastia we don't need it if we're going to use certbot instead of acme.sh. I was going to look into it today

matthieu-foucault:
Closing in favour of #573

PR #507, opened 2022-04-19
Feat/undo changes

BCerki:


app/lib/theme/widgets/MoneyWidget.tsx
```diff
@@ -29,7 +29,7 @@ export const MoneyWidget: React.FC<WidgetProps> = ({
         className="money"
         decimalScale={2}
         defaultValue={(schema as any).defaultValue}
-        value={value}
+        value={value || ""}
```
BCerki, line 32: If it's an empty string, it does show up in `newFormData` like this: 
Screenshot from 2022-04-19 10-40-18

app/components/Form/ProjectContactForm.tsx
```diff
@@ -177,17 +181,6 @@ const ProjectContactForm: React.FC<Props> = (props) => {
...
-        updateFormChange: {
-          formChange: {
-            ...formChange,
-            newFormData,
-            changeStatus: formChange.changeStatus,
-            // explicitely ignore the field below
-            projectRevisionByProjectRevisionId: undefined,
-          },
-        },
-      },
```
matthieu-foucault, line 190: Remind me why this needs to be removed?

matthieu-foucault (approved):
looks good overall. Needs a rebase and some conflict resolution

app/components/Form/ProjectContactForm.tsx
BCerki: I think because it wasn't doing anything (I don't remember why it wasn't working, though)

matthieu-foucault (changes requested):
Two issues I noticed:
 - The undo button doesn't remove project contacts that were added, or recreate contacts that were removed. You'll want to send mutations to delete `form_change` records with the `create` operation, and revert `form_changes` with the `archive` operation to be `update` instead
 - Adding the undo button brought back the issue of the page layout moving when the saving indicator changes. One minor change that Suha was suggesting is to simplify the saving indicator and remove the date and outline. You can do this in another PR just after this one

BCerki:
> Two issues I noticed:
> 
> * The undo button doesn't remove project contacts that were added, or recreate contacts that were removed. You'll want to send mutations to delete `form_change` records with the `create` operation, and revert `form_changes` with the `archive` operation to be `update` instead
> * Adding the undo button brought back the issue of the page layout moving when the saving indicator changes. One minor change that Suha was suggesting is to simplify the saving indicator and remove the date and outline. You can do this in another PR just after this one

Saving indicator PR will be here when ready: https://github.com/bcgov/cas-cif/pull/534

matthieu-foucault (approved):

PR #508, opened 2022-04-19
fix: postgres incremental backups should not run at 8am

matthieu-foucault:
Full backups are running at 8am, causing conflicts and jobs failing


mikevespi (approved):
Looks good! 🚢

PR #510, opened 2022-04-19
Feat/revision change reason

mikevespi:
- Added `change_reason` column to `project_revision` table.
- Added a text area to input the reason for a revision. Saves as you type using a debounced mutation. 
- Does not render if this revision is a project creation.
- If the revision is on an existing project, a change reason must be entered in order to submit the project revision.

Action shot:
changeReason

Not present on a new project: 
Screen Shot 2022-04-19 at 15 29 18

### DIFF
--- a/app/cypress/integration/cif/project-revision/index.spec.js
+++ b/app/cypress/integration/cif/project-revision/index.spec.js
@@ -275,7 +275,7 @@ describe("the new project page", () => {
     });
     cy.get("textarea").click().type("foo");
     // Allow the component to finish saving before taking screenshot
-    cy.wait(1500);
+    cy.contains("Changes saved").should("be.visible");
     cy.get("body").happoScreenshot({
       component: "Project Revision Summary",
       variant: "with_change_reason",
--- a/app/cypress/integration/cif/project-revision/index.spec.js
+++ b/app/cypress/integration/cif/project-revision/index.spec.js
@@ -274,6 +274,8 @@ describe("the new project page", () => {
       variant: "no_change_reason",
     });
     cy.get("textarea").click().type("foo");
+    // Allow the component to finish saving before taking screenshot
+    cy.wait(1500);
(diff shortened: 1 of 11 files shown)


app/cypress/integration/cif/project-revision/index.spec.js
```diff
@@ -268,6 +268,18 @@ describe("the new project page", () => {
...
       .should("have.text", "Not added");
 
+    cy.findByRole("button", { name: /submit/i }).should("be.disabled");
+    cy.get("body").happoScreenshot({
+      component: "Project Revision Summary",
+      variant: "no_change_reason",
+    });
+    cy.get("textarea").click().type("foo");
+    // Allow the component to finish saving before taking screenshot
+    cy.wait(1500);
```
dleard, line 278: We usually try to avoid cy.wait() if possible. Could the same thing be achieved with something like this:
`cy.get("textarea").contains("foo");`
or `cy.contains("Changes saved").should("be.visible");` 
to be sure the changes have saved before taking the screenshot?
cy commands are also 'thenable' if that doesn't work

schema/deploy/trigger_functions/commit_project_revision.sql
```diff
@@ -7,6 +7,10 @@ returns trigger as $$
 declare
 begin
 
+  if (new.change_status='committed') and (new.project_id is not null) and (new.change_reason is null) then
+    raise exception 'Cannot commit change with change_reason %', new.change_reason;
```
dleard, line 11: won't new.change_reason always be null in this exception? So the exception raised will print "cannot commit change with change_reason"
```suggestion
    raise exception 'Cannot commit change with null change_reason';
```

dleard (changes requested):
A couple of minor changes here, but otherwise looks good :+1:

dleard (approved):

PR #511, opened 2022-04-19
Feat/368 status of form in task list

gurjmatharu:
#368


matthieu-foucault (changes requested):
Had a quick look before a meeting, It'll need some more review. 

You'll want to add pgTap tests for `form_change_is_pristine`

app/components/Form/ProjectManagerFormSummary.tsx
```diff
@@ -17,7 +17,7 @@ const ProjectManagerFormSummary: React.FC<Props> = (props) => {
   const { projectManagerFormChangesByLabel } = useFragment(
     graphql`
       fragment ProjectManagerFormSummary_projectRevision on ProjectRevision {
-        projectManagerFormChangesByLabel {
+        projectManagerFormChangesByLabel(first: 500) {
```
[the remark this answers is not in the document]
gurjmatharu: Relay complains, that it doesn't match the same fragment in tasklist.

app/components/TaskList.tsx
```diff
@@ -30,6 +68,41 @@ const TaskList: React.FC<Props> = ({ projectRevision }) => {
...
+  const projectContactForms = tasklistProjectContactFormChanges.edges.map(
+    ({ node }) => node
+  );
+  const projectManagerForms = projectManagerFormChangesByLabel.edges
+    .filter(({ node }) => node)
+    .map(({ node }) => node.formChange);
+
+  const projectOverviewStatus = getStatus(projectForms);
+  const projectManagerStatus = getStatus(projectManagerForms);
+  const projectContactStatus = getStatus(projectContactForms);
```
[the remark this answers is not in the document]
gurjmatharu: Fixed - however not sure how to handle arrays.

app/components/TaskList.tsx
```diff
@@ -30,6 +59,50 @@ const TaskList: React.FC<Props> = ({ projectRevision }) => {
...
+  );
+  const projectManagerStatus = useMemo(
+    () => getStatus(projectManagerForms),
+    projectManagerForms
+  );
+  const projectContactStatus = useMemo(
+    () => getStatus(projectContactForms),
+    projectForms
+  );
+
```
matthieu-foucault, line 105: The code above won't do anything useful with `useMemo` as `projectForms`, `projectContactForms` and `projectManagerForms` will be new objects on every render. There's also a mistake with the dependency for `projectContactStatus` being `projectForms` instead of `projectContactForms`, ESlint should give you a warning about that, no?

something like this, with better formatting, is what you need here:

```suggestion
  const projectOverviewStatus = useMemo(
    () => getStatus([projectFormChange]),
    projectFormChange
  );
  const projectManagerStatus = useMemo(
    () => getStatus(projectManagerFormChangesByLabel.edges
    .filter(({ node }) => node)
    .map(({ node }) => node.formChange)),
    projectManagerFormChangesByLabel
  );
  const projectContactStatus = useMemo(
    () => getStatus(tasklistProjectContactFormChanges.edges.map(
    ({ node }) => node
  )),
    tasklistProjectContactFormChanges
  );

```

schema/deploy/mutations/add_contact_to_revision.sql
```diff
@@ -24,7 +25,8 @@ as $add_contact_form_change$
     'project_contact',
     $1,
     'pending',
-    'project_contact'
+    'project_contact',
+    '[{"message": "must have required property contactId"}]'
```
matthieu-foucault, line 29: This need a unit test (contacts added to a revision should have validation errors)

app/components/TaskList.tsx
```diff
@@ -39,7 +127,7 @@ const TaskList: React.FC<Props> = ({ projectRevision }) => {
   }, [id, router]);
 
   return (
-    <div>
+    <div style={{ width: "500px" }}>
```
matthieu-foucault, line 130: We should be able to have less than 500px for the width. It should also be specified in the `<style>` element at the bottom, not here.

app/components/TaskList.tsx
```diff
@@ -53,9 +141,20 @@ const TaskList: React.FC<Props> = ({ projectRevision }) => {
               aria-current={currentStep === "overview" ? "step" : false}
               className="bordered"
             >
-              <Link href={getProjectRevisionOverviewFormPageRoute(id)}>
-                <a>{mode === "update" ? "Edit" : "Add"} project overview</a>
-              </Link>
+              <div className="row">
```
matthieu-foucault, line 144: is this extra `div` needed? would a `ul > li` selector work instead?

app/components/TaskList.tsx
```diff
@@ -14,14 +14,79 @@ interface Props {
...
+  )
+    return "Incomplete";
+
+  if (
+    forms.some(
+      (form) =>
+        form.changeStatus === "staged" && form.validationErrors.length > 0
+    )
+  )
+    return "Attention Required";
```
matthieu-foucault, line 35:
```suggestion
    return <strong>Attention Required</strong>;
```

this will remove the need for conditional rendering below

app/components/TaskList.tsx
```diff
@@ -53,9 +141,20 @@ const TaskList: React.FC<Props> = ({ projectRevision }) => {
...
+                  <Link href={getProjectRevisionOverviewFormPageRoute(id)}>
+                    <a>{mode === "update" ? "Edit" : "Add"} project overview</a>
+                  </Link>
+                </div>
+                <div className="status">
+                  {projectOverviewStatus === "Attention Required" ? (
+                    <strong>{projectOverviewStatus}</strong>
+                  ) : (
+                    <>{projectOverviewStatus}</>
+                  )}
```
matthieu-foucault, line 155:
```suggestion
                  {projectOverviewStatus}
```

app/components/TaskList.tsx
```diff
@@ -145,9 +271,19 @@ const TaskList: React.FC<Props> = ({ projectRevision }) => {
...
+          display: flex;
+        }
+        .link {
+          flex: 50%;
+        }
+        .status {
+          padding-right: 10px;
+          text-align: right;
+          flex: 40%;
+        }
```
matthieu-foucault, line 286:
```suggestion
        ul > li {
          display: flex;
          justify-content: space-between;
        }
```

would this work?

matthieu-foucault (changes requested):
A bit of tweaking of the task list rendering and this should be good to go

app/components/TaskList.tsx
gurjmatharu: This causes padding/alignment issues.  I tried a few approaches, flex with row/col is the only way I got it to work
Screen Shot 2022-04-26 at 12 41 39 PM
.

schema/deploy/mutations/add_contact_to_revision.sql
gurjmatharu: Added!

matthieu-foucault (approved):

PR #513, opened 2022-04-21
refactor: form label is displayed by fields, not widgets

matthieu-foucault:
- render label in rjsf fields instead of widgets
- updated summary page to display `dt` and `dd` elements on the same row
- removed operator trade name from operator selection widget
- use rjsf's `ui:help` instead of a custom option


pbastia (approved):
Looks great!!

Issue #397, opened 2022-03-22
As a project coordinator, I want the answers summary to only display the differences since the last revision, so that I can focus on what did change


matthieu-foucault:
#### Description:

Rather than showing the entirety of the forms, the summary page should only display the fields that were changed.
This means that a `diff` between each `form_change` and its parent needs to be performed, and only those attributes that changed should be displayed.

Given that the objects that are compared are flat JSON objects whose structure will not change between revisions, we may not need a diff library.
The cases where contacts or managers are added or removed will be easily identifiable by looking at the `form_change`'s `operation`, so this use case is easier than if we were comparing two arbitrary project revisions

#### Acceptance Criteria:

Given that I am editing a project, and that I only updated a given set of fields
When review my answers
Then only the changes of the updated fields are displayed

Given that I am editing a project, and that I did not update a given form
When review my answers
Then the page should display `<form name> not updated` for the forms that did not change.

Given that I am editing a project, and that I added a secondary contact or manager
When review my answers
Then the page should display `Added <manager name>` or `Added <contact name>`

Given that I am editing a project, and that I removed a secondary contact or manager
When review my answers
Then the page should display `Removed <manager name>` or `Removed <contact name>`

#### Development Checklist:

- [ ] Meets the DOD

#### Wireframe:

![image](https://user-images.githubusercontent.com/2684022/161814001-f5397451-d7e0-4ff4-be5f-16345f367e72.png)

matthieu-foucault:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/397?planning-poker) with ZenHub @BCerki @dleard @gurjmatharu @mikevespi @pbastia @Sepehr-Sobhani

LindsayMacfarlane:
Thanks @dleard - this is fantastic!

Tested in dev by:
- Editing an existing project and making changes to total funding request and project status. I was clearly able to see what I edited and in which sections I made no changes. 
- Confirmed changes reflected in project lists

PR #533, opened 2022-04-27
Feat: summary diff

dleard:
issue #397

Screenshot from 2022-04-29 16-18-57


app/components/Form/ProjectFormSummary.tsx
```diff
@@ -4,112 +4,133 @@ import readOnlyTheme from "lib/theme/ReadOnlyTheme";
...
+    for (const [key, value] of Object.entries(projectFormChange.newFormData)) {
+      if (
+        value ===
+        projectFormChange.formChangeByPreviousFormChangeId?.newFormData?.[key]
+      )
+        delete filteredSchema.properties[key];
+      else newDataObject[key] = value;
+    }
+    return newDataObject;
+  }, [
```
pbastia, line 92: I don't think we need to keep in memory a copy of the original schema/formData and a copy of the new one. we can, depending on the `isFirstRevision` property, generate the data to use:
```tsx
const [formSchema, formData]  = useMemo( () => {
  if(isFirstRevision) return [projectSchema, projectFormChange.newFormData];

  const filteredSchema = JSON.parse(JSON.stringify(projectSchema));
  ...etc...
  return [filteredSchema, newDataObject]
}
```

This way we don't have logic in the JSX further down

app/lib/theme/CustomFields.tsx
```diff
@@ -0,0 +1,212 @@
...
+import { faLongArrowAltRight } from "@fortawesome/free-solid-svg-icons";
+
+const showStringDiff = (id: string, oldData: string, newData: string) => (
+  <>
+    <span id={id && `${id}-diffOld`} className="diffOld">
+      {oldData}
+    </span>
+    &nbsp;
+    <FontAwesomeIcon size="lg" color="black" icon={faLongArrowAltRight} />
+    &nbsp;
```
pbastia, line 14: I think we're not supposed to use `&nbsp` as part of layout instructions, we should use a style with margins on the `<FontAwesomeIcon ... />` instead

app/lib/theme/CustomFields.tsx
```diff
@@ -0,0 +1,212 @@
...
+    </span>
+    &nbsp;
+    <span>
+      <strong>
+        <em>REMOVED</em>
+      </strong>
+    </span>
+  </>
+);
+
```
pbastia, line 125: We could refactor the different display `<span> {data} / <NUMBER> / ADDED </span>` to avoid repeating them, but I kind of like how readable it is as is

app/lib/theme/CustomFields.tsx
```diff
@@ -0,0 +1,212 @@
...
+          id,
+          formContext?.oldData?.[props.name],
+          uiSchema?.["ui:widget"] === "MoneyWidget"
+        );
+      } else {
+        return <>DISPLAY ERROR</>;
+      }
+    }
+  },
+};
```
pbastia, line 210: This is some pretty intricate display logic, and could definitely use a few unit tests

app/lib/theme/CustomFields.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+export default CUSTOM_FIELDS;
```
pbastia, line 212: Should we name this one `DiffingFields` or something like that?

pbastia (changes requested):
Nice work! And nice use of rjsf custom fields

dleard:
closed in favor of #572

PR #534, opened 2022-04-27
style: remove outline and date from SavingIndicator

BCerki:


app/components/Form/SavingIndicator.tsx
```diff
@@ -6,15 +6,9 @@ interface Props {
   lastEdited: Date;
```
matthieu-foucault, line 6: you should remove this prop, and edit the places where this component is used. There's probably a fair amount of code that can be removed here

matthieu-foucault (changes requested):

app/components/Form/ProjectContactForm.tsx
```diff
@@ -100,13 +100,6 @@ const ProjectContactForm: React.FC<Props> = (props) => {
     props.query
   );
 
-  const lastEditedDate = useMemo(() => {
-    const mostRecentUpdate = projectContactFormChanges.edges
-      .map((e) => e.node.updatedAt)
-      .sort((a, b) => Date.parse(b) - Date.parse(a))[0];
-    return new Date(mostRecentUpdate);
-  }, [projectContactFormChanges]);
-
```
matthieu-foucault, line 113: now that this is removed, your linter should show some warnings on the `updatedAt` field in your relay fragment above, telling you it's not used. It can be removed. Maybe do a search for `updatedAt` in our tests too, a bunch of them might need to be removed from the mock query payloads

app/components/Form/SavingIndicator.tsx
```diff
@@ -27,7 +20,6 @@ const SavingIndicator: React.FC<Props> = ({ isSaved, lastEdited }) => {
             display: inline-flex;
             justify-content: flex-end;
             align-items: baseline;
-            border: solid 1px #939393;
             border-radius: 4px;
```
matthieu-foucault, line 23:
```suggestion
```

app/components/Form/SavingIndicator.tsx
```diff
@@ -27,7 +20,6 @@ const SavingIndicator: React.FC<Props> = ({ isSaved, lastEdited }) => {
...
             padding: 0.5rem;
           }
```
matthieu-foucault, line 25: the `.last-edited` style below can be removed too

matthieu-foucault (changes requested):

matthieu-foucault (approved):

PR #535, opened 2022-04-28
Feat: Document upload to gcp

pbastia:


app/config.js
```diff
@@ -149,17 +149,15 @@ var config = convict({
...
-  storageApiKey: {
-    // Not yet used in deployments, so no validation function yet.
-    doc: "Api key allowing access to storage api, needs to match what's in storage api env.",
-    default: "_doNotUseThisApiKeyInProduction_",
-    env: "STORAGE_API_KEY",
+  gcsCredentials: {
+    doc: "Google Cloud Storage credentials.",
+    default: "/credentials/credentials.json",
+    env: "GOOGLE_APPLICATION_CREDENTIALS",
+  },
```
pbastia, line 156: I'm wondering if this is necessary to provide a default value here, we'd want an alert if the variable is not set

matthieu-foucault: You'd want to validate that the file exists I guess, this could be left as the default.

app/server/middleware/attachmentDownloadRouter.ts
```diff
@@ -1,60 +1,54 @@
...
-    apiResult.headers.get("fastapi-content-length")
-  );
 
-  (apiResult.body as any).pipe(res);
-});
+  await storageClient
+    .bucket(bucketName)
+    .file(file)
+    .createReadStream()
+    .pipe(res);
```
matthieu-foucault, line 49: Do we need a `try/catch` and a call to `next(error)` if an exception is thrown? `express@4` doesn't handle exceptions in async middlewares

app/config.js
matthieu-foucault: I don't see this config item being used in the code below

app/tests/unit/server/middleware/attachmentDownloadRouter.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+      attachment: {
+        file: "fileUuid",
+        fileName: "testFileName.extension",
+        fileType: "mime/testfiletype",
+      },
+    },
+  } as any;
+});
+
+describe("The attachment download router", () => {
```
matthieu-foucault, line 41: can we add a test case that throws an error?

app/tests/unit/server/middleware/attachmentDownloadRouter.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+  beforeEach(() => {
+    jest.restoreAllMocks();
+  });
+  it("is configured properly", () => {
```
matthieu-foucault, line 45: can we be more explicit here?

matthieu-foucault (changes requested):

app/config.js
pbastia: It's being used implicitely by the `@google-cloud/storage` package

matthieu-foucault: can we add that to the docs?

app/tests/unit/server/middleware/attachmentDownloadRouter.test.tsx
pbastia: oops, yes 😁

matthieu-foucault (approved):

Issue #499, opened 2022-04-13
Validation error on funding request


gurjmatharu:
Describe the Bug:

Probability (how likely the bug is to happen, scored from 1-5): 5
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 2
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:
Go to a project and edit the project.
Navigate to `Edit project overview`
Ensure total funding amount has already been populated, and scroll down and `Submit Project Overview`
You'll see a validation error.

Screenshots:
![Screen Shot 2022-04-13 at 4.35.26 PM.png](https://images.zenhubusercontent.com/620bf57daa1ab92954017ec6/2dfebc34-7a62-4a68-89d6-592a2d4cb7e5)

Additional information:

- OS: [e.g. iOS] macOS
- Browser: [e.g. chrome, safari] chrome
- Version: [e.g. 22] develop?
- Additional context

matthieu-foucault:
@BCerki I believe this was fixed in #507 ?

LindsayMacfarlane:
Thank you @BCerki! I definitely noticed this bug while testing yesterday. 

Tested in dev by:
1) Creating a new record and adding funding amount. Received no error when saving. 
2) Editing an existing project, made a change to project status (did not change funding amount), submitted record and received no error message.

PR #536, opened 2022-04-28
Fix bug with total funding validation

BCerki:
Issue https://github.com/bcgov/cas-cif/issues/499
I manually tested Gurjeet's example and also tried making my own project to confirm the bug was only happening with the dev data.


matthieu-foucault (dismissed):
just needs a rebase

matthieu-foucault (approved):

Issue #396, opened 2022-03-22
As a project coordinator, I want the "check answers" page to display errors, so that I can fix them before submitting my project revision


matthieu-foucault:
#### Description:

The check answers summary page should clearly show any errors present in the revision before submitting. The errors should be shown in the same style as defined in the [fieldTemplate component](https://github.com/bcgov/cas-cif/blob/develop/app/lib/theme/FieldTemplate.tsx).

During this work, it may become apparent that we need some custom validation functions in the database to handle more complicated errors. There is an example of how this was done in CIIP. Database examples are [here](https://github.com/bcgov/cas-ciip-portal/tree/develop/schema/deploy/application_validation_functions). An example of the implementation of these custom validation functions can be found [here](https://github.com/bcgov/cas-ciip-portal/blob/develop/app/containers/Applications/ApplicationWizardConfirmation.tsx#L203).

The error messages should make sense to the user. It should be clear what is wrong with the field and how to fix it. Some of the default error messages are not very helpful and can be overridden with rjsf's customTransformErrors. Example of this is [here](https://github.com/bcgov/cas-cif/blob/develop/app/lib/theme/customTransformErrors.ts).

#### Acceptance Criteria:

Given I have errors in my project revision
When I navigate to the answers summary
Then I can clearly see what fields have errors

Given I have errors in my project revision
When I read the errors on the answers summary
Then the error messages are easily understood as to what is wrong

Given I have errors in my project revision
When I am on the answers summary
Then submission of the revision is blocked

#### Development Checklist:

- [ ] Answers summary shows error messages below the corresponding field, styled to match the [fieldTemplate component](https://github.com/bcgov/cas-cif/blob/develop/app/lib/theme/FieldTemplate.tsx).
- [ ] Any unclear error messages are overridden to contain language that the user can understand
- [ ] Custom validation functions exist for more complicated errors (if necessary)
- [ ] Errors prevent submission of the revision
- [ ] Meets the DOD

pbastia:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/396?planning-poker) with ZenHub @BCerki @dleard @gurjmatharu @matthieu-foucault @mikevespi

LindsayMacfarlane:
Thank you @gurjmatharu! A reminder to check-off the checklist items to help make it clear that all the steps were completed.

I tested in dev by:
[...]

PR #553, opened 2022-05-02
Feat/396 check answers page display errors

gurjmatharu:
#396

### DIFF
--- a/app/components/Form/ProjectContactFormSummary.tsx
+++ b/app/components/Form/ProjectContactFormSummary.tsx
@@ -123,11 +123,9 @@ const ProjectContactFormSummary: React.FC<Props> = (props) => {
             form: primaryContact?.node,
           }}
         />
-        {<label>Secondary Contacts</label>}
+        <label>Secondary Contacts</label>
         {secondaryContacts.length === 0 ? (
-          <dd>
-            <em>Secondary contact not added</em>
-          </dd>
+          <em>No secondary contacts</em>
         ) : (
           contactsJSX
         )}
--- a/app/components/Form/ProjectContactFormSummary.tsx
+++ b/app/components/Form/ProjectContactFormSummary.tsx
@@ -125,9 +125,9 @@ const ProjectContactFormSummary: React.FC<Props> = (props) => {
         />
         <label>Secondary Contacts</label>
         {secondaryContacts.length === 0 ? (
-          <div>
+          <dd>
             <em>No secondary contacts</em>
-          </div>
+          </dd>
         ) : (
           contactsJSX
         )}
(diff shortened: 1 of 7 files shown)


app/components/Form/ProjectContactFormSummary.tsx
```diff
@@ -104,44 +102,36 @@ const ProjectContactFormSummary: React.FC<Props> = (props) => {
...
+          formData={primaryContactNode?.node.newFormData}
+          formContext={{
+            query: props.query,
+            form: primaryContact?.node,
+          }}
+        />
+        {<label>Secondary Contacts</label>}
+        {secondaryContacts.length === 0 ? (
+          <dd>
+            <em>Secondary contact not added</em>
```
matthieu-foucault, line 129:
```suggestion
            <em>No secondary contacts</em>
```

app/components/Form/ProjectContactFormSummary.tsx
```diff
@@ -104,44 +102,36 @@ const ProjectContactFormSummary: React.FC<Props> = (props) => {
...
           )}
-        </>
-      )}
+          formData={primaryContactNode?.node.newFormData}
+          formContext={{
+            query: props.query,
+            form: primaryContact?.node,
+          }}
+        />
+        {<label>Secondary Contacts</label>}
```
matthieu-foucault, line 126:
```suggestion
        <label>Secondary Contacts</label>
```

app/components/Form/ProjectFormSummary.tsx
```diff
@@ -113,7 +114,14 @@ const ProjectFormSummary: React.FC<Props> = (props) => {
           }}
         />
       )}
-    </>
+      <style jsx>
+        {`
+          div :global(dd) {
+            margin-bottom: 2px;
```
matthieu-foucault, line 120: could we set this style on the widgets themselves rather than making it global. This would ensure consistency across forms

app/components/Form/ProjectContactFormSummary.tsx
```diff
@@ -104,44 +102,36 @@ const ProjectContactFormSummary: React.FC<Props> = (props) => {
...
+        {secondaryContacts.length === 0 ? (
+          <dd>
+            <em>Secondary contact not added</em>
+          </dd>
```
matthieu-foucault, line 130:
```suggestion
          <em>Secondary contact not added</em>
```
I think the `dd` is what is causing the accessibility issue in the e2e tests

matthieu-foucault (changes requested):
You'll need to undo the changes to the schema, as they were not needed, and updated your e2e tests, but this looks good

app/components/Form/ProjectFormSummary.tsx
gurjmatharu: Gave this a go - having trouble with rending this in other components.

app/components/Form/ProjectContactFormSummary.tsx
gurjmatharu: I've opened a new issue: #563 this effects multiple forms.

Issue #497, opened 2022-04-13
As a Project Coordinator, I want the submit contacts button to be outside of the box so that the interface is consistent with other pages.


elliottSlee:
#### Description:

During usability testing it became apparent that the submit contacts button is inside the box, but in other sections (Project Overview & Project Managers) the submit button is outside the box. It would be good to have this be consistent.

#### Acceptance Criteria:

Given that a user is on the project managers page of adding a contact
When they see the submit button
Then the button will be displayed outside of the box (frame)

#### Development Checklist:

- [x] Move submit button to outside of frame
- [ ] Meets the DOD

#### Notes:

-

LindsayMacfarlane:
Even though this is a low priority - this one really annoys me and would be good to fix. By implementing this ticket we would improve UI for user and provide better consistency between forms.

matthieu-foucault:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/497?planning-poker) with ZenHub @BCerki @dleard @gurjmatharu @pbastia @Sepehr-Sobhani

LindsayMacfarlane:
Thanks @Sepehr-Sobhani ! I tested in dev by:
1) creating a project and confirming that the submit button is outside of box
2) editing a project and confirming that the submit button is outside of the box

Friendly reminder to check-off the items listed under development checklist.

PR #564, opened 2022-05-04
style: move submit contact button out of projectcontactform border

Sepehr-Sobhani:
#497


matthieu-foucault (changes requested):
just needs to be rebased

matthieu-foucault (approved):

PR #565, opened 2022-05-04
Quarterly reports

BCerki:
Issue https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/456


app/components/Form/ProjectQuarterlyReportsForm.tsx
```diff
@@ -0,0 +1,308 @@
...
+                  <FontAwesomeIcon icon={faPlusCircle} /> Add another quarterly
+                  report
+                </Button>
+              </Grid.Row>
+              {projectRevision.projectQuarterlyReportFormChanges.edges.map(
+                ({ node }, index) => {
+                  return (
+                    <Grid.Row key={node.id} className="reportContainer">
+                      <Grid.Col span={6}>
+                        <h3>Quarterly Report {index + 1}</h3>
```
BCerki, line 237: Do we want the reports to renumber if one is deleted?

BCerki:
Scheduling a meeting with @suhafa and @pbastia to discuss naming and ordering of quarterly reports (currently they bounce around and change names)

app/tests/unit/components/Form/ProjectReportingRequirementsForm.test.tsx
```diff
@@ -0,0 +1,266 @@
...
+import compiledProjectQuarterlyReportFormQuery, {
+  quarterlyReportsFormQuery,
+} from "__generated__/quarterlyReportsFormQuery.graphql";
+import { ProjectQuarterlyReportForm_projectRevision } from "__generated__/ProjectQuarterlyReportForm_projectRevision.graphql";
+
+const testQuery = graphql`
+  query ProjectReportingRequirementsFormQuery @relay_test_operation {
+    query {
+      # Spread the fragment you want to test here
+      projectRevision(id: "Test Project Revision ID") {
```
BCerki, line 15: Using "Test Project Revision ID" seems to work fine, but I don't understand its relationship to `defaultMockResolver` below

app/tests/unit/components/Form/ProjectReportingRequirementsForm.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+            __id: "client:mock:__connection_projectQuarterlyReportFormChanges_connection",
+          },
+        };
+        return result;
+      },
+    };
+    componentTestingHelper.loadQuery(mockResolver);
+    componentTestingHelper.renderComponent();
+
+    userEvent.type(screen.getByLabelText(/General Comments \(optional\)/), "c");
```
BCerki, line 228: If I tried to type anything longer than one character, letters would be missing in the `expect`. I tried a few versions of wrapping things in `act()` but no dice

pbastia: I think it's because the mutation fails to return complete data (we can confirm that on Monday) and the optimistic update might get in the way. So on each character type, the change event is fired and the react component re-renders with the initial, empty state from the mocked resolver.

As an alternative, you can artificially fire a `change` event on the input:
```tsx
    fireEvent.change(screen.getByLabelText(/General Comments \(optional\)/), {
      target: { value: "some test comment" },
    });
```

BCerki: Yes, our change to the optimistic response solved this

app/cypress/support/commands.js
```diff
@@ -190,3 +190,49 @@ Cypress.Commands.add(
     cy.get(`input[value="${secondaryContact}"]`).should("be.visible");
   }
 );
+
+Cypress.Commands.add(
+  "fillReport",
```
pbastia, line 195: could be maybe more explicit, "fillQuarterlyReport"

schema/deploy/computed_columns/project_revision_project_quarterly_report_form_changes.sql
```diff
@@ -0,0 +1,23 @@
+-- Deploy cif:computed_columns/project_revision_project_quarterly_report_form_changes to pg
+
+begin;
+
+create or replace function cif.project_revision_project_quarterly_report_form_changes(project_revision cif.project_revision)
```
pbastia, line 5: This could eventually be refactored into a single computed column:
```sql
cif.project_revision_form_changes_for(cif.project_revision, table_name text, json_matcher jsonb default '{}')
```
or something similar, to avoid duplicating computed columns every time we add a new form

app/components/TaskList.tsx
```diff
@@ -144,7 +170,7 @@ const TaskList: React.FC<Props> = ({ projectRevision }) => {
                 "summary",
                 getProjectRevisionPageRoute(id),
                 "Review and submit information",
-                null
+                "sssssafawef"
```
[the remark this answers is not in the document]
pbastia: oh no >.<

app/cypress/integration/cif/project-revision/index.spec.js
```diff
@@ -147,7 +182,7 @@ describe("the new project page", () => {
     cy.checkContactsForm("", "");
   });
 
-  it("Allows to create and update a project", () => {
+  it.only("Allows to create and update a project", () => {
```
matthieu-foucault, line 185:
```suggestion
  it("Allows to create and update a project", () => {
```

app/data/jsonSchemaForm/validationSchemas.ts
```diff
@@ -7,6 +8,8 @@ import operatorSchema from "./operatorSchema";
 const validationSchemas = {
   project_contact: projectContactSchema,
   project_manager: projectManagerSchema,
+  //brianna --why is it not project_reporting_requirement?
```
matthieu-foucault, line 11: because that's the name of the table

matthieu-foucault (changes requested):
minor cleanup needed, doing some manual testing while that's done

app/components/Form/ProjectQuarterlyReportForm.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+              <Grid.Row>
+                <Button
+                  size="medium"
+                  variant="primary"
+                  onClick={stageQuarterlyReportFormChanges}
+                  disabled={isUpdating}
+                >
+                  Submit Quarterly Reports
+                </Button>
```
matthieu-foucault, line 287: This should be moved outside of the form border, to match other pages

app/components/Form/ProjectQuarterlyReportForm.tsx
```diff
@@ -0,0 +1,315 @@
...
+    };
+    addQuarterlyReportMutation({
+      variables: {
+        projectRevisionId: projectRevision.rowId,
+        newFormData: formData,
+        connections: [projectRevision.projectQuarterlyReportFormChanges.__id],
+      },
+      onError: (error) => {
+        console.log(error);
+      },
```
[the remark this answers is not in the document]
BCerki: I go this from the project manager component; I can remove it there too

app/components/Form/ProjectQuarterlyReportForm.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+  return (
+    <div>
+      <header>
+        <h2>Quarterly Reports</h2>
+
+        <SavingIndicator isSaved={!isUpdating && !isAdding} />
+      </header>
+
+      <Grid cols={10} align="center">
```
matthieu-foucault, line 219: I'm not a fan of using a grid. Its only use in this file is to display the `h3` and "remove" button on the same row, which can be done easily, better, and with much less code with a simple flexbox

app/components/Form/ProjectQuarterlyReportForm.tsx
```diff
@@ -0,0 +1,315 @@
...
+    "ui:widget": "date",
+  },
+  submittedDate: {
+    "ui:col-md": 12,
+    "bcgov:size": "small",
+    "ui:widget": "date",
+  },
+  comments: {
+    "ui:col-md": 12,
+    "bcgov:size": "small",
```
matthieu-foucault, line 38: the `comments` should use the text area widget

app/data/jsonSchemaForm/validationSchemas.ts
BCerki: Sorry, that was a confusing way to ask that question. I mean I'm not sure I'm using the `project` prefix right. I called all my variables `projectReportingRequirement`, `projectQuarterlyReports`, (vs. `reportingRequirement, quarterlyReports) because I figured the prefix meant the report was tied to a project (like how projectContacts are just for a project, but contacts are their own page). But if the name is supposed to be coming from the table, then maybe I should change it?

matthieu-foucault: Yes, I suppose that the `project` prefix should be removed when there is no ambiguity.

matthieu-foucault: Although one could think that naming a component `QuarterlyReportForm` is confusing, if all of the other project forms start with `Project`

BCerki: Yes, that's where I was initially coming from. If you're only looking at front-end files, it looks inconsistent. Perhaps I'll leave it as-is then?

matthieu-foucault: Yeah I think we can leave as-is for now

app/components/Form/ProjectQuarterlyReportForm.tsx
```diff
@@ -0,0 +1,299 @@
...
+      <style jsx>{`
+        div :global(button.pg-button) {
+          margin-left: 0.4em;
+          margin-right: 0em;
+        }
+        div :global(.right-aligned-column) {
+          display: flex;
+          justify-content: flex-end;
+          align-items: flex-start;
+        }
```
matthieu-foucault, line 281: is this used?

app/components/Form/ProjectQuarterlyReportForm.tsx
```diff
@@ -0,0 +1,299 @@
...
+        div :global(.reportContainer) {
```
matthieu-foucault, line 282:
```suggestion
        div.reportContainer {
```

app/components/Form/ProjectQuarterlyReportForm.tsx
BCerki: Nope! I'll also remove it in the place I borrowed it from that it isn't used

matthieu-foucault (approved):

pbastia (approved):

PR #568, opened 2022-05-05
Automatically stage all form changes that are not pristine

matthieu-foucault:
This PR:
 - add an effect that will stage all dirty form changes when the tasklist changes pages
 - fix the Money widget to prevent it from having a NaN value
 - initializes the project form_change to `null` so that its status is initially "Not" started
 - makes the tasklist sections expanded by default when they contain an "attention required" status


app/components/Form/ProjectAnnualReportForm.tsx
```diff
@@ -56,10 +56,6 @@ const ProjectAnnualReportForm: React.FC<Props> = (props) => {
               newFormData
               operation
               changeStatus
-              formChangeByPreviousFormChangeId {
```
dleard, line 59: Why did this get removed? Was it not used? It will be used after Gurjeet's PR makes it in since it's used in diffing.

app/components/Form/ProjectMilestoneReportForm.tsx
```diff
@@ -77,15 +77,10 @@ const ProjectMilestoneReportForm: React.FC<Props> = (props) => {
           __id
           edges {
             node {
-              rowId
               id
               newFormData
               operation
               changeStatus
-              formChangeByPreviousFormChangeId {
```
dleard, line 85: Same comment as above re: diffing

dleard (commented):
Nice! Just a couple pieces of code being removed that I think may cause merge conflicts with Gurjeet and Sepehr's PR's whre they get added back in

dleard (dismissed):
:shipit:

dleard:
There's a failing e2e test, but otherwise looks good to go. Looks like the error is in the contacts part of the test

Sepehr-Sobhani:
> There's a failing e2e test, but otherwise looks good to go. Looks like the error is in the contacts part of the test

@matthieu-foucault Is this the same issue as: https://github.com/bcgov/cas-cif/issues/684 ?

matthieu-foucault:
> > There's a failing e2e test, but otherwise looks good to go. Looks like the error is in the contacts part of the test
> 
> @matthieu-foucault Is this the same issue as: #684 ?

@Sepehr-Sobhani  it looks like it yes, the e2e test passed after rebasing. I added some extra assertions in the test between adding each secondary contact, it might fix it.

dleard (approved):

PR #570, opened 2022-05-05
chore: make project list headings like an accordion

Sepehr-Sobhani:
Had to do a couple of commit message rewording due to conflicts but I don't really know what happened to those commits. Anyways conflicts are resolved and I think everything is good 😄


app/components/TaskList.tsx
```diff
@@ -156,36 +127,13 @@ const TaskList: React.FC<Props> = ({ projectRevision }) => {
...
           text-indent: 15px;
         }
-        h3 {
-          font-size: 1rem;
-          line-height: 1;
-          border-bottom: 1px solid #d1d1d1;
-          text-indent: 15px;
-          padding: 10px 0 10px 0;
-          margin: 0;
-        }
```
matthieu-foucault, line 188: I think the styles below for the links are not needed if we use the `Link` components from the BC gov theme.
See the `Dashboard` component, which uses the `BCGovLink` component

app/components/TaskListItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+  listItemName: string;
+  listItemMode?: string;
+  formListItemArray: Array<{
+    stepName: string;
+    linkUrl: { pathname: string; query: { projectRevision: string } };
+    formTitle: string;
+    formStatus: string | JSX.Element;
+  }>;
+  currentStep: string;
+  mode: string;
```
matthieu-foucault, line 19:
```suggestion
  mode: "view" | "update" | "create";
```

app/components/TaskListItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+    index,
+    stepName,
+    linkUrl,
+    formTitle,
+    formStatus,
+  }) => {
+    return (
+      <li
+        aria-current={currentStep === stepName ? "step" : false}
+        className="bordered"
```
matthieu-foucault, line 44: is this class name having any effect?

app/components/TaskListItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+        key={index}
+      >
+        <Link href={linkUrl}>
+          <a>
```
matthieu-foucault, line 48: This is where you should use a `BCGovLink`

matthieu-foucault (changes requested):

app/components/TaskListItem.tsx
Sepehr-Sobhani: Actually, I didn't touch that one but yeah it adds borders at the bottom of each task list item.
Screen Shot 2022-05-10 at 4 01 51 PM

matthieu-foucault (dismissed):
Beautiful. Just needs to be rebased

matthieu-foucault (approved):

Issue #558, opened 2022-05-03
Adjacent fragments are causing data in the relay store to be overwritten (TaskList)


dleard:
### Describe the Bug:

Because we are retrieving different `form_change` data in adjacent fragments on `projectRevision` and rendering the components asking for that data together on one page, some of the fragment data in the query is overwriting data from the other fragments in the relay store. Specifically, it's the `TaskList` fragment that is requesting `Project`, `ProjectContact` and `ProjectManager` `form_change` data alongside the summary and form components requesting different `form_change` data (screenshot below).

This is also causing console warnings about parsing the data with the same query (screenshot below). The same error is printed several times when running `yarn test`.

#### Affected files:
`pages/cif/project-revision/[projectRevision]/index.tsx`
`pages/cif/project-revision/[projectRevision]/form/contacts.tsx`
`pages/cif/project-revision/[projectRevision]/form/managers.tsx`
`pages/cif/project-revision/[projectRevision]/form/overview.tsx`

Probability (how likely the bug is to happen, scored from 1-5): 4
Effect (how bad the bug is when it does happen, scored from 1-5): 2
(The Effect value will likely get worse as we add forms & functionality, but it is not currently causing any major issue in how the app works)

### Steps to reproduce the behaviour:

We are handling this currently by defining connections with different `first: x` values on the different queries where we are fetching data from the same form_changes.
eg:
TaskList has a projectContactFormChanges with a connection requesting first: 500
ProjectContactForm has a projectContactFormChanges with a connection requesting first: 750
ProjectContactFormSummary has a projectContactFormChanges with a connection requesting first: 1000

This appears to split the connections so that the data is not overwritten for now n our tests. One way to reproduce the bug and see what's going on is to remove these connections (or even just remove make the `first` argument equal across all connections) and run `yarn test`. Some tests will now fail. If you change the order of the fragments in `pages/cif/project-revision/[projectRevision]/index.tsx` you'll notice that different tests pass and fail, depending on the order of how the  fragments were defined.

If you run the app, open the `relay dev tools` & try changing up the `first` arguments to be the same/different you'll see that the `projectContactFormChanges` connections split into 3 when `first` is different and otherwise overwrite each other. The key of the connections appear to have no effect on this, only the value of the `first` argument on the connection.

### Suggested Solution:

[...]

PR #571, opened 2022-05-05
Fix: fragment adjacency

dleard:
issue #558


schema/deploy/computed_columns/project_revision_project_contacts_status.sql
```diff
@@ -0,0 +1,29 @@
...
+    case
+      when (select exists (select get_form_status from cif.get_form_status($1.id, 'project_contact') where get_form_status = 'Incomplete'))
+        then 'Incomplete'
+      when (select exists (select get_form_status from cif.get_form_status($1.id, 'project_contact') where get_form_status = 'Attention Required'))
+          then 'Attention Required'
+      when (select count(distinct get_form_status) from cif.get_form_status($1.id, 'project_contact')) = 1
+        and (select (select distinct get_form_status from cif.get_form_status($1.id, 'project_contact'))) = 'Complete'
+          then 'Complete'
+      else 'Not Started'
+    end;
```
matthieu-foucault, line 21: should we use 
```sql
with form_status as (
  select * from cif.get_form_status($1.id, 'project_contact')
)
```
to make this a bit more readable? Not sure if that's improving performance, I assume postgres optimizes this.

schema/deploy/util_functions/get_form_status.sql
```diff
@@ -0,0 +1,35 @@
+-- Deploy cif:util_functions/get_form_status to pg
+-- requires: tables/form_change
+
+begin;
+
+create or replace function cif.get_form_status(int, text)
```
matthieu-foucault, line 6:
```suggestion
create or replace function cif.get_form_status(project_revision_id int, form_data_table_name text)
```

matthieu-foucault (changes requested):

app/components/TaskList.tsx
```diff
@@ -14,46 +14,14 @@ interface Props {
...
-  )
-    return "Incomplete";
-
-  if (
-    forms.some(
-      (form) =>
-        form.changeStatus === "staged" && form.validationErrors.length > 0
-    )
-  )
-    return <strong>Attention Required</strong>;
```
matthieu-foucault, line 35: "Attention Required" needs to be bold, so you'll have to re-add some logic for that. A tiny `TaskListStatus` React component would do the trick

matthieu-foucault (changes requested):

app/components/TaskListStatus.tsx
```diff
@@ -0,0 +1,15 @@
...
+}
+
+const TaskListStatus: React.FC<Props> = ({ formStatus }) => {
+  if (formStatus === "Attention Required")
+    return (
+      <div className="status">
+        <strong>{formStatus}</strong>
+      </div>
+    );
+  return <div className="status">{formStatus}</div>;
```
matthieu-foucault, line 12: The following style needs to be moved from the `TaskList` component to here
```css
          .status {
            text-align: right;
            padding-right: 5px;
          }
```

matthieu-foucault (changes requested):

matthieu-foucault (approved):

Issue #527, opened 2022-04-26
postgres cluster yaml configuration should not override the default `image` fields


matthieu-foucault:
#### Describe the task

We should let the operator decide which image to use, as this could lead to incompatibilities when the operator is upgraded 

#### Acceptance Criteria

no `image` in `postgres.yaml`

#### Additional context

Relevant rocketchat discussion: 
- https://chat.developer.gov.bc.ca/channel/crunchydb?msg=gbt3wHkmkgCx9BHyF
- https://chat.developer.gov.bc.ca/channel/crunchydb#:~:text=Matthieu%20Foucault%20%40,3%3A47%20PM

PR #581, opened 2022-05-10
chore: remove default image from postgres.yaml

BCerki:
Issue #527 

I left in the proxy and monitoring images because they seem to be required per the docs: https://access.crunchydata.com/documentation/postgres-operator/v5/references/crd/#:~:text=When%20omitted%2C%20the%20value%20comes%20from%20an%20operator%20environment%20variable.

### DIFF
--- a/chart/cas-cif/templates/postgres.yaml
+++ b/chart/cas-cif/templates/postgres.yaml
@@ -6,7 +6,6 @@ metadata:
 spec:
   metadata:
     labels: {{ include "cas-cif.labels" . | nindent 6 }}
-  image: registry.developers.crunchydata.com/crunchydata/crunchy-postgres:centos8-14.2-0
   postgresVersion: 14
   instances:
     - name: pgha1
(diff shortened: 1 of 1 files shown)


matthieu-foucault (changes requested):
All of the `image` tags are optional. We want to remove them to let the operator decide for us, and to have our images upgrade automatically when the operator is upgraded.

While we're at it, setting resources for the `pgMonitor`, and `replicaCertCopy` containers, as shown in https://github.com/bcgov/how-to-workshops/pull/12/, would reduce the resource usage of the cluster

BCerki:
> replicaCertCopy

I added limits for the `pgMonitor` but I didn't see a `replicaCertCopy` in the file

matthieu-foucault:
You'll have to add the entry for the `replicaCertCopy` configuration, as it's currently just using the default values.
See https://access.crunchydata.com/documentation/postgres-operator/v5/references/crd/#postgresclusterspecinstancesindexsidecarsreplicacertcopy

matthieu-foucault (approved):

PR #582, opened 2022-05-10
Feat/460 filter status projects list

gurjmatharu:


app/pages/cif/projects.tsx
```diff
@@ -88,9 +100,11 @@ export function Projects({ preloadedQuery }: RelayProps<{}, projectsQuery>) {
...
       }),
       new TextFilter("Proposal Reference", "proposalReference"),
-      new TextFilter("Status", "status", {
-        orderByPrefix: "PROJECT_STATUS_BY_PROJECT_STATUS_ID__NAME",
-      }),
+      new SearchableDropdownFilter(
+        "Status",
+        "status",
+        allProjectStatuses.edges.map((e) => e.node.name)
+      ),
```
pbastia, line 107: - Could we add a configuration item to this, that turns on or off the `freeSolo` option of the Autocomplete in the `SerarchableDropdownFilter` component (for statuses, it doesn't make sense to select multiple / by string search)
- Could we have the `SearchableDropdownFilter` hide that 'X' button, since that's already something that the 'Clear' button does on the right hand side? You'll need to set the `disableClearable` option (see [docs](https://mui.com/material-ui/api/autocomplete/#props))

app/tests/unit/pages/projects.test.tsx
```diff
@@ -136,4 +136,16 @@ describe("The projects page", () => {
       screen.queryByLabelText("Filter by Project Managers")
     ).toBeInTheDocument();
   });
+
+  it("applys correct filter variables, to mutation", () => {
```
pbastia, line 140: this is not a mutation, we're just changing the query parameters
```suggestion
  it("renders the page with the status query variable from the status filter", () => {
```

pbastia (changes requested):
I think this is a good place to add a couple of UI improvements to the Filter component, looks good otherwise!

app/components/Table/Filters/TableFilter.tsx
```diff
@@ -4,6 +4,8 @@ import React from "react";
 import type { FilterComponent } from "./types";
 
 export interface FilterSettings {
+  isDisableClearable?: boolean;
+  isFreeSoloEnabled?: boolean;
```
pbastia, line 8: These shouldn't live here, since they're specific to the SearchableDropdownFilter.
What we should do is:
- extend the `FilterSettings` interface to another interface that includes those extra config options
- make the constructor of `SearchableDropdownFilter` use that new interface as the settings object type.
The `EnumFilter` is a good example of doing that


Also, can we name these "hideClearButton" and "allowFreeFormInput" or something like that? I don't expect devs to know what those particular mui-autocomplete settings do by just looking at the name

pbastia (changes requested):
We'll also need a test for each of these options that we configure

app/components/Table/Filters/SearchableDropdownFilter.tsx
```diff
@@ -10,23 +10,35 @@ const FullWidthPopper = (props) => (
...
   constructor(
     display: string,
     argName: string,
     private options: string[],
-    settings?: FilterSettings
+    settings?: SearchableDropdownSettings
   ) {
     super(display, argName, settings);
+    this.isFreeSoloEnabled = settings?.isFreeSoloEnabled ?? false;
+    this.hideClearButton = settings?.hideClearButton ?? false;
```
pbastia, line 27: this should be true by default, it'll be most of our use cases

pbastia (changes requested):
one more minor change

app/components/Table/Filters/SearchableDropdownFilter.tsx
```diff
@@ -10,23 +10,35 @@ const FullWidthPopper = (props) => (
...
 export default class SearchableDropdownFilter extends TableFilter<string> {
   constructor(
     display: string,
     argName: string,
     private options: string[],
-    settings?: FilterSettings
+    settings?: SearchableDropdownSettings
   ) {
     super(display, argName, settings);
+    this.allowFreeFormInput = settings?.allowFreeFormInput ?? true;
```
pbastia, line 26: I think this one should still default to false

app/pages/cif/projects.tsx
```diff
@@ -88,18 +100,22 @@ export function Projects({ preloadedQuery }: RelayProps<{}, projectsQuery>) {
...
       }),
       new TextFilter("Proposal Reference", "proposalReference"),
-      new TextFilter("Status", "status", {
-        orderByPrefix: "PROJECT_STATUS_BY_PROJECT_STATUS_ID__NAME",
-      }),
+      new SearchableDropdownFilter(
+        "Status",
+        "status",
+        allProjectStatuses.edges.map((e) => e.node.name),
+        { hideClearButton: true }
```
pbastia, line 107: This is the default now, we can remove it

pbastia (changes requested):

app/pages/cif/projects.tsx
gurjmatharu: removed

gurjmatharu: Removed

pbastia (approved):
awesome

Issue #473, opened 2022-04-07
Add screenshots to Cypress tests - small screen width


elliottSlee:
#### Describe the task

We can probably add some more screenshots in our Cypress tests, so that smaller width are tested and we can ensure that the page layout is responsive

#### Acceptance Criteria

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

PR #590, opened 2022-05-10
chore: add small screen sizes to happo

Sepehr-Sobhani:
Card: https://github.com/bcgov/cas-cif/issues/473

If breakpoints are not enough I can add more breakpoints for each browser.

### DIFF
--- a/app/.happo.js
+++ b/app/.happo.js
@@ -20,5 +20,13 @@ module.exports = {
       viewport: laptopViewport,
       scrollStitch: true,
     }),
+    "ios-safari": new RemoteBrowserTarget("ios-safari", {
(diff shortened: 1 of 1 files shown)


matthieu-foucault (changes requested):
The use case here is not users using the application on iOS/iPad, but rather desktop/laptop users splitting their screen to have the application on one side, and documents on the other. 

The targets should be chrome, firefox, edge and safari, with a `600x768` viewport, giving us a bit of margin if a laptop user want to have the application take a bit less than half of their screen.

pbastia (approved):
slick!

matthieu-foucault (approved):

PR #597, opened 2022-05-11
Chore/config relay update

BCerki:
This PR:
- changes the default role (crunchydb doesn't like underscores)
- bump relay to 13.2.0 and update a test that affected
- bump mui (there was a bug fix) and update the tests


BCerki:
Will update relay and mui in separate PRs

PR #600, opened 2022-05-12
Feat/duedate widget

mikevespi:
Widget for due dates that display days / weeks until due.

Packages added:
- [luxon](https://moment.github.io/luxon/#/?id=luxon)
- [react-datepicker](https://reactdatepicker.com/)
duedatewidget


app/tests/unit/lib/theme/widgets/DueDateWidget.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  it("calls onChange with the correct date", () => {
+    const expectedDate = "2050-12-15T23:59:59.999-08:00";
+    const handleOnChange = jest.fn();
+    const props: any = {
+      id: "test-id",
+      onChange: handleOnChange,
+      label: "Due date",
+      value: "2020-01-01T23:59:59.999-07:00",
```
BCerki, line 27: The widget receives props as a string in this format, so I thought it might be better to use a string in the test than a calculation

app/cypress/integration/cif/project-revision/index.spec.js
```diff
@@ -415,7 +420,10 @@ describe("the new project page", () => {
     cy.get("a")
       .contains(/Quarterly reports/i)
       .click();
-    cy.get('[label*="Due Date"]').eq(0).should("have.value", "1995-01-01");
+    // the date is formatted using locale and it can show up slightly differently in different browsers (e.g., sometimes there's a period after the month abbreviation, sometimes there's not)
+    cy.get('[aria-label*="Due Date"]')
+      .eq(0)
+      .contains(/Jan[.]? 01, 1995/);
```
BCerki, line 426: I tried to enforce formatting using [`toFormat`](https://moment.github.io/luxon/api-docs/index.html#toformat) but the headless browsers still add a period to abbreviate the month. Thoughts?

app/lib/theme/widgets/DueDateWidget.tsx
```diff
@@ -0,0 +1,92 @@
+import { WidgetProps } from "@rjsf/core";
+import React, { forwardRef, useRef } from "react";
+import DatePicker from "react-datepicker";
+import "react-datepicker/dist/react-datepicker.css";
+import CalendarTodayIcon from "@mui/icons-material/CalendarToday";
+import getTimestamptzFromDate from "lib/helpers/getTimestamptzFromDate";
+import { DateTime, Interval } from "luxon";
+
+const DueDateInput = forwardRef<HTMLDivElement, WidgetProps>(
```
pbastia, line 9: Any reason we use a `forwardRef` here, and not a regular `React.FC<WidgetProps>` or `React.FC<Partial<WidgetProps>>` ? is it required by the react-datepicker library?

mikevespi: It is required by `react-datepicker` unfortunately. The `onClick` needs to get forwarded from the datepicker to the input.

pbastia: Ah, makes sense. 👍

pbastia (approved):
very nice! I think the font of the placeholder in the widget is a bit off, but we can address that separately if we wish to do that

PR #605, opened 2022-05-13
Chore/mock router

mikevespi:
This PR addresses the issue of repeatedly mocking the `NextRouter` in our test files.

Added a helper function `createMockRouter` which returns a Next Router, with all values mocked. (Functions are mocked with a `jest.fn()`.)

Example usage:
```javascript
    expect(pageTestingHelper.router.push).toHaveBeenCalledWith(
      getProjectRevisionPageRoute("mock-proj-rev-id")
    );
```
```javascript
const mockPathname = "/cif/project-revision/[projectRevision]/form/overview";
pageTestingHelper.setRouterOptions({ pathname: mockPathname });
```

### DIFF
--- a/app/tests/unit/components/Attachment/AttachmentTableRow.test.tsx
+++ b/app/tests/unit/components/Attachment/AttachmentTableRow.test.tsx
@@ -1,15 +1,11 @@
 import { screen } from "@testing-library/react";
 import AttachmentTableRow from "components/Attachment/AttachmentTableRow";
-import { mocked } from "jest-mock";
-import { useRouter } from "next/dist/client/router";
 import { graphql } from "react-relay";
 import ComponentTestingHelper from "tests/helpers/componentTestingHelper";
 import compiledAttachmentTableRowTestQuery, {
   AttachmentTableRowTestQuery,
 } from "__generated__/AttachmentTableRowTestQuery.graphql";
 import { AttachmentTableRow_attachment } from "__generated__/AttachmentTableRow_attachment.graphql";
-
-jest.mock("next/dist/client/router");
 
 const testQuery = graphql`
   query AttachmentTableRowTestQuery @relay_test_operation {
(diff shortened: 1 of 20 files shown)


app/tests/unit/pages/project-revision/[projectRevision]/form/managers.test.tsx
```diff
@@ -327,19 +315,15 @@ describe("The Project Managers form page", () => {
       });
 
     pageTestingHelper.loadQuery();
-    pageTestingHelper.renderPage();
+    const mockPush = jest.fn();
+    pageTestingHelper.renderPageWithMockRouter({ push: mockPush });
     handleSubmit();
     expect(mockPush).toHaveBeenCalledWith(
       getProjectRevisionPageRoute("mock-proj-rev-2")
     );
```
Comment (pbastia) [inline app/tests/unit/pages/project-revision/[projectRevision]/form/managers.test.tsx:323]: I was thinking we could expose the mock router object on the helper, similarly to what we do for the relay environment. This way the user wouldn't have to setup their own mock functions and configure the helper, it would all be done automatically, and the API would look like this:

```ts
pageTestingHelper.renderPage();
expect(pageTestingHelper.router.push).toHaveBeenCalledWith(...)
```
I'm not even sure there would be value in \*not\* mocking the router

app/tests/unit/components/Attachment/AttachmentTableRow.test.tsx
```diff
@@ -59,15 +55,8 @@ const componentTestingHelper =
     getPropsFromTestQuery: (data) => ({ attachment: data.query.attachment }),
   });
 
-const routerPush = jest.fn();
-
 describe("The Attachment table row component", () => {
   beforeEach(() => {
-    jest.resetAllMocks();
```
BCerki, line 66: Do we want to remove this? The jest config clears all mocks automatically, but it doesn't like there's anything automatic for resetting mocks

BCerki (commented):
Looks good! I found the PR description helpful; maybe we could copy it into `example_usage.md`

PR #606, opened 2022-05-16
Chore/update relay

BCerki:
This PR:
- changes the default role (crunchydb doesn't like underscores)
- bump relay to 13.2.0 and update tests that affected

Note: In the happo shots, it looks like Safari is adding the current date to the date widgets if the user doesn't fill them out. Those are going to be replaced by custom widgets shortly, so is that behaviour okay temporarily?


pbastia (dismissed):
very nice, will need a rebase

PR #629, opened 2022-05-19
Update issue templates

Sriram-pc:


dleard:
combined with #628 & #631 into #654

PR #630, opened 2022-05-19
refactor: create a single page for project revision forms

matthieu-foucault:


app/hooks/useRedirectToValidFormIndex.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect } from "react";
+import { useRouter } from "next/router";
+/**
+ *  Hook that returns true if the formIndex is out of bounds of the formPages array,
+ *  and redirects to the first page of the form.
+ *  False otherwise.
+ */
+export default function useRedirectToValidFormIndex(
+  formIndex,
+  formPagesLength
```
matthieu-foucault, line 10:
```suggestion
  formIndex: number,
  formPagesLength: number
```

app/hooks/useRedirectToValidFormIndex.ts
```diff
@@ -0,0 +1,23 @@
...
+) {
+  const router = useRouter();
+  useEffect(() => {
+    if (formIndex < 0 || formIndex >= formPagesLength)
+      router.replace(
+        `${router.asPath.replace(`/form/${formIndex}`, "/form/0")}`
+      );
```
matthieu-foucault, line 17:
```suggestion
      router.replace({
        pathname: router.pathname,
        query: {
          ...router.query,
          fornIndex: 0
        }
      });
```

app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx
```diff
@@ -35,23 +35,35 @@ const pageQuery = graphql`
...
-        ...ProjectForm_projectRevision
         ...TaskList_projectRevision
+        ...ProjectForm_projectRevision
         ...ProjectFormSummary_projectRevision
+        ...ProjectContactForm_projectRevision
+        ...ProjectContactFormSummary_projectRevision
+        ...ProjectManagerFormGroup_projectRevision
+        ...ProjectManagerFormSummary_projectRevision
+        ...ProjectQuarterlyReportForm_projectRevision
+        ...ProjectAnnualReportForm_projectRevision
```
Comment (pbastia) [inline app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx:46]: I'm wondering if eventually, we can wrap this component in a factory that allows us to pass just the query we need instead of the entire list of fragments

app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx
```diff
@@ -109,30 +133,34 @@ export function ProjectOverviewForm({
...
             projectRevision={query.projectRevision}
             viewOnly={true}
           />
         </>
       ) : (
-        <ProjectForm
+        <EditComponent
           query={query}
           projectRevision={query.projectRevision}
           onSubmit={handleSubmit}
```
Comment (pbastia) [inline app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx:166]: This means that if for some reason there is no view component (like it's the case for quarterly and annual reports), then we show the edit form even in view mode. It might lead to a bug?

Comment (mikevespi) [reply app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx:166]: @pbastia  Are the report pages not going to get a view only component?

app/tests/unit/components/TaskList.test.tsx
```diff
@@ -103,8 +97,8 @@ describe("The ProjectManagerForm", () => {
     fireEvent.click(screen.getByText(/Edit project overview/i));
 
     expect(componentTestingHelper.router.push).toHaveBeenCalledWith(
-      "/cif/project-revision/[projectRevision]/form/overview?projectRevision=test-project-revision-id",
-      "/cif/project-revision/test-project-revision-id/form/overview",
+      "/cif/project-revision/[projectRevision]/form/[formIndex]?projectRevision=test-project-revision-id&formIndex=0",
+      "/cif/project-revision/test-project-revision-id/form/0",
```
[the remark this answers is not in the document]
pbastia: The "real" URLs are the ones from the file structure we setup, with the square brackets and the traditional query params - and this is what the express server handles.

The second argument in the `push` function instructs the Next `Router` object to mess with the browser's url to display the substituted `[...]` for their query params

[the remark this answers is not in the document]
pbastia: There is info about the `as` param [here](https://nextjs.org/docs/api-reference/next/router#routerpush)

pbastia (approved):
A couple comments, but this is absolutely amazing 🥇

app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx
Comment (pbastia) [reply app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx:166]: They are! Just being mindful of writing code that can handle our potential mistakes or forgetfulness :)

Comment (mikevespi) [reply app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx:166]: I replaced the `viewComponent: null` with `viewComponent: () => null` to avoid people making changes to the reports that aren't being saved anywhere. Good catch, it was rendering the form in edit mode despite being in view mode, so changes could be made but wouldn't get saved anywhere. In the future, if we have components that actually don't have a `viewComponent` we'll have to rethink this maybe.

PR #650, opened 2022-05-24
Feat/quarterly report undo changes

Sepehr-Sobhani:


app/components/Form/ProjectQuarterlyReportForm.tsx
```diff
@@ -205,10 +206,56 @@ const ProjectQuarterlyReportForm: React.FC<Props> = (props) => {
...
+        });
+      });
+      completedPromises.push(promise);
+    });
+    try {
+      await Promise.all(completedPromises);
+    } catch (e) {
+      // the failing mutation will display an error message and send the error to sentry
+    }
+  };
```
matthieu-foucault, line 252: There is a lot of code duplication going on between all of our front-end files. Can we create a database function instead that receives an array of `formChange.rowId`, undoes them if needed, and returns the corresponding form_change records?

Something like:

```sql
create function cif.undo_form_changes(form_change_ids int[])
returns setof cif.form_change
```

Sepehr-Sobhani: Sure thing, I'm working on that with the help of Dylan.

Sepehr-Sobhani:
This one was replaced with #651

PR #651, opened 2022-05-24
Make primary contact optional

matthieu-foucault:


app/components/Form/ProjectQuarterlyReportForm.tsx
```diff
@@ -71,6 +72,15 @@ const ProjectQuarterlyReportForm: React.FC<Props> = (props) => {
...
     props.projectRevision
   );
+
+  // TODO: make it a reusable hook
+  // this aims to delete the form reference from the formRefs object when hitting the undo changes button
+  useEffect(() => {
+    Object.keys(formRefs.current).forEach((key) => {
+      if (!formRefs.current[key]) delete formRefs.current[key];
+    });
+  }, [projectRevision.projectQuarterlyReportFormChanges]);
```
matthieu-foucault, line 82: @Sepehr-Sobhani this needs to be added to Contacts and Managers

schema/deploy/mutations/undo_form_changes.sql
```diff
@@ -0,0 +1,39 @@
...
+
+  foreach form_change_id in ARRAY $1 loop
+    select * from cif.form_change where id = form_change_id into fc;
+    if fc.previous_form_change_id is not null then
+      update cif.form_change set operation = 'update', new_form_data = (select new_form_data from cif.form_change where id = fc.previous_form_change_id)
+        where id = fc.id;
+    else
+    -- we need to treat project overview table differently as it needs an empty object as form data
+      if fc.form_data_table_name = 'project' then
+        update cif.form_change set new_form_data = '{}' where id = fc.id;
```
matthieu-foucault, line 26:
```suggestion
    -- we need to treat project overview table differently as it needs a null value as form data
      if fc.form_data_table_name = 'project' then
        update cif.form_change set new_form_data = null where id = fc.id;
```

Sepehr-Sobhani (approved):

PR #658, opened 2022-05-25
Chore/627 GitHub registry

gurjmatharu:


pbastia (changes requested):
we're missing the changes to artifactory url in the `values.yaml` file of the helm chart.
@matthieu-foucault do you know if it's something we need to setup, or if we can just use a predetermined proxy?

matthieu-foucault:
Looks like there is no remote repository setup for `ghcr.io`, so we'll have to stop using artifactory, which is okay since there are no cost/quotas.

See https://github.com/BCDevOps/developer-experience/blob/master/apps/artifactory/DEVHUB-README.md#how-do-i-find-out-what-repos-are-available to find out the available repos

matthieu-foucault (approved):
Looks good to me. A rebase might fix CI

pbastia (approved):

pbastia (changes requested):
wait wait

chart/cas-cif/values.yaml
```diff
@@ -6,9 +6,9 @@ replicaCount: 2
 
 image:
   app:
-    repository: artifacts.developer.gov.bc.ca/google-docker-remote/ggl-cas-storage/cas-cif-app
+    repository: ghcr.io/bcgov/cas-cif-schema/cas-cif-app
```
pbastia, line 9: This should be:

```suggestion
    repository: ghcr.io/bcgov/cas-cif-app
```

pbastia (changes requested):
something was copy pasted just a bit too quickly

pbastia (approved):

PR #663, opened 2022-05-26
chore: placeholder functions for pages with no view component

mikevespi:
Render null in view mode for pages with no view component yet. Avoids having a fillable form that submitting does nothing.

ViewComponents will be added soon.


pbastia (changes requested):
We can remove the check for the `ViewComponent` here:
https://github.com/bcgov/cas-cif/blob/183d7d4ec6ad470b25bf28de2cf708ed54f0f93b/app/pages/cif/project-revision/%5BprojectRevision%5D/form/%5BformIndex%5D/index.tsx#L154

pbastia (approved):

PR #680, opened 2022-05-31
Feat: attachments in tasklist

pbastia:
Bringing back the attachments!
image


app/tests/unit/pages/project/attachments.test.tsx
```diff
@@ -50,15 +51,30 @@ describe("The project's attachment page", () => {
     // 5 rows: 1 header, 1 filter, 1 for the pagination, and 2 for the attachments
     expect(screen.getAllByRole("row")).toHaveLength(5);
   });
+
+  // eslint-disable-next-line jest/expect-expect
```
matthieu-foucault, line 55: interesting. Wondering if we can configure eslint to recognize `expectMutationToBeCalled`

matthieu-foucault (approved):

PR #702, opened 2022-06-07
Fix: Due date indicator off by one day

pbastia:
Fixing inconsistent date handling in the report due date indicator
image


mikevespi (dismissed):
Good catch and fix :D

mikevespi (approved):

Issue #532, opened 2022-04-27
As a project coordinator, I want to see a computed reporting status for quarterly and annual reports


matthieu-foucault:
#### Description:

Show a reporting status, computed based on the state of the reporting requirements.
This is only about the computed status. Any manual edit will be described in another User Story.

#### Acceptance Criteria:

Given I am a CIF administrator
When I view the Quarterly or Annual milestones
Then I can see a badge with the status of the project

Given I am a CIF administrator
When I view the Quarterly or Annual milestones
And there are late reporting requirements
Then the status is "Late"

Given I am a CIF administrator
When I view the Quarterly or Annual milestones
And all the past due reporting requirements have been marked complete
Then the status is "On Track"

Given I am a CIF administrator
When I view the Quarterly or Annual milestones
And there are no upcoming reporting requirements 
Then the status is "On Track"

#### Development Checklist:

- [ ] A `reporting_requirement_status` table in the database with "On Track" and "Late" values
- [ ] two computed columns: returning the `reporting_requirement_status` type  (one for quarterly, one for annual)
- [x] A react component to show the status as a badge
- [x] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

pbastia:
![image.png](https://images.zenhubusercontent.com/5faafee70e671ac198e88bc7/009681df-1c90-41a7-a128-aeef16f85cee)

pbastia:
[...]

PR #719, opened 2022-06-08
Overall report status badge (quarterly and annual)

BCerki:
Card: https://github.com/bcgov/cas-cif/issues/532
Note to self: check Project Contacts Form with errors happo shots after the rebase; there were extra fields for some reason

### DIFF
--- a/app/components/Form/ProjectAnnualReportForm.tsx
+++ b/app/components/Form/ProjectAnnualReportForm.tsx
@@ -20,7 +20,7 @@ import {
   stageReportFormChanges,
   getSortedReports,
 } from "./reportingRequirementFormChangeFunctions";
-import { determineVariant, isOverdue } from "lib/helpers/reportStatusHelpers";
+import { getReportingStatus, isOverdue } from "lib/helpers/reportStatusHelpers";
 import StatusBadge from "components/StatusBadge";
 
 interface Props {
@@ -101,15 +101,20 @@ const ProjectAnnualReportForm: React.FC<Props> = (props) => {
     );
   }, [projectRevision.projectAnnualReportFormChanges]);
 
-  const overdue = isOverdue(
+  const reportDueDate =
     projectRevision.upcomingAnnualReportFormChange?.asReportingRequirement
-      .reportDueDate
-  );
+      .reportDueDate;
+  const overdue = useMemo(() => {
+    return isOverdue(reportDueDate);
+  }, [reportDueDate]);
 
-  const variant = determineVariant(
-    projectRevision.projectAnnualReportFormChanges,
-    overdue
-  );
+  const reportSubmittedDates = useMemo(() => {
+    return projectRevision.projectAnnualReportFormChanges.edges.map(
+      ({ node }) => node.newFormData.submittedDate
+    );
+  }, [projectRevision.projectAnnualReportFormChanges.edges]);
+
+  const variant = getReportingStatus(reportSubmittedDates, overdue);
 
   return (
     <div>
--- a/app/components/Form/ProjectAnnualReportForm.tsx
+++ b/app/components/Form/ProjectAnnualReportForm.tsx
@@ -20,6 +20,8 @@ import {
   stageReportFormChanges,
   getSortedReports,
 } from "./reportingRequirementFormChangeFunctions";
+import { determineVariant, isOverdue } from "lib/helpers/reportStatusHelpers";
+import StatusBadge from "components/StatusBadge";
 
 interface Props {
   onSubmit: () => void;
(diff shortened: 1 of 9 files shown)


app/tests/unit/lib/helpers/reportStatusHelpers.test.ts
```diff
@@ -0,0 +1,184 @@
+import { screen } from "@testing-library/react";
```
BCerki, line 1: I'm testing the helpers like this because I want to make sure they work together properly and show the right status based on the given data. I'm not convinced this test is the best way to do that, though--thoughts?

app/lib/helpers/reportStatusHelpers.tsx
```diff
@@ -0,0 +1,36 @@
+import { DateTime } from "luxon";
+
+export const daysUntilReportDue = (reportDueDate) => {
```
matthieu-foucault, line 3: can we type those date parameters, not sure if we're expecting a `string`, `Date` or luxon `DateTime`

app/lib/helpers/reportStatusHelpers.tsx
```diff
@@ -0,0 +1,36 @@
...
+  );
+
+  return Math.ceil(diff.days);
+};
+
+export const isOverdue = (reportDueDate) => {
+  return reportDueDate ? daysUntilReportDue(reportDueDate) < 0 : false;
+};
+
+export const determineVariant = (reportFormChanges, isReportOverdue) => {
```
matthieu-foucault, line 20: not sure about the function name. I seems to return a status.

app/tests/unit/lib/helpers/reportStatusHelpers.test.ts
matthieu-foucault: It sounds overkill to use testing-library and graphql to test some simple functions, no?

Sepehr-Sobhani:
> Card: #532 Note to self: check Project Contacts Form with errors happo shots after the rebase; there were extra fields for some reason

@BCerki Mattheiu updated contact form e2e tests [here](https://github.com/bcgov/cas-cif/pull/568/commits/8ba4de5112586ed28bfa6deed663202ae4bf2b61) and it should be fine now.

app/cypress/integration/cif/project-revision/index.spec.js
```diff
@@ -94,6 +94,7 @@ describe("the new project page", () => {
     cy.contains("Changes saved").should("be.visible");
     cy.addQuarterlyReport(3, "1993-01-01", "1992-12-31");
     cy.contains("Changes saved").should("be.visible");
+    cy.get('span[class*="MuiChip-label"]').should("have.text", "Complete");
```
matthieu-foucault, line 97: if this it the best way you can select the status badge, it's probably missing some accessible role (e.g. [status](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role))

app/components/ReportingRequirement/ReportDueIndicator.tsx
```diff
@@ -96,7 +84,7 @@ const ReportDueIndicator: React.FC<Props> = ({
                 {hasValidReportDueDate ? (
                   <>
                     {overdue ? "Overdue by " : "Due in "}
-                    <b>{Math.abs(reportDueIn.days)} day(s)</b>
+                    <b>{Math.abs(reportDueIn)} day(s)</b>
```
matthieu-foucault, line 87: This will eventually need refactoring in a separate function, so that is displays a number of months/years when appropriate

app/components/Form/ProjectAnnualReportForm.tsx
```diff
@@ -86,15 +96,24 @@ const ProjectAnnualReportForm: React.FC<Props> = (props) => {
...
 
+  const overdue = isOverdue(
+    projectRevision.upcomingAnnualReportFormChange?.asReportingRequirement
+      .reportDueDate
+  );
+
+  const variant = determineVariant(
+    projectRevision.projectAnnualReportFormChanges,
+    overdue
+  );
```
matthieu-foucault, line 107: These function (especially the second one as it iterates on the form changes) should use `useMemo`.

app/lib/helpers/reportStatusHelpers.tsx
matthieu-foucault: The arguments should be typed, especially the first one as the type could be changed by modifying a relay query, breaking this function

app/lib/helpers/reportStatusHelpers.tsx
```diff
@@ -0,0 +1,36 @@
...
+  if (areAllReportsComplete && reportFormChanges.edges.length !== 0) {
+    variant = "complete";
+  } else if (reportFormChanges.edges.length === 0) {
+    variant = "none";
+  } else if (isReportOverdue) {
+    variant = "late";
+  } else {
+    variant = "onTrack";
+  }
+  return variant;
```
matthieu-foucault, line 35: no need for a `variant` variable here, you can just return the value.

matthieu-foucault (changes requested):

app/lib/helpers/reportStatusHelpers.tsx
BCerki: > not sure about the function name. I seems to return a status.

The `StatusBadge` component calls the prop variant. (`interface Props {
  label?: string;
  variant: "complete" | "late" | "onTrack" | "inReview" | "customText" | "none";
}`)

BCerki:
@matthieu-foucault, I've applied suggestions and rebased

matthieu-foucault (approved):

Issue #587, opened 2022-05-10
As a CIF administrator, I want to see the Milestone reporting requirements on the project view page, so that I can get information about them


pbastia:
#### Description:

Adding the milestones form to the project view page.
We can base the UI look based on what was already done for the other read-only forms, and add the same visual cues about reporting requirement status than in the wireframes.

#### Acceptance Criteria:

Given
When
Then

#### Development Checklist:

- [ ] Checklist item
- [ ] Checklist item
- [ ] Checklist item
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

pbastia:
@suhafa @nanyangpro there might be a tiny bit of wireframing to do here, because this form is quite complex?

dleard:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/587?planning-poker) with ZenHub @BCerki @gurjmatharu @matthieu-foucault @mikevespi @pbastia @Sepehr-Sobhani

suhafa:
See [Figma file](https://www.figma.com/file/PzOY8RyQnVXnjZJdSv2AqQ/Wireframing_Figma?node-id=1415%3A2663) for illustration of the differences between viewing modes. Feel free to add any comments if the documentation/wireframes need further elaboration!

LindsayMacfarlane:
Thank you @gurjmatharu! Tested in dev by:
[...]

PR #727, opened 2022-06-09
feat: #587 milestone reporting read only

gurjmatharu:


app/tests/unit/components/Form/ProjectMilestoneReportFormSummary.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+    componentTestingHelper.reinit();
+  });
+
+  it("Only displays the data fields that have changed", () => {
+    componentTestingHelper.loadQuery();
+    componentTestingHelper.renderComponent();
+
+    // First Node
+    expect(screen.getByText("Updated Test comment")).toBeInTheDocument();
+    expect(screen.getByText(/Jan[.]? 2, 2020/)).toBeInTheDocument();
```
BCerki, line 142: These assertions are the same as in the test below--maybe in this test it would be better to look for the field labels you'd expect to see/not see? `ProjectManagerFormSummary.test.tsx` does it that way

BCerki (changes requested):

BCerki (dismissed):

BCerki (approved):

Issue #728, opened 2022-06-09
Update eslint realy to be strict


Sepehr-Sobhani:
update the eslint relay rules to be strict and output an error instead of a warning that would be good to ensure we follow this rule as much as possible.

Take a look at : https://github.com/relayjs/eslint-plugin-relay#how-to-use

LindsayMacfarlane:
Thanks @BCerki - no review required on my end, moving to closed.

PR #731, opened 2022-06-10
Tech debt: set relay eslint rules to strict

BCerki:
Card: https://github.com/bcgov/cas-cif/issues/728

When removing something from a graphql query broke the tests or the build, I disabled eslint for the relevant lines.

### DIFF
--- a/app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx
+++ b/app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable relay/must-colocate-fragment-spreads*/
 import { Button } from "@button-inc/bcgov-theme";
 import DefaultLayout from "components/Layout/DefaultLayout";
 import TaskList from "components/TaskList";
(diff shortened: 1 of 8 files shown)


app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable relay/must-colocate-fragment-spreads*/
```
Comment (BCerki) [inline app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx:1]: OK to this globally for this file? Otherwise I'd have to do it 15 times for all the fragments

Comment (Sepehr-Sobhani) [reply app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx:1]: I think makes sense for that file

Comment (Sepehr-Sobhani) [reply app/pages/cif/project-revision/[projectRevision]/form/[formIndex]/index.tsx:1]: @BCerki Not sure but I think we need to be more specific here regarding rules. according to this [page](https://github.com/relayjs/eslint-plugin-relay/blob/main/eslint-plugin-relay.js), some rules under strict mode don't comply with our project (like `flow types`). But maybe I'm wrong 😃 worth asking more experienced devs.

matthieu-foucault (approved):

matthieu-foucault (approved):

PR #739, opened 2022-06-14
Chore/use generic form changes for

mikevespi:
We have a generic `project_revision_form_changes_for` computed column that can be used generically for form changes of a certain type. The exceptions are managers and project form. Managers are handled differently, and `projectFormChange` is referenced for the record ID by other pages.

https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/637


dleard (approved):
Awesome!
Screenshot from 2022-06-14 15-46-00

dleard:
This looks great.
There is one issue: I see in the happo diffs, that _our_ diff functionality was affected. It appears to no longer show `REMOVED` when something was removed from the project. 
Screenshot from 2022-06-14 15-49-52

dleard (approved):
:shipit:

Issue #716, opened 2022-06-08
Dates cannot be cleared for reporting requirements [4]


pbastia:
Describe the Bug:

Dates cannot be cleared on the custom date widgets we use for the reporting requirements.
This is an issue if a mistake is made, and for example a "submitted date" is entered by accident.
Happens in Chrome and Firefox

![image.png](https://images.zenhubusercontent.com/5faafee70e671ac198e88bc7/c441b652-f7d1-46c3-85b2-4b0a9985148b)

Probability (how likely the bug is to happen, scored from 1-5): 3

Effect (how bad the bug is when it does happen, scored from 1-5): 4

Steps to reproduce the behaviour:
Create or edit reporting requirements for a project

Screenshots:

LindsayMacfarlane:
Tested in dev - there is now a little blue circle with an X that allows a user to quickly clear a reporting date. Confirmed functionality was visible for all report dates.

PR #740, opened 2022-06-14
feat: adding clear to date pickers

gurjmatharu:
#716


mikevespi:
Looks good to me! Can we just get some coverage added to the widgets tests for clearing the date?
Other than that there's some redundant css between the two that could be pulled out, but doesn't have to happen here.

PR #784, opened 2022-06-27
Feat/sqitch checks

matthieu-foucault:


dleard (approved):
Tested the .sh files & the make release / yarn release-it <version> commands, looks good!

Issue #642, opened 2022-05-24
As an Ops Team Member, I want to enter and track budgeting-related data on the Budgets, Expenses & Payments form


nanyangpro:
#### Description:

As part of the funding agreement, the province will pay up to a maximum funding amount, which is based on total project value and province's share percentage.

Draft Figma design: https://www.figma.com/file/PzOY8RyQnVXnjZJdSv2AqQ/Wireframing_Figma?node-id=1878%3A3066

- total project value
- maximum funding amount
- province's share (percentage)
- performance milestone holdback percentage
- anticipated funding amount

#### Acceptance Criteria:

Given I am a CIF administrator
When I create or edit a project
Then I can see a "Budgets, Expenses and Payments" section in the tasklist

Given I am a CIF administrator
When I create or edit a project
And I click on the "Budgets, Expenses and Payments" section in the tasklist
Then I can see a budgeting section
And its title is "Funding Agreement Overview"

Given I am a CIF administor
When I am on the "Budgets, Expenses and Payments" form of a project
Then I can enter the following budgeting fields:
- maximum funding amount
- total project value
- province's share (percentage)
- anticipated funding amount
- performance milestone holdback percentage

Given I am a CIF administor
When I am on the "Budgets, Expenses and Payments" form of a project
Then the following fields have the following defaults:
- province's share (percentage): 50%
- performance milestone holdback percentage: 10%

#### Development Checklist:
- [ ] create a `funding_parameters` table with the fields:
     - fkey to project(id)
     - total project value
     - maximum funding amount
     - province's share (percentage)
     - performance milestone holdback percentage
     - anticipated funding amount
- [ ] tasklist section and new form for the budgets, expenses and payments
- [ ] new form in the form config
- [ ] manual input for now, just with defaults - default values should be a frontend concern (at least for now)
- [ ] Meets the DOD

#### Notes:

- For initial development, this field can be manual input; however, in the future, there could be opportunity to automate by calculating total project value & province's share percentage. 

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [x] User story is included
- [x] User role and type are identified
- [x] Acceptance criteria are included
- [x] Wireframes are included (if required)
- [x] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

[...]

PR #796, opened 2022-06-30
feat: #642 budget related data

gurjmatharu:
ZH: [642](https://github.com/bcgov/cas-cif/issues/642)


schema/deploy/tables/funding_parameter.sql
```diff
@@ -0,0 +1,46 @@
...
+
+end
+$grant$;
+
+comment on table cif.funding_parameter is 'Table containing budget related data';
+comment on column cif.funding_parameter.id is 'Unique ID for the budget data';
+comment on column cif.funding_parameter.project_id is 'The related project for this data';
+comment on column cif.funding_parameter.total_project_value is 'Total project value';
+comment on column cif.funding_parameter.max_funding_amount is 'Maximum funding amount for this project';
+comment on column cif.funding_parameter.province_share_percentage is 'Province share percentage';
```
dleard, line 42: When writing column comments, we want to be as explicit as possible. It may not be someone who is familiar with the project that is reading these (for example a DBA external to our team be the one looking at our metadata).

```suggestion
comment on column cif.funding_parameter.province_share_percentage is 'The percent of the total funding for a project that the Province is providing';
```

app/components/Form/ProjectFundingAgreementForm.tsx
```diff
@@ -0,0 +1,192 @@
...
+                !isAddingFundingParameterFormChange
+              }
+            />
+          </header>
+
+          <FormBase
+            id="ProjectFundingAgreementForm"
+            validateOnMount={fundingAgreement?.changeStatus === "staged"}
+            idPrefix="ProjectFundingAgreementForm"
+            schema={createFundingAgreementSchema() as JSONSchema7}
```
dleard, line 166:
```suggestion
            schema={fundingAgreementSchema as JSONSchema7}
```
(if the createFundingAgreementSchema() function is removed)

dleard (changes requested):

app/components/Form/ProjectFundingAgreementForm.tsx
```diff
@@ -0,0 +1,192 @@
...
+  holdbackPercentage: {
+    "ui:widget": "TextWidget",
+  },
+  anticipatedFundingAmount: {
+    "ui:widget": "MoneyWidget",
+  },
+};
+
+// Setting default values for some fields
+const createFundingAgreementSchema = () => {
```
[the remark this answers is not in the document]
Sepehr-Sobhani: To keep the note here, if we put the default in the schema, we would always have the default values in the view mode even if we don't have a funding agreement associated with the project.

schema/deploy/tables/funding_parameter.sql
```diff
@@ -0,0 +1,46 @@
...
+end
+$grant$;
+
+comment on table cif.funding_parameter is 'Table containing funding agreement data';
+comment on column cif.funding_parameter.id is 'Unique ID for the funding agreement data';
+comment on column cif.funding_parameter.project_id is 'The related project for this data';
+comment on column cif.funding_parameter.total_project_value is 'Total project value';
+comment on column cif.funding_parameter.max_funding_amount is 'Maximum funding amount for this project';
+comment on column cif.funding_parameter.province_share_percentage is 'The percent of the total funding for a project that the Province is providing';
+comment on column cif.funding_parameter.holdback_percentage is 'The designated holdback percentage for each payment';
```
dleard, line 43: This could still use some extra info I think
```suggestion
comment on column cif.funding_parameter.holdback_percentage is 'The designated holdback percentage for each payment. CIF holds back this designated holdback percentage from each payment, and the cumulated holdback amount is the total possible value of the final (performance) milestone payment. The proponent is paid a portion of this possible total based on their project's emission intensity performance relative to the target emission intensity;
```

dleard (changes requested):

dleard (approved):

PR #814, opened 2022-07-06
Add graphql cypress helpers and docs

BCerki:
I made some graphql helper functions as part of a bug fix PR. Ultimately the bug was elsewhere and I didn't use these helpers, but they might come in handy later so I've put them in the separate PR.


app/cypress/utils/graphql-test-utils.js
```diff
@@ -0,0 +1,12 @@
+// Utility to match GraphQL mutation based on the operation name
+export const hasOperationName = (req, operationName) => {
+  const { body } = req;
+  return body.hasOwnProperty("id") && body.id === operationName;
+};
+
+// Alias mutation if operationName matches
+export const aliasMutation = (req, operationName) => {
```
pbastia, line 8: This one actually works for both Mutations and Queries
```suggestion
export const aliasOperation = (req, operationName) => {
```

app/cypress/utils/example_usage.md
```diff
@@ -0,0 +1,26 @@
+## Cypress testing helpers
+
+At the beginning of the test file import : `import { aliasMutation } from "../../../utils/graphql-test-utils";`
```
pbastia, line 3:
```suggestion
At the beginning of the test file import : `import { aliasOperation } from "../../../utils/graphql-test-utils";`
```

pbastia (changes requested):
Nice!

pbastia (approved):

PR #822, opened 2022-07-07
Fix milestone status badge in project table

BCerki:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/754

### DIFF
--- a/app/components/Project/ProjectMilestoneDue.tsx
+++ b/app/components/Project/ProjectMilestoneDue.tsx
@@ -25,7 +25,7 @@ const ProjectMilestoneDue: React.FC<Props> = ({ project }) => {
           orderBy: REPORT_DUE_DATE_ASC
           filter: {
             submittedDate: { isNull: false }
-            reportType: { includes: "Milestone" }
+            reportTypeByReportType: { isMilestone: { equalTo: true } }
           }
           first: 1
         ) {
--- a/app/components/Project/ProjectMilestoneDue.tsx
+++ b/app/components/Project/ProjectMilestoneDue.tsx
@@ -15,7 +15,7 @@ const ProjectMilestoneDue: React.FC<Props> = ({ project }) => {
     graphql`
       fragment ProjectMilestoneDue_project on Project {
         latestCommittedProjectRevision {
-          upcomingReportingRequirementFormChange {
+          upcomingReportingRequirementFormChange(reportType: "Milestone") {
             asReportingRequirement {
               reportDueDate
             }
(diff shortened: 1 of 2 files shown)


app/cypress/integration/cif/project-revision/index.spec.js
```diff
@@ -524,6 +524,8 @@ describe("the new project page", () => {
     });
     cy.findByRole("button", { name: /^submit/i }).click();
     cy.url().should("include", "/cif/projects");
+    // this checks that the project view list shows the milestone report status vs. the other report statuses
+    cy.findAllByRole("status").should("have.text", "Late");
```
BCerki, line 528: It would be good to have a dev chat about how to best test this bug fix (and generally, how to test queries) in our next promyze section. For now I added this assertion since it's just the one line cc @pbastia

app/components/Project/ProjectMilestoneDue.tsx
```diff
@@ -15,15 +15,18 @@ const ProjectMilestoneDue: React.FC<Props> = ({ project }) => {
...
               reportDueDate
             }
           }
         }
         latestCompletedReportingRequirements: reportingRequirementsByProjectId(
           orderBy: REPORT_DUE_DATE_ASC
-          filter: { submittedDate: { isNull: false } }
+          filter: {
+            submittedDate: { isNull: false }
+            reportType: { includes: "Milestone" }
```
matthieu-foucault, line 28: there is an `isMilestone` column that we should be able to use in the `ReportType`, if we want to be more resilient to changes

app/cypress/integration/cif/project-revision/index.spec.js
matthieu-foucault: Ideally the bugfix would be tested in our integration tests (jest) instead of cypress

matthieu-foucault (approved):
Looks good, resolving comments is optional and could be made into tech debts cards instead

app/cypress/integration/cif/project-revision/index.spec.js
BCerki: https://github.com/bcgov/cas-cif/issues/827

app/components/Project/ProjectMilestoneDue.tsx
matthieu-foucault: image

@BCerki  the `isMilestone` field is in `reportType`, so you have to follow the `reportTypeByReportType` foreign key

matthieu-foucault (approved):

Issue #744, opened 2022-06-15
Sort function in Project Managers triggers error message


LindsayMacfarlane:
#### Describe the Bug:
After selecting the sort function in 'Project Managers', the warning message gets triggered. 

Probability (how likely the bug is to happen, scored from 1-5):
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5):
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

#### Steps to reproduce the behaviour:
1. Go to project lists
2. Select the sort button in Project Managers
3. Get error message "Sorry, something went wrong".

#### Screenshots:
https://images.zenhubusercontent.com/402862646/fe8d6c42-4bcd-44f4-a241-b5cadbe30676/https___dev_cif_gov_bc_ca_cif___google_chrome_2022_06_15_15_14_55.mp4

#### Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

dleard:
@LindsayMacfarlane I was looking into this and I'm not sure we want this column sortable at all. Since it's a list of users what would be the expected outcome  / use of a sort? I think since it's searchable, if someone wanted to find projects assigned to a manager that they would use the search functionality.

Thoughts @matthieu-foucault @pbastia ?

pbastia:
I agree - it would be hard to find a sorting rule. Would we sort by First name? Last name? And what do we do when there are multiple managers?
I'm in favour of removing the sort function

LindsayMacfarlane:
I support your recommendation to remove the sort function - especially since there is a searchable list.

Random question - is there a reason and or difference as to why some of our filter lists have an arrow and others do not?
![image.png](https://images.zenhubusercontent.com/60f9db089e3a388facde96d1/3fbfacb5-f238-4092-9562-985b2c1ecf28)

dleard:
Yes. The arrow indicates a predefined set of values you can pick from (eg: the set of CIF statuses), whereas the filter box for project managers is a free-form filter, there is no pre-defined set of values here. The boxes with arrows can function as a dropdown, but also allow searching where the boxes without are search only

LindsayMacfarlane:
[...]

PR #829, opened 2022-07-07
fix: remove sorting for project managers from project list

dleard:
Issue #744
Removing the ability to sort managers. The sort was causing an error and there is no use in sorting this column as each row contains a list of users.


pbastia (dismissed):

pbastia (approved):
:+1: :+1:

PR #842, opened 2022-07-12
fix: manager and contact form_change records not staged when archived

BCerki:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/577

Manager deletion is staged:
image
Secondary contact deletion is stage:
image


app/tests/unit/components/Form/ProjectContactForm.test.tsx
```diff
@@ -323,12 +336,14 @@ describe("The ProjectContactForm", () => {
...
     screen.getByText(/submit/i).click();
-    expect(
-      componentTestingHelper.environment.mock.getMostRecentOperation().request
-        .variables.input
-    ).toMatchObject({
-      formChangePatch: { changeStatus: "staged" },
-    });
+    const allOperations =
+      componentTestingHelper.environment.mock.getAllOperations();
+    expect(allOperations.length).toEqual(5); // first operation is special; next four are form_change
```
BCerki, line 341: I know "special" isn't the right word but I can't remember what @pbastia called it

app/tests/unit/components/Form/ProjectManagerFormGroup.test.tsx
```diff
@@ -456,7 +478,7 @@ describe("The ProjectManagerForm", () => {
     expect(mutationUnderTest.request.variables).toMatchObject({
       input: {
         // The whole `formChange` object is null before a manager is selected, so there's no form change id (comes from `rowId`). TODO: update this test and mock data when/if that changes.
```
dleard, line 480: Looks like this is from an older change, but TODO comments in code are not great. They get forgotten or ignored most of the time and never get attended to. Instead of adding TODOs to the code itself, most TODOs should be created in our tracker as 'tech debt' cards (this would probably be a good thing to solidify as a dev team agreement). 

Though I think in this case, a note that this test should be updated to mock data if the underlying architecture changes to return a non-null form_change object would suffice without the 'TODO'.

app/tests/unit/components/Form/ProjectContactForm.test.tsx
dleard: The first operation returned by `allOperations` is the query itself

dleard, line 341:
```suggestion
    expect(allOperations.length).toEqual(5); // first operation is the initial query; next four are form_change mutations
```

app/tests/unit/components/Form/ProjectManagerFormGroup.test.tsx
```diff
@@ -252,12 +272,14 @@ describe("The ProjectManagerForm", () => {
...
     screen.getByText(/submit/i).click();
-    expect(
-      componentTestingHelper.environment.mock.getMostRecentOperation().request
-        .variables.input
-    ).toMatchObject({
-      formChangePatch: { changeStatus: "staged" },
-    });
+    const allOperations =
+      componentTestingHelper.environment.mock.getAllOperations();
+    expect(allOperations.length).toEqual(4); // first operation is special; next three are form_change (there are 4 managers but only 3 have form_change)
```
dleard, line 277:
```suggestion
    expect(allOperations.length).toEqual(4); // first operation is the initial query; next three are form_change  mutations (there are 4 managers but only 3 have form_change)
```

dleard (changes requested):
A couple of nitpicky things on comments, but otherwise looks great! :100:

dleard (approved):

Issue #852, opened 2022-07-13
Dev data is missing sector information on the project overview [6]


pbastia:
Describe the Bug:

Probability (how likely the bug is to happen, scored from 1-5): 5
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 2

Steps to reproduce the behaviour:
- Edit a project created by the dev script without touching the project overview
- Save the revision

There is an error on the project overview form

Screenshots:

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

Sepehr-Sobhani:
This card was addressed as part of [this](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/832) card.

PR #837, opened 2022-07-11
Refactor e2e tests

BCerki:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/832

Notes
- changing certain tests to jest tests will be a separate ticket and PR
- locally having trouble with cypress spontaneously redirecting during the edit test; TBC if it's a CI issue too. Comments in the code
- add more assertions to the diff page for funding agreement?
- edit test has the `dl` accessibility violation; can turn on the temp rules with the `happoAndAxe` helper function

PR #863, opened 2022-07-15
fix: add missing sector dev data

Sepehr-Sobhani:
ZH: https://github.com/bcgov/cas-cif/issues/852

### DIFF
--- a/schema/data/dev/004_cif_project.sql
+++ b/schema/data/dev/004_cif_project.sql
@@ -2,10 +2,12 @@ begin;
 
 do $$
   declare current_revision cif.project_revision;
+  declare random_sector_name text;
   begin
     for index in 1..50 loop
 
       current_revision := cif.create_project();
+      select sector_name from cif.sector order by random() limit 1 into random_sector_name;
 
       update cif.form_change
       set new_form_data= jsonb_build_object(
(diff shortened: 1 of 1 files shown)


schema/data/dev/004_cif_project.sql
```diff
@@ -2,10 +2,12 @@ begin;
 
 do $$
   declare current_revision cif.project_revision;
+  declare random_sector_name text;
```
BCerki, line 5: This dev data is used in the cypress tests, so I don't think we want sector to be randomly assigned. Any sector-related assertions could break, and we'll get happo diffs. Probably all the projects can have the same sector? They all have the same operator, summary, etc.

Sepehr-Sobhani:
As Brianna mentioned, this is addressed by: https://github.com/bcgov/cas-cif/pull/837/files#diff-ffc4c024a09ae611b6958e01782eed6d4b2bfa92763619b51cd48d6a7d4afa4aR18-R19:~:text=as%20bigint)-,%27totalFundingRequest%27%2C%20cast(rpad(index%3A%3Atext%2C%203%2C,%27sectorName%27%2C%20%27Agriculture%27,-Commenting%20on%20lines

Issue #846, opened 2022-07-12
Money amounts are missing the dollar sign in front when reviewing before project submission [6]


suhafa:
When user reviews amounts entered in the application (such as Total Project Value, or Max Funding Amount, Anticipated Funding Amount, Total Eligible Expenses etc.) in editing mode, the amount appears without the dollar sign. 

This looks like it is a bug with the `CustomDiffFields` component. We have an `isMoney` boolean populated from the uiSchema & are conditionally rendering the $. However, it is not showing up. Looking [here](https://github.com/bcgov/cas-cif/blob/develop/app/lib/theme/CustomDiffFields.tsx#L115) and working backwards to find out where the boolean is not being set properly is likely the best place to start.

This could also be an issue with the ReadOnlyWidget since the $ is not being shown when viewing in read-only. (See Suha's comment below)

Probability: 5 (happens for all dollar amounts, which all projects have)
Effect: 1 (The number is not clearly a dollar amount)

Screenshot of the issue: ![image.png](https://images.zenhubusercontent.com/6201b10c9827341d8d5cbe9c/67155a86-9d19-48c9-a7db-643956d9250f)

#### Acceptance Criteria

Given I am a CIF administrator
When I am on the project overview form
And I am looking at information related to budgets
Then they are displayed with the dollar symbol 

#### Notes:
- As noted by Ops team during the testing of the app before soft launch
- Derived from notes in #555

suhafa:
@dleard sounds good Dylan!

suhafa:
<img width="1103" alt="image" src="https://user-images.githubusercontent.com/99216375/179818788-539e9c66-8d1b-4ec5-8906-a26fe9fdbb28.png">
These values appear to not have $ signs or % signs. Would this ticket's scope cover this too @dleard

dleard:
@suhafa I've fixed the $ issue, but we don't have anything for displaying a '%' sign at the end of a percent value. Should this feature be added to our backlog as an enhancement story?

mikevespi:
@dleard this looks great, it works on funding agreement but the milestone payments still aren't showing it on dev. 

They are money widgets in the `ui:schema` so I would guess it's something in the diffs?

dleard:
Closing this as the milestone issue is handled in #886

PR #872, opened 2022-07-19
Fix: display dollar sign

dleard:
issue #846
includes an update to webpack-terser-plugin, fixes CVE-2022-25858


pbastia (approved):
👌 :shipit:

PR #873, opened 2022-07-19
Fix: handle empty project overview

mikevespi:
Card: [848](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/848) 

Fixing the bug in the card lead to the `undo changes` button not working. It was calling the `undoFormChange` mutation but would trigger an update with the previous form data, leaving you in the same state as before clicking undo. Fixed by having the function set `new_form_data` to `{}` rather than `null::jsonb`. 
This lead to the tasklist status for project overview showing `in progress` rather than `not started` on an empty form. Fixed by adding a case to `get_form_status` specifically for this scenario.

### DIFF
--- a/app/components/Form/ProjectForm.tsx
+++ b/app/components/Form/ProjectForm.tsx
@@ -163,6 +163,7 @@ const ProjectForm: React.FC<Props> = (props) => {
       ...revision.projectFormChange.newFormData,
       ...changeData,
     };
+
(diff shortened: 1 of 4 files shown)


mikevespi:
New branch made to properly handle sqitch reworks

PR #874, opened 2022-07-19
chore: handle empty project overview form statuses

mikevespi:
Card: [848](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/848)

Fixing the bug in the card lead to the undo changes button not working. It was calling the undoFormChange mutation but would trigger an update with the previous form data, leaving you in the same state as before clicking undo. Fixed by having the function set new_form_data to {} rather than null::jsonb.
This lead to the tasklist status for project overview showing in progress rather than not started on an empty form. Fixed by adding a case to get_form_status specifically for this scenario.


schema/deploy/util_functions/get_form_status.sql
```diff
@@ -10,6 +10,14 @@ $function$
...
   select
     case
+        -- return not started for empty project form
+      when $2 = 'project'
+        and (fc.change_status = 'pending')
+        and (select cif.form_change_is_pristine((select row(form_change.*)::cif.form_change from cif.form_change where id=fc.id)) is distinct from true)
+        and (
+          ((select new_form_data from cif.form_change where id=fc.id) is null)
+          or ((select new_form_data from cif.form_change where id=fc.id) = jsonb_build_object()))
+        then 'Not Started'
```
pbastia, line 20: We're going to need tests for this function, but that can be in a tech debt card

schema/test/unit/mutations/undo_form_changes_test.sql
```diff
@@ -124,9 +124,9 @@ select is(
     select new_form_data from cif.form_change
     where id = 3
   ),
-  null::jsonb
+  jsonb_build_object()
```
pbastia, line 127: This would make sure that we don't depend on the `jsonb_build_object` implementation - and improve readability
```suggestion
  '{}'::jsonb
```

pbastia (changes requested):
Nice!

schema/deploy/util_functions/get_form_status.sql
mikevespi: I was going to leave it for the card about refactoring the statuses, happy to start one out this morning though if we would rather get one in now

pbastia (approved):

Issue #850, opened 2022-07-13
Provincial and holdback percentages aren't discarded [6]


BCerki:
Describe the Bug:

Probability (how likely the bug is to happen, scored from 1-5): 2
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 4
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:
- in dev
- edit an existing project's payment form
- discard the revision
- view the project's payment form. Provincial share and holdback percentage weren't discarded

Screenshots:
https://www.loom.com/share/b3706f2142e54eb6ac0438f95cf07d49

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

pbastia:
Works well, thank you!

Issue #849, opened 2022-07-13
Province share and holdback percentages don't handle blank/0 well [5]


BCerki:
Describe the Bug:

Probability (how likely the bug is to happen, scored from 1-5): 2
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 3
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:
- in localhost
- create a new project and navigate to the budgets form
- backspace the default percentage. If you don't type the new number fast enough, it'll change back to the default and then whatever you typed is added, so you'll get something like 5033
- enter 0 for provincial and holdback percentages
- submit the form and navigate away
- return to the form
- note that the 0 percentage is now blank

Screenshots:
https://www.loom.com/share/4bea164cacd341e690c7d0cfe0fd4896

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

BCerki:
Sepehr has probably solved this as part of the work on https://github.com/bcgov/cas-cif/issues/850

Sepehr-Sobhani:
Addressed in https://github.com/bcgov/cas-cif/pull/885

pbastia:
> - enter 0 for provincial and holdback percentages
> - submit the form and navigate away
> - return to the form
> - note that the 0 percentage is now blank

This second part of the bug is still happening. No need to revisit this card, since #895 will refactor next week.

PR #885, opened 2022-07-21
Fix province holdback percentage issues

Sepehr-Sobhani:
ZH: https://github.com/bcgov/cas-cif/issues/850 & https://github.com/bcgov/cas-cif/issues/849


dleard (approved):
Nice work :shipit:

Issue #449, opened 2022-04-04
Page content/spacing bounces around when viewing Task List at different screen sizes [5]


BCerki:
Describe the Bug:

Probability (1-5): 4

Effect (1-5): 1

Steps to reproduce the behaviour: Navigate to the project overview page and resize the window to a width somewhere between ~1014px and ~550px. Switch between tasks in the task list and you'll see the spacing change and the content move (look at where the task list lines up with the top nav bar).

Screenshots:
![Screenshot from 2022-04-04 10-49-25](https://user-images.githubusercontent.com/77306764/161602140-c05d0df5-c627-4d6e-ab28-deba2d7b2f10.png)
![Screenshot from 2022-04-04 10-49-38](https://user-images.githubusercontent.com/77306764/161602156-f549d28e-a98e-44ca-aac3-f317ea642d9b.png)

Additional information:

- OS: Pop OS
- Browser: Chrome and Firefox
- Version: [e.g. 22]
- Additional context

matthieu-foucault:
Possible solution: when the layout switches to one column, the `main` content should use the entire width of the page

Sepehr-Sobhani:
@nanyangpro @suhafa As Matthiew mentioned, a possible solution could be to have forms in full width like the below loom video, how do you feel about that?!
https://www.loom.com/share/9f6081bc75684201a1ed610e101e9f8e

suhafa:
There is sufficient margin around the form which is great, looks good to me. Thanks! @Sepehr-Sobhani

pbastia:
It works!! Thank you @Sepehr-Sobhani

PR #891, opened 2022-07-22
fix: fix page content bouncing

Sepehr-Sobhani:
ZH: https://github.com/bcgov/cas-cif/issues/449


Sepehr-Sobhani:
Got merged accidentally; hit the `pull` command instead of `push` when rebasing.

pbastia (approved):
Nice catch!

PR #894, opened 2022-07-23
chore: remove cypress wait commands

Sepehr-Sobhani:
ZH: https://github.com/bcgov/cas-cif/issues/907

### DIFF
--- a/app/cypress/integration/cif/project-revision/edit-project-revision.spec.js
+++ b/app/cypress/integration/cif/project-revision/edit-project-revision.spec.js
@@ -118,6 +118,7 @@ describe("when editing a project, the project page", () => {
     cy.findByText(/No reports due/).should("be.visible");
     cy.contains("Changes saved.");
 
+    // this assertion is not necessary, but it's needed to slow down the test and run the accessibility checks correctly
     cy.findByText(/Complete/i).should("be.visible");
     cy.happoAndAxe("Project quarterly reports Form", "editing", "main");
     cy.findByRole("button", { name: /^submit/i }).click();
--- a/app/cypress/integration/cif/project-revision/edit-project-revision.spec.js
+++ b/app/cypress/integration/cif/project-revision/edit-project-revision.spec.js
@@ -118,7 +118,7 @@ describe("when editing a project, the project page", () => {
     cy.findByText(/No reports due/).should("be.visible");
     cy.contains("Changes saved.");
 
-    cy.wait(1000);
+    cy.findByText(/Complete/i).should("be.visible");
     cy.happoAndAxe("Project quarterly reports Form", "editing", "main");
     cy.findByRole("button", { name: /^submit/i }).click();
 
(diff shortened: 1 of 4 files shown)


app/cypress/integration/cif/project-revision/edit-project-revision.spec.js
```diff
@@ -118,7 +118,7 @@ describe("when editing a project, the project page", () => {
     cy.findByText(/No reports due/).should("be.visible");
     cy.contains("Changes saved.");
 
-    cy.wait(1000);
+    cy.findByText(/Complete/i).should("be.visible");
```
BCerki, line 121: Does the test still work if we add a `should('be.visible')` assertion to line 119 instead? (or replace with `cy.findByText(/Changes saved/i).should("be.visible");`) 

If we ever change the test so the status is something other than complete, we'll have to update this line too.

app/cypress/support/commands.js
```diff
@@ -147,7 +147,7 @@ Cypress.Commands.add(
     cy.contains(opsTeamPrimary).click();
 
     // FIXME: adding project managers does not trigger the saving indicator
```
BCerki, line 149: Delete comment? If the line 150 assertion works, then I think the saving indicator has to be working?

app/cypress/support/commands.js
```diff
@@ -173,23 +173,18 @@ Cypress.Commands.add(
...
-  // TODO: figure out why we need to wait when setting the primary contact
-  cy.wait(1000);
-  cy.get("label")
-    .contains("Secondary Contacts")
-    .parent()
-    .find("input")
-    .last()
+  cy.findByLabelText(/^Primary contact/i).click();
+  cy.findAllByRole("option").contains(primaryContact).click();
+  cy.findByText("bob.l003@example.com").should("be.visible");
```
BCerki, line 178: Email could be an argument instead so the function is more general

app/cypress/support/commands.js
```diff
@@ -173,23 +173,18 @@ Cypress.Commands.add(
...
+  cy.findByRole("button", { name: /add a secondary contact/i }).click();
+  cy.findAllByRole("combobox").should("have.length", 2);
+  cy.findByText(/^Secondary contact/i)
+    .next()
+    .findAllByRole("combobox")
     .click();
   cy.contains(secondaryContact).click();
-  // TODO: figure out why we need to wait when setting the primary contact
-  cy.wait(1000);
+  cy.findByText("bob.l003@example.com").should("be.visible");
```
BCerki, line 187: same comment as above

app/cypress/integration/cif/project-revision/edit-project-revision.spec.js
Sepehr-Sobhani: Further to our conversation, this line gives enough time for Axe to get run correctly and not raise accessibility issues on the due date status badge.

BCerki (approved):

PR #901, opened 2022-07-25
Fix: enter amount widget behaves poorly

dleard:
- removed the conditionalAmountWidget & replaced with moneyWidget
- conditionally alter the milestoneSchema based on reportType


app/components/Form/Functions/projectMilestoneFormFunctions.tsx
```diff
@@ -86,6 +86,24 @@ export const createMilestoneSchema = () => {
...
+      type: "number",
+      title: "Total Eligible Expenses",
+      default: undefined,
+    };
+    schema.properties.maximumAmount = {
+      type: "number",
+      title: "Maximum Amount",
+      default: undefined,
+    };
+  }
```
pbastia, line 105: I think it would be best to keep our schemas as static as possible (with the exception of UI-specific entries, like the `anyOf` item), so that we can use them for server-side validation as well.
In this case, we might want to have 2 separate schemas, one for the reporting milestone and one for the other, instead of computing it on-the-fly every time we render

pbastia (changes requested):
💯 improved usability! I think this is breaking the server-side validation, or making it weaker, we can probably simplify it by having just separate schemas

app/components/Form/Functions/projectMilestoneFormFunctions.tsx
mikevespi: As per todays discussion with @pbastia :
Have two milestone schemas, one with expenses  and one without. This means we need to change the value of `json_schema_name` on `form_change` records based on the `reporting_requirement_has_expenses` computed column. It shouldn't however require any migrations, and prevents backend validation  needing to know about a dynamic schema.

dleard: We will probably have to update `json_schema_name` with a mutation chain, otherwise the change won't be transactional. Changing the report_type is on the form_change record for the `reporting_requirement` whereas the expense info is on the form_change record for `milestone_report` (this is the form_change with the 2 different schemas and where we need to set json_schema_name). So when updating report_type on the reporting_requiremnt form_change record we'll have to update json_schema_name on the corresponding milestone_report form change.

pbastia (approved):

Issue #857, opened 2022-07-15
As an Ops team member, I want to have an updated version of the dashboard, so that it is better suited to my needs


suhafa:
#### Description:

In view of the soft launch, we've reiterated the dashboard. All updates to be implemented are covered in [**the wireframe**](https://www.figma.com/file/PzOY8RyQnVXnjZJdSv2AqQ/Wireframing_Figma?node-id=2693%3A2698) based on work done in #714

#### Acceptance Criteria:

Given I am a CIF administrator
When I am using the CIF app
And I am looking at the top navigation bar
Then I see 'Home' label that when selected will take me back to the main page

Given I am a CIF administrator
When I am on the Home page
And I am looking at the 'Projects' category
Then I see the following helpful text under the Projects heading: Create, view and manage projects

Given I am a CIF administrator
When I am on the Home page
And I am looking at the 'Reporting Operations' category
Then I see the following helpful text under the Projects heading: Create, manage and search

Given I am a CIF administrator
When I am on the Home page
And I am looking at the 'Administration' category
Then I see a link called "Data Insights (Metabase)"

Given I am a CIF administrator
When I have selected the "Data Insights (Metabase)" link
Then a new tab is opened in my browser that takes me to the correct page in Metabase (https://cas-metabase.nrs.gov.bc.ca/)

Given I am a CIF administrator
When I am on the Home page
And I am looking at the 'Administration' category
Then I see a link called "Report a Problem"

Given I am a CIF administrator
When I have selected the "Report a Problem" link
Then my email pops up pre-populated with the correct email address

Given I am a CIF administrator
When I am using the CIF app
And I am looking at the bottom navigation bar
Then I see 'Login Page' label that when selected will take me back to the CIF login page

Given I am a CIF administrator
When I am using the CIF app
And I am looking at the bottom navigation bar
Then I see the following links (with their URL):
- Login Page (https://cif.gov.bc.ca - or dev or test)
- Disclaimer (https://www2.gov.bc.ca/gov/content/home/disclaimer)
- Privacy (https://www2.gov.bc.ca/gov/content/home/privacy)
- Accessibility (https://www2.gov.bc.ca/gov/content/home/accessible-government)
- Copyright (https://www2.gov.bc.ca/gov/content/home/copyright)
- Contact Us (mailto: ggircs email, subject: CIF App: Report a problem!)

#### Development Checklist:

- [x] Update 'Dashboard' to Home in navigation bar
[...]

PR #902, opened 2022-07-25
Chore/dashboard enhancement

Sepehr-Sobhani:
ZH: https://github.com/bcgov/cas-cif/issues/857


app/components/Dashboard.tsx
```diff
@@ -108,8 +115,22 @@ const Dashboard: React.FC<Props> = ({ query: queryKey }) => {
...
         {isAdmin && (
           <section>
-            <h3 id="admin-heading">Administration</h3>
-            <nav aria-labelledby="admin-heading"></nav>
+            <div id="admin-heading">
+              <h3>Administration</h3>
+            </div>
+            <nav aria-labelledby="admin-heading">
+              <Link passHref href={getMetabaseRoute()}>
+                <a target="_blank">Data Insights (Metabase)</a>
```
BCerki, line 123: Why `<a>` vs`<BCGovLink>` like elsewhere? (question, not suggestion, I don't know much about `<BCGovLink>`)

app/tests/unit/components/Dashboard.test.tsx
```diff
@@ -115,4 +118,58 @@ describe("The Dashboard", () => {
       "/cif/project-revision/mock-id-1/form/0?anchor="
     );
   });
+  it("Shows dashboard links with admin privileges", () => {
```
BCerki, line 121: Should there also be a test to confirm that non-admins can't see the admin info?

app/components/Dashboard.tsx
Sepehr-Sobhani: Good point, thanks for noticing that. I don't know much about that too but it's better to be consistent and use that rather than `<a>` tag.
The only difference I noticed is that `BCGovLink` doesn't get purple if you visit that link.

BCerki (dismissed):

app/components/Dashboard.tsx
```diff
@@ -108,8 +115,22 @@ const Dashboard: React.FC<Props> = ({ query: queryKey }) => {
...
+            </div>
+            <nav aria-labelledby="admin-heading">
+              <Link passHref href={getMetabaseRoute()}>
+                <BCGovLink target="_blank">Data Insights (Metabase)</BCGovLink>
+              </Link>
+              <Link
+                passHref
+                href={decodeURI(
+                  getSupportEmailMailTo("CIF App: Report a problem!").pathname
+                )}
```
pbastia, line 129: Do you know why the original path is URL-encoded at that point?

app/tests/unit/components/Dashboard.test.tsx
```diff
@@ -40,6 +41,8 @@ const componentTestingHelper = new ComponentTestingHelper<DashboardTestQuery>({
   defaultComponentProps: {},
 });
 
+const supportEmail = getConfig()?.publicRuntimeConfig?.SUPPORT_EMAIL;
```
pbastia, line 44: we should mock this value instead of using the real one here
```suggestion
jest.mock("next/config");
mocked(getConfig).mockImplementation(() => ({
  publicRuntimeConfig: {
    SUPPORT_EMAIL: "test@email.com"
  }
}));
```

and add the test email in the jest expectation

app/tests/unit/components/Layout/Footer.test.tsx
```diff
@@ -0,0 +1,14 @@
+import Footer from "components/Layout/Footer";
+import { render, screen } from "@testing-library/react";
+import footerLinks from "data/dashboardLinks/footerLinks";
+
+describe("The Footer Component", () => {
+  it("Renders footer links", () => {
+    render(<Footer />);
+    footerLinks.forEach(({ name, href }) => {
```
pbastia, line 8: We want the test to break if the content of the `footerLinks` changes, so I'd recommend hard-coding the expectations here

app/data/dashboardLinks/footerLinks.ts
```diff
@@ -0,0 +1,32 @@
...
+    name: "Accessibility",
+    href: "https://www2.gov.bc.ca/gov/content/home/accessible-government",
+  },
+  {
+    name: "Copyright",
+    href: "https://www2.gov.bc.ca/gov/content/home/copyright",
+  },
+  {
+    name: "Contact Us",
+    href: `mailto:${supportEmail}?subject=CIF App: Report a problem!`,
```
pbastia, line 28: you can use your `getSupportEmailMailTo` utility function here

pbastia (changes requested):
Just a few things to make it a bit cleaner, great work!

app/components/Dashboard.tsx
Sepehr-Sobhani: since we are passing `mailto` subject like `?subject=...` the question mark turns into `%3F` in the JSX and that causes `mailto` to not properly populated with the subject.

pbastia: I can remove it with no effect on my end, and it's not required in the `Footer` component, I don't think `decodeURI(...)` is doing anything here

Sepehr-Sobhani: Hmmm strange, I've tried that on my end and the subject was getting added to the email rather than in the subject field. I will try again and post the screenshot if I face it again.

Sepehr-Sobhani: You are right, it doesn't need it. I don't know why it was acting up at that time 😄

pbastia (approved):

pbastia (changes requested):
The footer doesn't wrap properly when the page is narrower :(
image

Sepehr-Sobhani:
> The footer doesn't wrap properly when the page is narrower :( image

changing the font size or padding on mobile breakpoint would work?!

pbastia:
Not sure. This seems to be a bug with the original library:
image

pbastia:
We can solve it with a bit of CSS for now, we'll need to submit an issue on the service development toolkit
```tsx
<>
  <div>
    <BCGovFooter>
      <ul>
        {footerLinks.map(({ name, href }) => (
          <li key={name}>
            <Link passHref href={href}>
              <BCGovLink target="_blank" style={{ textDecoration: "none" }}>
                {name}
              </BCGovLink>
            </Link>
          </li>
        ))}
      </ul>
    </BCGovFooter>
  </div>
  <style jsx>
    {`
      div :global(.pg-footer-footer) {
        height: auto;
      }
    `}
  </style>
</>
```

Sepehr-Sobhani:
> We can solve it with a bit of CSS for now, we'll need to submit an issue on the service development toolkit
> 
> ```tsx
> <>
>   <div>
>     <BCGovFooter>
>       <ul>
>         {footerLinks.map(({ name, href }) => (
>           <li key={name}>
>             <Link passHref href={href}>
>               <BCGovLink target="_blank" style={{ textDecoration: "none" }}>
>                 {name}
>               </BCGovLink>
>             </Link>
>           </li>
>         ))}
>       </ul>
>     </BCGovFooter>
>   </div>
>   <style jsx>
>     {`
>       div :global(.pg-footer-footer) {
>         height: auto;
>       }
>     `}
>   </style>
> </>
> ```

haha, I did the exact same thing 😄

pbastia:
@Sepehr-Sobhani related issue: https://github.com/button-inc/service-development-toolkit/issues/446

pbastia (approved):

Issue #895, opened 2022-07-25
As a CIF user, I want to see a percent sign for percent values, so that it's clear that value is a percentage


dleard:
#### Description:

There is currently no indication that our percent values are a percent. A new `PercentWidget` should be created similar to the `MoneyWidget` to display the `%` sign.

#### Acceptance Criteria:

Given I am a CIF user
When I am viewing a value that is a percentage
Then I can see a % sign after the value

#### Development Checklist:

- [ ] A `PercentWidget` is created
- [ ] The widget is applied to our percentage values
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [x] User story is included
- [x] User role and type are identified
- [x] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

dleard:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/895?planning-poker) with ZenHub @BCerki @gurjmatharu @pbastia @Sepehr-Sobhani

pbastia:
How are the percentage value stored in the database: a number between 0 and 100? 0 and 1? Do we constrain the number of decimal places?

LindsayMacfarlane:
Thank you @gurjmatharu! Tested in Dev by reviewing:
1) Editing an existing project
2) Creating a new project
3) Project overview page
Percent signs displayed related to budgets/payments, no decimal values, and display/navigation of fields with 0% (as per checklist).

Thank you for your work on this one :)

PR #913, opened 2022-07-29
Feat/calculated intensity performance

BCerki:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/766

A few notes:
- Once we have a percent widget (https://github.com/bcgov/cas-cif/issues/895#issuecomment-1203040562) we could use that instead of `isPercent` is the widgets
- I didn't write any cypress assertions because the happo shot should cover the presence of the new field, and the workings are covered by jest.
- I turned off the `<dl>` check on the TEIMP form because we've got a [tech debt issue about it](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/563); okay?

### DIFF
--- a/app/lib/theme/widgets/FieldLabel.tsx
+++ b/app/lib/theme/widgets/FieldLabel.tsx
@@ -15,7 +15,10 @@ const FieldLabel: React.FC<Props> = ({
     return null;
   }
 
-  const displayedLabel = label + (required ? "" : " (optional)") + " ";
+  const displayedLabel =
+    label === "GHG Emission Intensity Performance"
+      ? "GHG Emission Intensity Performance"
+      : label + (required ? "" : " (optional)") + " ";
 
   if (tagName === "label")
     return <label htmlFor={htmlFor}>{displayedLabel}</label>;
--- a/app/lib/theme/widgets/FieldLabel.tsx
+++ b/app/lib/theme/widgets/FieldLabel.tsx
@@ -3,6 +3,7 @@ interface Props {
   required: boolean;
   htmlFor: string;
   tagName?: "label" | "dt";
+  uiSchema?: any;
 }
 
 const FieldLabel: React.FC<Props> = ({
@@ -10,15 +11,15 @@ const FieldLabel: React.FC<Props> = ({
   required,
   htmlFor,
   tagName = "label",
+  uiSchema,
 }) => {
   if (!label) {
     return null;
   }
 
-  const displayedLabel =
-    label === "GHG Emission Intensity Performance"
-      ? "GHG Emission Intensity Performance"
-      : label + (required ? "" : " (optional)") + " ";
+  const displayedLabel = uiSchema?.hideOptional
+    ? label
+    : label + (required ? "" : " (optional)") + " ";
 
   if (tagName === "label")
     return <label htmlFor={htmlFor}>{displayedLabel}</label>;
--- a/app/lib/theme/widgets/ReadOnlyAdjustableCalculatedValueWidget.tsx
+++ b/app/lib/theme/widgets/ReadOnlyAdjustableCalculatedValueWidget.tsx
@@ -6,8 +6,9 @@ const ReadOnlyAdjustableCalculatedValueWidget: React.FC<WidgetProps> = (
 ) => {
   const { id, value, uiSchema, label, formContext } = props;
 
-  // If we are using this widget to show numbers as money, we can set `isMoney` to true in the uiSchema.
+  // If we are using this widget to show numbers as money or percent, we can set `isMoney` or `isPercent` to true in the uiSchema.
   const isMoney = uiSchema?.isMoney;
+  const isPercent = uiSchema?.isPercent;
 
   const calculatedValue =
     formContext[uiSchema.calculatedValueFormContextProperty];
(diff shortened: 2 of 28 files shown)


schema/test/unit/computed_columns/emission_intensity_report_calculated_ei_performance_test.sql
```diff
@@ -0,0 +1,65 @@
...
+overriding system value
+values (1, 'first operator legal name', 'first operator trade name', 'AB1234567', 'ABCD');
+
+insert into cif.project(id, operator_id, funding_stream_rfp_id, project_status_id, proposal_reference, summary, project_name)
+overriding system value
+values (1, 1, 1, 1, '000', 'summary', 'project 1');
+
+insert into cif.reporting_requirement(id,report_due_date, comments, project_id, report_type, reporting_requirement_index)
+overriding system value
+values (1,'2020-01-01', 'comment_1', 1, 'General Milestone',1);
```
BCerki, line 17: I had to add all these or else the test threw errors for missing foreign keys

app/lib/theme/widgets/ReadOnlyAdjustableCalculatedValueWidget.tsx
```diff
@@ -21,10 +22,11 @@ const ReadOnlyAdjustableCalculatedValueWidget: React.FC<WidgetProps> = (
           {
             <NumberFormat
               thousandSeparator
-              fixedDecimalScale={isMoney}
+              fixedDecimalScale={isMoney || isPercent}
               id={id}
               prefix={isMoney ? "$" : ""}
-              decimalScale={isMoney ? 2 : 10}
+              suffix={isPercent ? "%" : ""}
```
pbastia, line 28: we're missing that suffix + decimalScale change in the `<NumberFormat .../>` just below

app/lib/theme/widgets/ReadOnlyCalculatedValueWidget.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <NumberFormat
+          fixedDecimalScale={isMoney || isPercent}
+          prefix={isMoney ? "$" : ""}
+          suffix={isPercent ? "%" : ""}
+          id={id}
+          decimalScale={isMoney || isPercent ? 2 : 10}
+          value={0}
+          displayType="text"
+          aria-label={label}
+        />
```
pbastia, line 37: If the calculated value is not set, this will look like we have a calculated value of "0" instead of something like "not available" (or just nothing at all, or an error message?)

app/lib/theme/widgets/FieldLabel.tsx
```diff
@@ -15,7 +15,10 @@ const FieldLabel: React.FC<Props> = ({
     return null;
   }
 
-  const displayedLabel = label + (required ? "" : " (optional)") + " ";
+  const displayedLabel =
+    label === "GHG Emission Intensity Performance"
+      ? "GHG Emission Intensity Performance"
+      : label + (required ? "" : " (optional)") + " ";
```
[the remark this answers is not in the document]
pbastia: If the goal is to not display the "(optional)" string here we could make it an option in the "ui:options" of the UI schema instead

pbastia (changes requested):
minor edits, looks great!

app/lib/theme/widgets/FieldLabel.tsx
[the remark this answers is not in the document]
BCerki: I couldn't get it to work via `ui:options` (I think because `FieldLabel` overrides that), but I got it in with another ui schema property

schema/deploy/computed_columns/form_change_as_emission_intensity_report.sql
```diff
@@ -0,0 +1,38 @@
+-- Deploy cif:computed_columns/form_change_as_emission_intensity_report to pg
+
+begin;
+
+create or replace function cif.form_change_as_emission_intensity_report(cif.form_change)
+returns cif.emission_intensity_report
+as $$
+    select
+      /**
+        Given form_data_record_id can be null for some form_change records, it is not a reliable id value for the returned project_contact record.
```
pbastia, line 10:
```suggestion
        Given form_data_record_id can be null for some form_change records, it is not a reliable id value for the returned emission_intensity_report record.
```

schema/deploy/computed_columns/form_change_as_emission_intensity_report.sql
```diff
@@ -0,0 +1,38 @@
...
+        The returned id must not be null, so we use the form_change id being passed in as a parameter (multiplied by -1 to ensure we are not touching any existing records).
+        This means the id value is not going to be the correct id for the returned project_contact record, which should be ok since we're only interested
```
pbastia, line 12:
```suggestion
        This means the id value is not going to be the correct id for the returned emission_intensity_report record, which should be ok since we're only interested
```

pbastia (changes requested):
one missing change in a comment. Ready to merge after that!

pbastia (approved):
🥇

PR #916, opened 2022-08-03
fix: adding project contacts to dev data

gurjmatharu:


app/cypress/integration/cif/project-revision/edit-project-revision.spec.js
```diff
@@ -120,6 +120,7 @@ describe("when editing a project, the project page", () => {
 
     // this assertion is not necessary, but it's needed to slow down the test and run the accessibility checks correctly
     cy.findByText(/Complete/i).should("be.visible");
+    cy.wait(1000); // added a wait to prevent accessibility errors
```
pbastia, line 123: Awaiting is really not recommended in our cypress tests...is there really no other way here?

pbastia (changes requested):
Just a code comment to delete. Looks great!

app/cypress/integration/cif/project-revision/edit-project-revision.spec.js
gurjmatharu: not sure, if we don't wait it throws axe errors.  I think it's due to the text flashing when it renders.

Sepehr-Sobhani: Sorry to interrupt your conversation here. just wanted to share my experience here when I was removing all `cy.wait()` commands based on ticket https://github.com/bcgov/cas-cif/issues/907. I faced this issue at that time too but I added an assertion on line 122 to slow down the test a little so we can have that badge rendered completely and then run the accessibility test. It's weird that we have this error again 🤦‍♂️ maybe if we add another assertion here, it can remove the need for the wait command.

pbastia: We can look into the text flashing as well, if this is the root cause of the issue

gurjmatharu: @pbastia @Sepehr-Sobhani created a bug https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/933

pbastia (approved):
👍

PR #921, opened 2022-08-04
chore: add more helpful error message for expectMutationToBeCalled

mikevespi:
Card: [851](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/851) 

Added better error message for `expectMutationToBeCalledWith`. It handles scenarios where the same mutation was called more than once, and none of the calls match the input variables. 

I know the `\n`s aren't pretty, but they make the error in the console much nicer.


BCerki (approved):
Looks nice, will make debugging much easier

Issue #925, opened 2022-08-04
`Login Page` link in the bottom navigation bar must address correct environment


Sepehr-Sobhani:
Describe the Bug:

**currently, the `Login Page` link in the bottom navigation bar only takes the user to prod environment. we need to change it so if the user is in dev environment, takes the user to dev login page and so on...**

Probability (how likely the bug is to happen, scored from 1-5): 5
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 5
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

**open the cif app in any environment and click the `Login Page` link at the bottom navigation.**

LindsayMacfarlane:
Thank you for your quick action on this bug @Sepehr-Sobhani!

One observation - when I select the login page link in the navigation bar, it now opens up the dev environment (as expected), but it takes the user to the home page and not the login page.

Sepehr-Sobhani:
> Thank you for your quick action on this bug @Sepehr-Sobhani!
> 
> One observation - when I select the login page link in the navigation bar, it now opens up the dev environment (as expected), but it takes the user to the home page and not the login page. I think it is supposed to open to the login page?

Thanks for mentioning that @LindsayMacfarlane . In this case, I'm assuming we need to end the users' session by clicking that link. Am I right @pbastia ?!

pbastia:
I'm wondering if a mistake snuck in on the wireframe - on other projects like CIIP or GGIRCS, that link is called "Home", this is probably what we want here (then the behaviour is indeed correct).

LindsayMacfarlane:
But then that is the same functionality as the 'Home' in the top navigation bar. I feel the purpose of the Login on the bottom navigation was to take the user back to the login page. @suhafa - can you confirm?

pbastia:
Wouldn't this mean that we're logging the user out, without the action being explicit anywhere?

suhafa:
[...]

PR #928, opened 2022-08-04
fix: change login page link based on environment

Sepehr-Sobhani:
ZH: https://github.com/bcgov/cas-cif/issues/925


pbastia (approved):
So quick!

PR #951, opened 2022-08-09
refactor: ReadonlyCalculatedWidget can be configured for any value from form context

pbastia:
This was preventing us from using multiple `ReadonlyCalculatedWidget` widgets on one form

### DIFF
--- a/app/lib/theme/widgets/ReadOnlyCalculatedValueWidget.tsx
+++ b/app/lib/theme/widgets/ReadOnlyCalculatedValueWidget.tsx
@@ -11,16 +11,19 @@ const ReadOnlyCalculatedValueWidget: React.FC<WidgetProps> = ({
   const isMoney = uiSchema?.isMoney;
   const isPercent = uiSchema?.isPercent;
 
+  const calculatedValue =
+    formContext[uiSchema.calculatedValueFormContextProperty];
+
   return (
     <dd>
-      {formContext.calculatedValue ? (
+      {calculatedValue ? (
         <NumberFormat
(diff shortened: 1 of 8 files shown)


app/lib/theme/widgets/ReadOnlyCalculatedValueWidget.tsx
```diff
@@ -11,16 +11,19 @@ const ReadOnlyCalculatedValueWidget: React.FC<WidgetProps> = ({
   const isMoney = uiSchema?.isMoney;
   const isPercent = uiSchema?.isPercent;
 
+  const calculatedValue =
+    formContext[uiSchema.calculatedValueFormContextProperty];
+
   return (
     <dd>
-      {formContext.calculatedValue ? (
+      {calculatedValue ? (
```
matthieu-foucault, line 19: Are we displaying `0` if there is no calculated value? How can users differentiate a `0` from a `null` value?

pbastia: Good catch, we probably want to leave that up to the user:
- only display the calculated value if one has been provided by the user
- allow falsy values to be entered as a calculated values

I'll make a quick update

BCerki: Looping @suhafa into this; we talked about using 0 for the emissions performance calculations specifically

BCerki (dismissed):

BCerki (dismissed):

BCerki (approved):

Issue #762, opened 2022-06-21
As an Ops Team Member, I want the status of a TEIMP report for a project to be automatically populated, so that I can easily recognize whether a project needs attention or not


matthieu-foucault:
#### Description:

As part of the emissions performance funding stream, funded agreements need to submit information related to project emission intensity to help show the impact to ghg reductions associated to a project. In order to support project status tracking of the emissions intensity reports, CIF Ops Team members would like to be able to see at a glance when a report is due to help with project management and alert them to when reports are late so that they can follow up with proponents.

Wireframe: https://www.figma.com/file/PzOY8RyQnVXnjZJdSv2AqQ/Wireframing_Figma?node-id=2430%3A2725

#### Acceptance Criteria:

Given I am a CIF administrator
When I am creating, editing or viewing an emissions intensity report
Then the Status of project TEIMP reporting is displayed at the top of the page

Given I am a CIF administrator
When I am creating, editing or viewing an emissions intensity report
Then I can see a badge with the status of the project

Given I am a CIF administrator
When I am creating, editing or viewing an emissions intensity report
And there are late reporting requirements
Then the status is "Late"

Given I am a CIF administrator
When I am creating, editing or viewing an emissions intensity report
And all the reporting requirements have been marked complete
Then the status is "Completed (MM DD YYYY)"

Given I am a CIF administrator
When I am creating, editing or viewing an emissions intensity report
And there are no upcoming reporting requirements
Then the status is "Not due"

Given I am a CIF administrator
When I am creating, editing or viewing an emissions intensity report
And the TEIMP end date has been reached
Then the status is either "Due in xx days" or "Due in xx weeks"

#### Development Checklist:

- [x] Project statuses include the following: Due in XX days, Due in xx weeks, Not due, Late, Completed (MM, DD, YYYY)
- [x] For Not Due - this status is only displayed until the TEIMP End Date is reached. The Report Due date would be 150 days after that. 
- [ ] Checklist item
- [ ] Meets the DOD

#### Notes:

- 

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

[...]

PR #952, opened 2022-08-09
Feat/teimp report status

Sepehr-Sobhani:
ZH: https://github.com/bcgov/cas-cif/issues/762

### DIFF
--- a/app/components/Form/EmissionIntensityReportStatus.tsx
+++ b/app/components/Form/EmissionIntensityReportStatus.tsx
@@ -42,8 +42,8 @@ export const EmissionIntensityReportStatus: React.FC<Props> = (props) => {
         );
       }
     } else if (
-      !isEmissionIntensityReportEnded &&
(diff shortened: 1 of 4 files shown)


app/components/Form/EmissionIntensityReportStatus.tsx
```diff
@@ -0,0 +1,90 @@
...
+      return (
+        <StatusBadge
+          variant="dueIn"
+          label={`Due in ${dueIn} ${dueIn > 1 ? "days" : "day"}`}
+        />
+      );
+    } else {
+      return <StatusBadge variant="none" />;
+    }
+  }, [props]);
```
pbastia, line 72: The `useMemo` is unnecessary here, the component would only be re-rendered if the props change anyways.

pbastia (dismissed):
Nice!! ...even too nice, there's code that can be removed ;)

matthieu-foucault (approved):

PR #953, opened 2022-08-10
Chore: TEIMP reporting duration

mikevespi:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/763

### DIFF
--- a/app/components/Form/ProjectEmissionIntensityReportForm.tsx
+++ b/app/components/Form/ProjectEmissionIntensityReportForm.tsx
@@ -56,7 +56,7 @@ export const createEmissionIntensityReportUiSchema = (
   );
   const reportDurationSuffix = reportDuration && (
     <b className="contentSuffix">
-      Duration: {reportDuration}{" "}
+      Duration: {reportDuration}
       <style jsx>{`
         .contentSuffix {
           font-size: 0.8rem;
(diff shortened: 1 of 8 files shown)


app/components/Form/ProjectEmissionIntensityReportForm.tsx
```diff
@@ -47,6 +50,21 @@ export const createEmissionIntensityReportUiSchema = (
     );
   };
 
+  const reportDuration = getDurationFromDates(
+    measurementPeriodStartDate,
+    measurementPeriodEndDate
+  );
+  const reportDurationSuffix = reportDuration && (
```
BCerki, line 57: Prefix instead of suffix in these names? "Duration" preceding the day count seems like the more meaningful part of the string

BCerki (commented):
I'll check this out again once CI is passing, but for now:
- would be nice to have a test for `getDurationFromDates.ts`
- would also be nice to have a couple assertions and/or tests in the existing date widget tests to cover duration

app/components/Form/ProjectEmissionIntensityReportForm.tsx
mikevespi: A good point! Suffix here is referring to the element being a suffix to the "widget content", so it's lining up with the ui property `contentSuffix`

Issue #904, opened 2022-07-26
Bug: adding and removing secondary contacts causes the revision to fail submission [6]


pbastia:
Describe the Bug:

When creating or editing a project that doesn't have contacts, adding secondary rows and then removing them causes the submission of the revision to fail.

Probability (how likely the bug is to happen, scored from 1-5): 2
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 4
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

- create a new project
- fill out the project overview
- go to the contacts page
  - click on "add a secondary contact"
  - click "delete" on that secondary contact row
  - submit the contacts form (empty primary contact at this point)
- go to the project "review and submit" and click "Submit"

=> there is an error thrown with no details provided

Screenshots:

Additional information:
The project_contact table is throwing a 'violates not-null constraint' on 'contact_id'
```
✔ 0 error(s) as cif_admin in 4.70ms :: mutation updateChangeReasonMutation($input: UpdateProjectRevisionInput!) { updateProjectRevision(input: $input) { projectRevision { id changeReason } } }        11:42:27
::ffff:127.0.0.1 - - [26/Jul/2022:18:42:27 +0000] "POST /graphql HTTP/1.1" 200 122 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:102.0) Gecko/20100101 Firefox/102.0"
2022-07-26 11:42:28.794 PDT [34804] ERROR:  null value in column "contact_id" of relation "project_contact" violates not-null constraint
2022-07-26 11:42:28.794 PDT [34804] DETAIL:  Failing row contains (1, 7, null, 1, 7, 2022-07-26 11:42:28.781687-07, 7, 2022-07-26 11:42:28.781687-07, null, null).
2022-07-26 11:42:28.794 PDT [34804] CONTEXT:  SQL statement "insert into cif.project_contact (project_id,contact_index) values ('7','1') returning id"
	PL/pgSQL function cif_private.commit_form_change() line 51 at EXECUTE
	SQL statement "update cif.form_change
	    set change_status = new.change_status
[...]

PR #956, opened 2022-08-10
chore: debug adding and removing secondary contact form

Sepehr-Sobhani:
ZH: https://github.com/bcgov/cas-cif/issues/904

### DIFF
--- a/app/components/Form/ProjectContactForm.tsx
+++ b/app/components/Form/ProjectContactForm.tsx
@@ -295,33 +295,33 @@ const ProjectContactForm: React.FC<Props> = (props) => {
     formChangeId: string,
     formChangeOperation: FormChangeOperation
   ) => {
-    discardFormChange({
-      formChange: { id: formChangeId, operation: formChangeOperation },
-      onCompleted: () => {
-        delete formRefs.current[formChangeId];
-        // Will remove the primary contact form change if it is not filled out and we have only one secondary contact form
-        // This is to fix the issue where the primary contact form change is not removed when the secondary contact is deleted
-        if (
-          !primaryContactForm.newFormData.contactId &&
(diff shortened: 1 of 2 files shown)


app/components/Form/ProjectContactForm.tsx
```diff
@@ -301,6 +291,39 @@ const ProjectContactForm: React.FC<Props> = (props) => {
...
+          !primaryContactForm.newFormData.contactId &&
+          alternateContactForms.length === 1
+        ) {
+          undoFormChanges({
+            variables: {
+              input: {
+                formChangesIds: formChangeIds,
+              },
+            },
+            onCompleted: () => {
```
pbastia, line 314: We've had issues in the past with nested `onCompleted` callbacks (mostly because the closure is not up-to-date with the form data once the first onComplete call resolves - as you noticed `alternateContactForms.length === 1` should actually be 0 at that point)

We should be able to decide beforehand:

```tsx
  const deleteContact = (
    formChangeId: string,
    formChangeOperation: FormChangeOperation
  ) => {
    if(primaryContactForm.newFormData.contactId ||  alternateContactForms.length > 1)
      discardFormChange(...)
    else
      undoFormChanges(...)
  }
```

pbastia (changes requested):
Thanks for tackling this, nice work with the formRef magic!

app/components/Form/ProjectContactForm.tsx
Sepehr-Sobhani: Absolutely!

pbastia: also, it'll be much easier to test 😁

pbastia (approved):
💯

PR #996, opened 2022-08-17
Fix/attention required status

BCerki:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/970


BCerki:
The db refactor fixed this bug; closing PR and card

Issue #995, opened 2022-08-17
Task: Revision Type database pre-work


pbastia:
Database tasks to support #992

- [x] A `revision_type` table in the database (similar to `cif.report_type`) containing the types:
    - "Amendment"
    - "General Revision"
    - "Minor Revision"
- [x] A `revision_type` field on the `project_revision` table referencing the types from the above table
- [x] A `comments` field on the `project_revision` table (varchar)
- [x] A migration adding the "General Revision" type to all existing revision records / form changes
- [x] Making sure the migration function is tested

pbastia:
@dleard feel free to add more to this card!

pbastia:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/995?planning-poker) with ZenHub @BCerki @dleard @gurjmatharu @matthieu-foucault @mikevespi @Sepehr-Sobhani

BCerki:
Re this checkbox: "A migration adding the "General Revision" type to all existing revision records / form changes", from Dylan:
>the `project_revision` table doesn't go through our `form_change` flow, so no need to worry about that part.

BCerki:
@LindsayMacfarlane, there's nothing to test on the front-end for this

PR #1001, opened 2022-08-18
Add revision type to db

BCerki:
Card: https://github.com/bcgov/cas-cif/issues/995

Notes:
- I noticed a missing column in the verify I wrote in a previous PR so I added it here
- I modelled the migration off https://github.com/bcgov/cas-ciip-portal/blob/develop/schema/deploy/tables/application_review_step.sql#L56-L74 (but I'm not dropping the function so we can test it, per Matthieu's suggestion)


dleard (changes requested):

schema/deploy/data_migration/data_migration_add_general_revision.sql
```diff
@@ -0,0 +1,27 @@
...
+  temp_row record;
+begin
+    alter table cif.project_revision disable trigger _100_committed_changes_are_immutable;
+
+    for temp_row in
+      select id from cif.project_revision
+     loop
+      update cif.project_revision
+      set revision_type='General Revision';
+    end loop;
```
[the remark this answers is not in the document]
BCerki: > This can be done with a single update statement & the function can be converted to pure SQL (instead of plgpsql)

You mentioned this the other day and I forgot to ask--why is SQL better than plgpsql?

schema/deploy/data_migration/add_general_revision.sql
```diff
@@ -0,0 +1,21 @@
+-- Deploy cif:data_migration/add_general_revision to pg
+
+begin;
+
+create or replace function cif.add_general_revision()
+  returns void as $$
+declare
+  temp_row record;
+begin
+    alter table cif.project_revision disable trigger _100_committed_changes_are_immutable;
```
matthieu-foucault, line 10: disable the timestamps trigger to not alter the `updated_by`

schema/deploy/data_migration/add_general_revision.sql
```diff
@@ -0,0 +1,21 @@
...
+
+update cif.project_revision set revision_type = 'General Revision'
```
matthieu-foucault, line 12:
```suggestion
update cif.project_revision set revision_type = 'General Revision' where revision_type is null;
```

schema/deploy/tables/project_revision_001.sql
```diff
@@ -0,0 +1,10 @@
+-- Deploy cif:tables/project_revision_001 to pg
+
+begin;
+
+alter table cif.project_revision add column revision_type varchar(1000) not null references cif.revision_type(type), add column comments varchar(10000);
```
matthieu-foucault, line 5: This will not work, given that the `project_revision` table contains committed revisions. We need to set a default value. Triggers need to be disabled and re-enabled here

schema/deploy/tables/project_revision_001.sql
```diff
@@ -0,0 +1,10 @@
+-- Deploy cif:tables/project_revision_001 to pg
```
matthieu-foucault, line 1:
```suggestion
-- Deploy cif:tables/project_revision_001_add_revision_type to pg
```

matthieu-foucault (changes requested):

schema/deploy/tables/project_revision_001.sql
BCerki: Now that there's a default value, do we even need the `add_general_revision` function?

schema/deploy/tables/project_revision_001.sql
```diff
@@ -0,0 +1,16 @@
+-- Deploy cif:tables/project_revision_001_add_revision_type to pg
+
+begin;
+
+alter table cif.project_revision disable trigger _100_committed_changes_are_immutable, disable trigger _100_timestamps;
+
+alter table cif.project_revision
```
BCerki, line 7: add a comment to explain not null + default

schema/deploy/tables/revision_type.sql
```diff
@@ -0,0 +1,43 @@
...
+
+-- Grant cif_internal permissions
+perform cif_private.grant_permissions('select', 'revision_type', 'cif_internal');
+perform cif_private.grant_permissions('insert', 'revision_type', 'cif_internal');
+perform cif_private.grant_permissions('update', 'revision_type', 'cif_internal');
+
+-- Grant cif_admin permissions
+perform cif_private.grant_permissions('select', 'revision_type', 'cif_admin');
+perform cif_private.grant_permissions('insert', 'revision_type', 'cif_admin');
+perform cif_private.grant_permissions('update', 'revision_type', 'cif_admin');
```
pbastia, line 25: We probably only need `select` permission here, at least while we don't have an admin/management view for this

```suggestion
-- Grant cif_internal permissions
perform cif_private.grant_permissions('select', 'revision_type', 'cif_internal');

-- Grant cif_admin permissions
perform cif_private.grant_permissions('select', 'revision_type', 'cif_admin');
```

schema/test/unit/computed_columns/project_revision_project_manager_form_changes_by_label_test.sql
```diff
@@ -52,15 +52,21 @@ insert into cif.project(id, operator_id, funding_stream_rfp_id, project_status_i
     (5, 1, 1, 1, '005', 'summary', 'project 1'),
     (6, 1, 1, 1, '006', 'summary', 'project 1');
 
-insert into cif.project_revision(id, change_status, change_reason, project_id)
+insert into cif.revision_type (type)
+  values
+    ('Amendment'),
+    ('General Revision'),
+    ('Minor Revision');
```
pbastia, line 59: These should be set by the migration, do we truncate them?

schema/test/unit/tables/project_revision_test.sql
```diff
@@ -46,6 +49,100 @@ select throws_ok(
...
+insert into cif.project_revision(project_id, revision_type)
+values
+  (3, 'Amendment'),
+  (4, 'Minor Revision');
+
+select is(
+  (select count(*) from cif.project_revision where revision_type='General Revision'),
+  0::bigint,
+  'project_revision_001_add_revision_type does not change the revision_type of projects that already have revision types');
+
```
pbastia, line 145: I'm a little confused about what we test here, the migration adds the `revision_type` column, so making sure that existing records with a revision_type don't get impacted has little sense. We can probably remove this test

schema/verify/tables/revision_type.sql
```diff
@@ -0,0 +1,18 @@
...
+
+-- cif_internal Grants
+select cif_private.verify_grant('select', 'revision_type', 'cif_internal');
+select cif_private.verify_grant('update', 'revision_type', 'cif_internal');
+select cif_private.verify_grant('insert', 'revision_type', 'cif_internal');
+
+-- cif_admin Grants
+select cif_private.verify_grant('select', 'revision_type', 'cif_admin');
+select cif_private.verify_grant('insert', 'revision_type', 'cif_admin');
+select cif_private.verify_grant('update', 'revision_type', 'cif_admin');
```
pbastia, line 15:
```suggestion
-- cif_internal Grants
select cif_private.verify_grant('select', 'revision_type', 'cif_internal');

-- cif_admin Grants
select cif_private.verify_grant('select', 'revision_type', 'cif_admin');
```

pbastia (changes requested):
Some unnecessary permissions, looks great otherwise!

schema/test/unit/computed_columns/project_revision_project_manager_form_changes_by_label_test.sql
BCerki: I think it's getting truncated as part of a cascade? If I don't include these here I get this error: `DETAIL:  Key (revision_type)=(General Revision) is not present in table "revision_type".`

pbastia: This happens because this test does a
```
truncate table cif.cif_user restart identity cascade;
```
which deletes all records in the database where the `created_by` or `updated_by` is set - including our `cif.revision_type` records. This won't be happening on our deployed instances since a superuser runs the script, but this is not the case when we run tests locally.

Instead, we should leave the `cif_user` table alone since it already contains basic users. If you need to control exactly what is in there you can update its values like this 
```
insert into cif.cif_user(id, uuid, etc...)
  overriding system value
  values (1, '11111111-1111-1111-1111-111111111111', etc...);
```

pbastia: We would want to refactor this, otherwise we don't test that our records in `revision_type` are inserted properly, and rely on extra code instead. this is also the case for the `cif.change_status` insert in this test

pbastia (changes requested):
Just a small refactor to make sure we don't have to re-add records that are part of a migration

schema/test/unit/computed_columns/project_revision_project_manager_form_changes_by_label_test.sql
```diff
@@ -5,42 +5,28 @@ select plan(7);
...
-  ('00000000-0000-0000-0000-000000000003', 'user 4', 'Testuser', 'cif@somemail.com');
-
 set jwt.claims.sub to '00000000-0000-0000-0000-000000000000';
 
 insert into cif.operator(id, legal_name)
   overriding system value
   values (1, 'test operator');
 
-insert into cif.funding_stream (name, description) values ('EP', 'Emissions Performance'), ('IA', 'Innovation Accelerator');
+-- insert into cif.funding_stream (name, description) values ('EP', 'Emissions Performance'), ('IA', 'Innovation Accelerator');
```
pbastia, line 17: comment can be deleted

pbastia (dismissed):
Just a comment that can be deleted. Looks great!

pbastia (approved):

PR #1009, opened 2022-08-19
refactor: wip on tasklist config

gurjmatharu:


app/components/TaskList/index.tsx
```diff
@@ -121,215 +121,269 @@ const TaskList: React.FC<Props> = ({ projectRevision, mode }) => {
...
-            formStatus={projectContactsStatus}
-            currentStep={currentStep}
-            mode={mode}
-          />
-        </TaskListSection>
+  const projectOverviewSection = (step: string) => {
+    return (
+      <TaskListSection
+        defaultExpandedState={
+          currentStep === "0" ||
```
pbastia, line 128: These steps are still hardcoded in the different components. What we want to achieve with this refactor is to make everything generated, including links and navigation, so we can easily reorder, add, remove or hide items without breaking magic numbers in different places.
There is definitely a bit of research to be done on how to do that, I'm happy to talk about it one of these days to find out what the best path would be for the implementation

gurjmatharu: Agreed. Tried a few different approaches and didn't get a good outcome (passing step, and currentStep into the config).  Discussing this would be good.

PR #1023, opened 2022-08-23
Feat/optional contact phone number

BCerki:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/869

re this bullet in the developer checklist: `Phone number is not enforced in the database (can be null)`--I didn't have to change anything, it was already set up like that


Sepehr-Sobhani (approved):
💯

Issue #998, opened 2022-08-18
Task: Amendment type database pre-work


pbastia:
Pre-work for #983

- [x] a new `amendment_type` table with `name : varchar(1000)` as primary key
- [x] `amendment_type` table is pre-populated with the values (Schedule, Scope, Cost)
- [x] a `project_revision_amendment_type` join table with:
    - `id` pkey
    - `project_revision_id` fkey to `project_revision(id)`
    - `amendment_type` fkey to `amendment_type(name)`
- [x] the `project_revision_amendment_type` has unique indexes to make sure that a revision only has at most one amendment_type record of each type

pbastia:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/998?planning-poker) with ZenHub @BCerki @dleard @gurjmatharu @matthieu-foucault @mikevespi @Sepehr-Sobhani

Sepehr-Sobhani:
@LindsayMacfarlane Moved this task into `Accepted by PO or Dev` as it contains only code changes and nothing to test.

PR #1024, opened 2022-08-23
Feat/amendment type pre work

Sepehr-Sobhani:
ZH: https://github.com/bcgov/cas-cif/issues/998


pbastia (changes requested):
Looks great! Just a few extra permissions we can remove

pbastia (approved):

pbastia (changes requested):
Do you mind adding a check for "Changes Saved." just before the happo screenshot for budgets "with errors" ? We do have a happo diff that looks mid-flight

pbastia (approved):

Issue #987, opened 2022-08-16
As a CIF user, I want to filter projects only by Primary Managers


nanyangpro:
#### Description:
On the CIF Projects page, currently if filtered by Project Managers, the table displays all projects which have the person as either Primary Manager or Secondary Manager. Users want to see projects which have the person only as Primary Manager instead.

#### Acceptance Criteria:

Given I am a CIF Administrator
When I am on the Project List page
And I am filtering on the Project Manager column
Then only the primary manager(s) are displayed / available for filtering. 

#### Development Checklist:

- [x] The project managers column is renamed to `primary project managers`
- [x] The query in app/pages/cif/project/[project]/index.tsx filters to only include primary managers (This can probably be done with something like: ```filter: {projectManagerLabelByProjectManagerLabelId: {label: {includesInsensitive: "primary"}}}``` )
- [x] The project managers column should only show primary managers
- [ ] Meets the DOD

filter: {projectManagerLabelByProjectManagerLabelId: {label: {includesInsensitive: "primary"}}}
#### Notes:

- Do we need to update the column heading?

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

LindsayMacfarlane:
During CIF Prioritization session on August 16, CIF team indicated that this is an easy win of high impact and medium/high urgency.

LindsayMacfarlane:
[...]

PR #1050, opened 2022-08-30
Chore/project list by primary manager

Sepehr-Sobhani:
ZH: https://github.com/bcgov/cas-cif/issues/987

### DIFF
--- a/app/pages/cif/projects.tsx
+++ b/app/pages/cif/projects.tsx
@@ -51,11 +51,13 @@ export const ProjectsQuery = graphql`
         }
         projectManagersByProjectId: {
           some: {
-            cifUserByCifUserId: {
-              fullName: { includesInsensitive: $primaryProjectManager }
-            }
-            projectManagerLabelByProjectManagerLabelId: {
-              label: { includesInsensitive: "primary" }
+            and: {
+              cifUserByCifUserId: {
+                fullName: { includesInsensitive: $primaryProjectManager }
+              }
+              projectManagerLabelByProjectManagerLabelId: {
+                label: { includesInsensitive: "primary" }
+              }
             }
           }
         }
--- a/app/pages/cif/projects.tsx
+++ b/app/pages/cif/projects.tsx
@@ -70,18 +70,20 @@ export const ProjectsQuery = graphql`
         }
       }
     }
-    allProjectManagers(
+    allCifUsers(
       filter: {
-        projectManagerLabelByProjectManagerLabelId: {
-          label: { includesInsensitive: "primary" }
+        projectManagersByCifUserId: {
+          some: {
+            projectManagerLabelByProjectManagerLabelId: {
+              label: { includesInsensitive: "primary" }
+            }
+          }
         }
       }
     ) {
       edges {
         node {
-          cifUserByCifUserId {
-            fullName
-          }
+          fullName
         }
       }
     }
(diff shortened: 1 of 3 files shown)


app/pages/cif/projects.tsx
```diff
@@ -67,10 +70,18 @@ export const ProjectsQuery = graphql`
         }
       }
     }
-    allCifUsers {
+    allProjectManagers(
```
BCerki, line 73: I don't totally follow what was previously going on here--why was the query using users instead of managers?

BCerki (commented):
A couple notes:
- add/update a test for the filter? There are some filter tests in `app/tests/unit/components/Table/Filters/`
- ~~I'm not clear on the AC/UX for this one so I'm going to put my questions over in the card so Nan can weigh in more easily~~ Nevermind, the comments clarify. The column shouldn't show the secondary managers anymore, currently it does: https://www.loom.com/share/03722ab80f8b43d38d89d3b426eb6ee4

app/pages/cif/projects.tsx
Sepehr-Sobhani: I'm not sure either; When I was trying to search for a name, I found my name there too while I'm not a manager. but since AC only requires primary managers to show up in the search list and that's why I changed it to allProjectManagers.

Sepehr-Sobhani:
> * add/update a test for the filter? There are some filter tests in `app/tests/unit/components/Table/Filters/`

I think those tests are mainly for general use cases, if we want to test this one we need to test the whole `projects.tsx` component.

Sepehr-Sobhani:
> * I'm not clear on the AC/UX for this one so I'm going to put my questions over in the card so Nan can weigh in more easily

Sounds good 👍

app/pages/cif/projects.tsx
BCerki: Actually, late-breaking thought, it might be because we don't have a way (as far as I know) for users to set up managers. That query comes back with April Ludgate etc. because we set her up as a manager in the dev data, but we might get no one back in test/prod. @pbastia, thoughts?

Sepehr-Sobhani:
> * ~I'm not clear on the AC/UX for this one so I'm going to put my questions over in the card so Nan can weigh in more easily~ Nevermind, the comments clarify. The column shouldn't show the secondary managers anymore, currently it does: https://www.loom.com/share/03722ab80f8b43d38d89d3b426eb6ee4

Thanks for mentioning that @BCerki. The code is now updated.

app/pages/cif/projects.tsx
pbastia: Something like this will do what you want:
```graphql
query MyQuery {
  allCifUsers(
    filter: {projectManagersByCifUserId: {some: {projectManagerLabelByProjectManagerLabelId: {label: {includesInsensitive: "primary"}}}}}
  ) {
    edges {
      node {
        fullName
      }
    }
  }
}
```

app/pages/cif/projects.tsx
```diff
@@ -52,7 +52,10 @@ export const ProjectsQuery = graphql`
         projectManagersByProjectId: {
           some: {
             cifUserByCifUserId: {
-              fullName: { includesInsensitive: $projectManagers }
+              fullName: { includesInsensitive: $primaryProjectManager }
+            }
+            projectManagerLabelByProjectManagerLabelId: {
+              label: { includesInsensitive: "primary" }
             }
           }
```
pbastia, line 60: I don't know what the default behaviour is for graphql, but it wouldn't hurt to specify the `and` logic between both conditions:
```graphql
some: {
   and: {
    cifUserByCifUserId:{...}
    projectManagerLabelBy...
  }
}
```

pbastia (changes requested):
one last little thing!

pbastia (approved):

pbastia (approved):

PR #1055, opened 2022-08-31
Chore/db amendment status

BCerki:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/999

### DIFF
--- a/schema/verify/tables/project_revision_002.sql
+++ b/schema/verify/tables/project_revision_002.sql
@@ -1,0 +1,9 @@
+-- Verify cif:tables/project_revision_002.sql on pg
+
+begin;
+
+select column_name
+from information_schema.columns
+where table_schema='cif' and table_name='project_revision' and column_name='amendment_status';
+
+rollback;
--- a/app/schema/schema.graphql
+++ b/app/schema/schema.graphql
@@ -1391,6 +1391,860 @@ enum AdditionalFundingSourcesOrderBy {
   UPDATED_BY_DESC
 }
 
+"""Table containing information about amendment statuses"""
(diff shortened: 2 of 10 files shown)


schema/verify/tables/project_revision_002.sql
```diff
@@ -0,0 +1,9 @@
+-- Verify cif:tables/project_revision_002.sql on pg
+
+begin;
+
+select column_name
+from information_schema.columns
+where table_schema='cif' and table_name='project_revision' and column_name='amendment_status';
```
pbastia, line 7: This will not fail if the column doesn't exist, since we'll be returning null instead of throwing (see [docs](https://sqitch.org/docs/manual/sqitchtutorial/#trust-but-verify))

We could do something like that (I'm not a fan of the divide by zero trick, even though it's concise):

```suggestion
do $$
begin
  assert (select column_name
  from information_schema.columns
  where table_schema='cif' and table_name='project_revision' and column_name='amendment_status')
  is not null;
end $$;
```

pbastia (changes requested):

schema/verify/tables/project_revision_002.sql
BCerki: I've seen this in a few of the revert files--I've made a tech debt issue to check them out: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/1049

pbastia (approved):
🥇

Issue #942, opened 2022-08-08
As a CIF administrator, I want to see the total project value as a calculated field on the "Budget, expenses and payments" form, so that I have context for all other payment amounts


pbastia:
#### Description:

A "Total Project Value" calculated field needs to be added to the "Budget, expenses and payments" form.

It should be calculated with the following formula:
Total Project Value = Maximum Funding Amount + Proponent Cost + Additional Funding Amount(s)

[Formula with examples](https://bcgov.sharepoint.com/:w:/t/00608-ScrumTeam/EcXm1DnetBZNgFOkakDR_IEB1RS8r51PTHqH2h4C5L6URw?e=RbiNV2) - Please open it with the Word app instead of browser for formatting and better readability.

[Design](https://www.figma.com/file/PzOY8RyQnVXnjZJdSv2AqQ/CIF-Design?node-id=2898%3A2863&t=0b38GJFZh0GzOQrz-1) in Figma

#### Acceptance Criteria:

Given I am a CIF administrator
When I create or edit a project
And I am on the "budget, expenses, and payments" form
Then I can see a "Total Project Value" calculated field

Given I am a CIF administrator
When I am on the "Budget, expenses and payments form" for a project revision
Then the "Total Project Value" field is auto-calculated
And the formula is: Total Project Value = Maximum Funding Amount + Proponent Cost + Additional Funding Amount(s)

#### Development Checklist:

- [x] Computed column `project_revision_total_project_value` returns the right value (numeric)
- [x] The field displaying the value should use the `MoneyWidget` and be a `ReadonlyCalculatedValueWidget`
- [x] Remove existing `total_project_value` from `funding_parameter` table and update tests
- [x] Add custom discard mutation for additional funding sources
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [x] User story is included
- [x] User role and type are identified
- [x] Acceptance criteria are included
- [x] Wireframes are included (if required)
- [x] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [x] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
[...]

PR #1059, opened 2022-08-31
Chore: milestone form refactor

dleard:
- [x] staging and removing form changes
- [ ] custom update mutation (sql) - We will deal with this when we implement the calculated budget values (#942)
- [ ] custom mutation will need to scrub the new_form_data from expenses-related keys if the selected report type doens't have expenses
- [x] validate_record is not verifying as-is (schema is on the back-end now)
- [x] hasExpenses value is not properly updated. We need a way to efficiently set that value in the formData
- [x] hasExpenses field should be properly hidden, the title currently shows up and space is made on the form for it
- [x] tests for the hasExpenses logic
- [x] ProjectMilestoneReportFormSummary is doing an early `return null` because it is still relying on 3 form_change records & breaks. It will need similar refactoring.
- [x] ProjectMilestoneReportFormGroup is what is being refactored, it should be renamed at some point (as well as all fragments and references in places like the taskList index).
- [x] fix dev data
- [x] fix anchors from tasklist to directly navigate to a specific milestone
- [x] tests for `ProjectMilestoneReportForm` and summary components need to use the same json schema as is deployed to the database.
- [x] migrate existing revisions with the tuple of form_changes (reporting_requirement, milestone_report, payment) to only have the single form_change with all the data. + good tests
- [x] `commit_form_change_internal` should use the custom handler for the milestone form
- [x] fix test for `handle_milestone_form_change_commit` (add a project form_change with a record_id)
- [x] create_project_revision needs refactoring. It should no longer reference the schemas 'milestone_report' or 'payment'. It should also differentiate 'reporting_requirement' table records that use a 'reporting_requirement' schema from the records that use the 'milestone' schema.
- [x] rebase: create `sqitch rework` items for changes made since the rc5 release

--- optional ---
- [x] can be removed: add_milestone_to_revision & discard_milestone_form_change mutations (both front-end and sql)
- [x] can be removed: ProjectMilestoneReportForm component + projectMilestoneFormFunctions


Sepehr-Sobhani:
@pbastia As discussed:

- Changing milestone type from Performance to Reporting causes error on the summary page: 
https://www.loom.com/share/430f35efbb514cde8235476df079dbdd

gurjmatharu:
The tasklist status is not accurate when the first milestone is filled, and 2nd and third need attention.
Screen Shot 2022-10-27 at 10 36 40 AM

Sepehr-Sobhani (approved):
🥳

Issue #1042, opened 2022-08-30
Project Contacts: removing secondary contact will bring back removed primary contact [4]


Sepehr-Sobhani:
Describe the Bug:

removing a primary contact, then adding a secondary contact, then removing that secondary contact adds the previously removed primary contact back in

Note: This is caused by calling undoFormChangeMutation when removing secondary contact (when we only have one secondary contact)

Probability (how likely the bug is to happen, scored from 1-5): 2
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 2
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

Go to an existing project that already has a primary project contact -> go to the project contacts section -> clear primary contact -> add secondary contact -> remove secondary contact (primary contact is back again)

Screenshots:

https://www.loom.com/share/67b4aca0cb8d4ad196eec7fc944ace22

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

LindsayMacfarlane:
No longer able to recreate the bug. Thank you @Sepehr-Sobhani !

PR #1060, opened 2022-09-01
fix: replace undo change mutation with discard change mutation to fix project contact issue

Sepehr-Sobhani:
ZH: https://github.com/bcgov/cas-cif/issues/1042

### DIFF
--- a/app/tests/unit/components/Form/ProjectContactForm.test.tsx
+++ b/app/tests/unit/components/Form/ProjectContactForm.test.tsx
@@ -465,30 +465,26 @@ describe("The ProjectContactForm", () => {
       componentTestingHelper.environment.mock.getAllOperations()
     ).toHaveLength(3);
 
-    const firstMutationUnderTest =
-      componentTestingHelper.environment.mock.getAllOperations()[1];
-
-    expect(firstMutationUnderTest.fragment.node.name).toBe(
-      "deleteFormChangeWithConnectionMutation"
+    componentTestingHelper.expectMutationToBeCalled(
+      "deleteFormChangeWithConnectionMutation",
+      ...[
+        {
+          connections: [
+            'client:Test Project Revision ID:__connection_projectContactFormChanges_connection(formDataTableName:"project_contact")',
+          ],
+          input: {
+            id: "Form ID 1",
+          },
+        },
+        {
+          connections: [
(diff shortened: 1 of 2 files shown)


app/tests/unit/components/Form/ProjectContactForm.test.tsx
```diff
@@ -463,18 +463,31 @@ describe("The ProjectContactForm", () => {
...
+    expect(secondMutationUnderTest.fragment.node.name).toBe(
+      "deleteFormChangeWithConnectionMutation"
+    );
+
+    expect(secondMutationUnderTest.request.variables).toMatchObject({
       input: {
-        formChangesIds: [1, 2],
+        id: "Form ID 2",
       },
     });
```
pbastia, line 492: we should refactor this using the testing helper, something like that (you will probably need the connection in there, or use an `expect.objectContaining({...})` )
```ts
componentTestingHelper.expectMutationToBeCalled(
  "deleteFormChangeWithConnectionMutation",
  {
    variables: {
      input: {
        id: "Form ID 1"
      }
    }
  });
```

pbastia (changes requested):

app/tests/unit/components/Form/ProjectContactForm.test.tsx
```diff
@@ -465,30 +465,26 @@ describe("The ProjectContactForm", () => {
...
+        {
+          connections: [
+            'client:Test Project Revision ID:__connection_projectContactFormChanges_connection(formDataTableName:"project_contact")',
+          ],
+          input: {
+            id: "Form ID 2",
+          },
+        },
+      ]
     );
```
pbastia, line 488: We should be testing the 2 mutations separately, I think giving an array in the `variables` parameter of the helper confuses it

```suggestion
    componentTestingHelper.expectMutationToBeCalled(
      "deleteFormChangeWithConnectionMutation",
        {
          connections: expect.any(Array),
          input: {
            id: "Form ID 1",
          },
        }
    );
    
   componentTestingHelper.expectMutationToBeCalled(
      "deleteFormChangeWithConnectionMutation",
        {
          connections: expect.any(Array),
          input: {
            id: "Form ID 2",
          },
        }
    );
```

pbastia (changes requested):

pbastia (approved):

PR #1067, opened 2022-09-03
Feat/funding additional dates

Sepehr-Sobhani:
[ZH Card](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/1020)


app/cypress/support/commands.js
```diff
@@ -640,3 +653,36 @@ Cypress.Commands.add(
       .should("have.text", status);
   }
 );
+
+// TODO: possible candidate to replace other date related commands
+Cypress.Commands.add("setDate", (date, ariaLabel, reportNumber = 0) => {
```
pbastia, line 658: for a better flowing API, we should be more precise in the name and invert the arguments:
```suggestion
Cypress.Commands.add("setDateInPicker", (ariaLabel, date, reportNumber = 0) => {
```

this way we can call it with something like
```js
cy.setDateInPicker("Contract Start Date", "2022-01-01");
```

app/cypress/support/commands.js
```diff
@@ -640,3 +653,36 @@ Cypress.Commands.add(
...
+  );
+  cy.get(`.react-datepicker__day--0${receivedDateTZ.toFormat("dd")}`)
+    .not(`.react-datepicker__day--outside-month`)
+    .click();
+  cy.get(`[aria-label*="${ariaLabel}"]`)
+    .eq(reportNumber - 1)
+    .contains(`${receivedDateTZ.toFormat("MMM dd, yyyy")}`);
+  cy.contains("Changes saved").should("be.visible");
+  // need to return a Cypress promise (could be any cy. command) to let Cypress know that it has to wait for this call
+  return cy.url().should("include", "/form");
```
pbastia, line 687: This will prevent us from using this helper on any page that does not include `/form` in the URL

schema/verify/tables/funding_parameter_002_add_auto_generating_report_date_columns.sql
```diff
@@ -0,0 +1,13 @@
...
+
+begin;
+
+select column_name
+from information_schema.columns
+where table_schema='cif' and table_name='funding_parameter' and column_name='contract_start_date';
+
+select column_name
+from information_schema.columns
+where table_schema='cif' and table_name='funding_parameter' and column_name='project_assets_life_end_date';
```
pbastia, line 11: These will not fail but merely return `null` if the column doesn't exist, and won't cause the `verify` script to fail.

Something like that would fail instead:
```sql
do $$
  begin
    assert (
      select column_name
      from information_schema.columns
      where table_schema='cif' and table_name='funding_parameter' and column_name='contract_start_date'
    ), 'column "contract_start_date" is not defined on table "funding_parameter"';
  end;
$$;
```

pbastia (changes requested):
just a couple of things, looks great!

schema/verify/tables/funding_parameter_002_add_auto_generating_report_date_columns.sql
pbastia: >  I think we need to change the `select column_name` to `select true`

yes, that's right

For the refactor, we can probably do this in a separate PR but that would be good :)

pbastia (approved):
very nice 👌

PR #1072, opened 2022-09-07
chore: add idempotent insert in form schema after each deploy

Sepehr-Sobhani:
No card associated


Makefile
```diff
@@ -135,19 +135,16 @@ deploy_db_migrations:
 	@$(SQITCH) --chdir mocks_schema deploy
 
 deploy_dev_data: ## deploy the database migrations with sqitch and load the data for local development
-deploy_dev_data: deploy_db_migrations
+deploy_dev_data: deploy_db_migrations deploy_prod_data
```
pbastia, line 138: we'll want to guarantee the order here:
```suggestion
deploy_dev_data: | deploy_db_migrations deploy_prod_data
```

app/cypress/integration/cif/project-revision/index.spec.js
```diff
@@ -252,6 +252,11 @@ describe("the new project page", () => {
     cy.findByRole("button", { name: /^submit/i }).click();
     cy.get(".error-detail").should("have.length", 4);
     cy.contains("Changes saved").should("be.visible");
+    // below assertion is not necessary, but it's here to make sure the happo
+    // snapshot is taken after the form is submitted
+    cy.findByText(/Add emissions intensity report/i)
+      .next()
+      .should("have.text", "Attention Required");
```
pbastia, line 259: nice.

schema/data/deploy-data.sh
```diff
@@ -0,0 +1,43 @@
...
+
+__dirname="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+pushd "$__dirname" > /dev/null
+
+_psql() {
+  psql -d "$database" -h "$host" -p "$port" -U "$user" -qtA --set ON_ERROR_STOP=1 "$@" 2>&1
+}
+
+deployProdData() {
+  _psql -f "./prod/001_form.sql"
```
pbastia, line 39: Should we make this run all the files in `/prod/` ? Or do we prefer constraining which files go in there?

pbastia (changes requested):

schema/data/deploy-data.sh
Sepehr-Sobhani: I have seen this in ciip too but in there we are constraining which files to go. I'm not sure how many more insert SQL files we might need to add here but if it gets too much we might want to run all files within `/prod/` in a proper way. anyway, it's just my thoughts, what do you prefer?!

pbastia: We can leave this as is, and refactor when needed

pbastia (approved):

PR #1107, opened 2022-09-27
Refactor/date widgets

BCerki:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/786

Notes:
- the happo diffs seem to have some false positives--they're flagging fields that look the same to me
- Previously, dates that weren't due or received dates were surrounded by brackets. Are they supposed to be? @nanyangpro or @suhafa ? Eg (left is previous, right is new)
image


pbastia (approved):
👌

PR #1148, opened 2022-10-25
feat: new amendments page

gurjmatharu:


Sepehr-Sobhani (commented):
Great work Gurjeet 🙏

app/data/jsonSchemaForm/projectRevisionCreateSchema.ts
```diff
@@ -0,0 +1,19 @@
...
+      enum: ["Amendment", "General Revision", "Minor Revision"],
+    },
+  },
+};
+
+export const projectRevisionCreateUISchema = {
+  revisionType: {
+    "ui:widget": "radio",
+  },
+};
```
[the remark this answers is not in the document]
pbastia: Eventually these will live in the database, so it makes sense to keep them the same if we can. the enum values should be pulled from the database from the `cif.revision_type` table

schema/revert/mutations/create_project_revision_001.sql
```diff
@@ -0,0 +1,104 @@
+-- Revert cif:mutations/create_project_revision_001 from pg
+
+begin;
+
+drop function cif.create_project_revision(integer, varchar(1000));
+
+begin;
```
pbastia, line 7:
```suggestion

```

schema/test/unit/mutations/create_project_revision_test.sql
```diff
@@ -62,10 +62,10 @@ select cif.commit_project_revision(1);
...
 select results_eq(
   $$
-  select id, project_id, change_status from cif.create_project_revision(1)
+  select id, project_id, change_status, revision_type from cif.create_project_revision(1, 'Amendment')
   $$,
   $$
-  values (2, 1, 'pending'::varchar)
+  values (2, 1, 'pending'::varchar, 'Amendment'::varchar)
   $$
 );
```
pbastia, line 70: Test is missing text about what it is testing

pbastia: We're also missing a test about the default value if the revision type is not provided

pbastia (changes requested):
Looks good! Just a few things here and there

app/pages/cif/project-revision/[projectRevision]/create.tsx
```diff
@@ -0,0 +1,97 @@
...
+          getProjectRevisionFormPageRoute(
+            response.createProjectRevision.projectRevision.id,
+            0
+          )
+        );
+      },
+    });
+  };
+  const revisionEnum = allRevisionTypes.edges.map((e) => e.node.type);
+  projectRevisionSchema.properties.revisionType.enum = revisionEnum;
```
Comment (pbastia) [inline app/pages/cif/project-revision/[projectRevision]/create.tsx:71]: This is actually modifying the singleton in place, we might want to grab a copy of the schema instead:
```ts
  const localSchema = JSON.parse(JSON.stringify(projectRevisionSchema));
  localSchema.properties.revisionType.enum = revisionEnum;
```

pbastia (approved):
Looks good to me! Great work!

app/pages/cif/project-revision/[projectRevision]/create.tsx
Comment (gurjmatharu) [reply app/pages/cif/project-revision/[projectRevision]/create.tsx:71]: Noted.  I'll add this change to my next PR.

PR #1140, opened 2022-10-17
Fix/report auto generation timezone

Sepehr-Sobhani:
[ZH Card](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/1136)

PR #1161, opened 2022-11-02
chore: show date fields on report auto-generator component

Sepehr-Sobhani:
[ZH Card](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/1157)

- Will add e2e test once #1140 is merged.

### DIFF
--- a/app/components/Form/ProjectAnnualReportForm.tsx
+++ b/app/components/Form/ProjectAnnualReportForm.tsx
@@ -178,26 +178,27 @@ const ProjectAnnualReportForm: React.FC<Props> = (props) => {
         reportDueFormChange={projectRevision.upcomingAnnualReportFormChange}
       />
       <FormBorder>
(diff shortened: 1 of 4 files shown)


Sepehr-Sobhani:
> Looking good! Just one note--this isn't explicit in the AC so I'm not sure if it should be addressed, but the "Generate Reports" heading shows up on the page: Screenshot from 2022-11-07 12-11-28 Should that be removed too since the button is gone and you can't generate anymore?

Thanks for bringing this up @BCerki! I'm not sure either. 
@nanyangpro would you please take a look at this?!

nanyangpro:
Good call, @BCerki - Yes, once the reports are generated, the title and the button should all be hidden to avoid confusions. We just want to always show the dates to users.

BCerki (approved):

Issue #984, opened 2022-08-16
As an Ops Team member, I want to see calculated expenses and payments to date when I enter milestone information, so that I have context for the other calculated values


pbastia:
#### Description:

showing:
- total eligible expenses to date (calculated) + Tooltip
- total gross payment amount to date (calculated) + Tooltip
- total net payment amount to date (calculated) + Tooltip

https://www.figma.com/file/PzOY8RyQnVXnjZJdSv2AqQ/Wireframing_Figma?node-id=3713%3A2681
more details in #656

#### Acceptance Criteria:

Given I am an ops team member
When I enter milestone information
Then I can see calculated values for:
- total eligible expenses to date
- total gross payment amount to date
- total net payment amount to date

Given I am an ops team member
When I enter milestone information
And I look at the calculated fields above
Then I can see contextual help tooltips on each of these fields

#### Development Checklist:

- [x] check in with @dleard to see what of the following was done as part of #986
- [ ] create `reporting_requirement_net_payment_amount_to_date`
- [ ] create `reporting_requirement_gross_payment_amount_to_date` 
- [x] Create a `project_revision_total_eligible_expenses_to_date` computed column (done in #986)
- [ ] Add fields for these calculated values in the rjsf using a `ReadOnlyCalculatedValue` widget
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [x] User story is included
- [x] User role and type are identified
- [x] Acceptance criteria are included
- [x] Wireframes are included (if required)
- [x] Design / Solution is accepted by Product Owner
- [x] Dependencies are identified (technical, business, regulatory/policy)
- [x] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

LindsayMacfarlane:
[...]

Issue #717, opened 2022-06-08
As an Ops Team Member, I want to see the expenses and payments to date on the Budgets, Expenses & Payments form


nanyangpro:
#### Description:
[Formula with examples](https://bcgov.sharepoint.com/:w:/t/00608-ScrumTeam/EcXm1DnetBZNgFOkakDR_IEB1RS8r51PTHqH2h4C5L6URw?e=RbiNV2) - Please open it with the Word app instead of browser for formatting and better readability.

[Updated design](https://www.figma.com/file/PzOY8RyQnVXnjZJdSv2AqQ/CIF-Design?node-id=2898%3A2863&t=0b38GJFZh0GzOQrz-1) in Figma

Display calculated percentage, so that the user is alerted when it's below certain threshold.
- **Proponent's Share Percentage (calculated)** (Warning: it should be NO LESS than 25%. For 2021 projects, 10% instead.)

Display calculated amounts under the **Expenses and Payments Tracker** section.
- **Total Eligible Expenses To Date (calculated)** 
- **Total Gross Payment Amount To Date (calculated)** 
- **Total Holdback Payment Amount To Date (calculated)** 
- **Total Net Payment Amount To Date (calculated)** 

The calculated values for {total_eligible_expenses_to_date, gross_payment_amount_to_date, net_payment_amount_to_date} are taken care of via computed columns in #986.

#### Acceptance Criteria:

Given I am a CIF administrator
When I create or edit a project
And I click on the "Budgets, Expenses and Payments" section in the tasklist
Then I can see a tracker section
And its title is "Expenses and Payments Tracker"

Given I am a CIF administrator
When I am on the "Budgets, Expenses and Payments" form of a project
Then I can see the following amounts and percentage calculated:
- Total Eligible Expenses To Date (calculated)
- Total Gross Payment Amount To Date (calculated)
- Total Holdback Amount To Date (calculated)
- Total Net Payment Amount To Date (calculated)
- Proponent's Share Percentage (calculated) <- this can be done on the front end, it's just Proponent Cost / Total Project Value (implemented in #942)

#### Development Checklist:

- [x] add a new jsonSchemaForm to display the calculated values. This form can be stored on the front-end as it does not have data that needs to be persisted (does not need to go in the `schema/data/prod/jsonSchemaForm` directory.
- [x] display the calculated values at the bottom of the budgets, expenses and payments page as shown in figma
- [ ] Meets the DOD

#### Notes:

[...]

PR #1179, opened 2022-11-15
Chore: cumulative milestone calculations

dleard:
Adds computed columns for the following cumulative calculations:
- eligible_expenses
- gross_payments
- net_payments
- holdback_amount

~~This should unblock #984~~ 984 has been dropped. The functions have been reworked to be used in #717

### DIFF
--- a/schema/deploy/computed_columns/form_change_eligible_expenses_to_date.sql
+++ b/schema/deploy/computed_columns/form_change_eligible_expenses_to_date.sql
@@ -11,6 +11,7 @@ $fn$
     round(sum((fc.new_form_data->>'totalEligibleExpenses')::numeric), 2)
     from cif.form_change fc
     where fc.project_revision_id = $1.project_revision_id
+    and json_schema_name = 'milestone'
     and (fc.new_form_data->>'hasExpenses')::boolean = true;
 
 $fn$ language sql stable;
--- a/schema/deploy/computed_columns/form_change_eligible_expenses_to_date.sql
+++ b/schema/deploy/computed_columns/form_change_eligible_expenses_to_date.sql
@@ -1,0 +1,19 @@
+-- Deploy cif:computed_columns/form_change_eligible_expenses_to_date to pg
+-- requires: tables/form_change
+
+begin;
+
+create or replace function cif.form_change_eligible_expenses_to_date(parameter_fc cif.form_change)
+returns numeric as
+$fn$
+
+  select
+    round(sum((fc.new_form_data->>'totalEligibleExpenses')::numeric), 2)
+    from cif.form_change fc
+    where fc.project_revision_id = parameter_fc.project_revision_id
+    and (fc.new_form_data->>'hasExpenses')::boolean = true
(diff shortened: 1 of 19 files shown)


schema/deploy/computed_columns/form_change_eligible_expenses_to_date.sql
```diff
@@ -0,0 +1,21 @@
+-- Deploy cif:computed_columns/form_change_eligible_expenses_to_date to pg
+-- requires: tables/form_change
+
+begin;
+
+create or replace function cif.form_change_eligible_expenses_to_date(parameter_fc cif.form_change)
```
pbastia, line 6: this function will return the same value for all milestones, should we only sum the milestones that have been submitted before / including this milestone?

dleard: These functions are no longer calculating these values for individual milestones. They are for the budgets, payments and expenses page. So they're meant to be a total sum of all milestones.

schema/deploy/computed_columns/form_change_gross_payments_to_date.sql
```diff
@@ -0,0 +1,21 @@
...
+-- requires: tables/form_change
+
+begin;
+
+create or replace function cif.form_change_gross_payments_to_date(parameter_fc cif.form_change)
+returns numeric as
+$fn$
+
+  select
+    round(sum(coalesce((fc.new_form_data->>'adjustedGrossAmount')::numeric, (fc.new_form_data->>'calculatedGrossAmount')::numeric, (fc.new_form_data->>'maximumAmount')::numeric)), 2)
```
pbastia, line 11: I wonder if in the `payments_to_date` we should include the maximum amounts, which are payments that haven't been made yet

dleard: This is what I understood from talking with @nanyangpro but we can wait for confirmation before merging

dleard: You're right. We are not to include the maximum amounts here, only completed milestones (ones with adjusted/calculated values) are to be counted. Fixed.

schema/test/unit/computed_columns/form_change_eligible_expenses_to_date_test.sql
```diff
@@ -0,0 +1,79 @@
...
+  (
+    with record as (
+    select row(form_change.*)::cif.form_change
+    from cif.form_change where id=1
+    ) select cif.form_change_eligible_expenses_to_date((select * from record))
+  ),
+  (
+    60000::numeric
+  ),
+  'Returns the correct amount for the sum of all milestones with expenses'
```
pbastia, line 74: We might want to add one more milestone where `totalEligibleExpenses` is not set in the test suite, to make sure it doesn't break

pbastia (changes requested):
Just one minor test case! This is great 💯

pbastia (approved):

pbastia (approved):

Issue #1049, opened 2022-08-30
Check for SQL errors in revert and verify files [8]


BCerki:
###Description of the Tech Debt###

NOTE: If we decide to get rid of the `revert` and `verify` files, this issue is irrelevant.

We don't typically use/run the `revert` and `verify` files locally, so if there's an error in them, we may not know about it.

###Tech Debt Triage#
The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

Risk Value Scoring:

| Level  | Value |
| ------ | ----- |
| High   | 3     |
| Medium | 2     |
| Low    | 1     |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             |       |    2   |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     |       |   1    |
| System Fault Risk - Risk of system errors or application downtime                                                             |       |    2   |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? |       |   2    |
| Time Sink Risk - How much will this tech debt slow the development process down                                               |       |    1   |
| TOTAL SCORE:                                                                                                                  |       |   8    |

BCerki:
@dleard, does your PR https://github.com/bcgov/cas-cif/pull/1190 fix this issue?

dleard:
I think this card was about creating an automated check for revert/verify. I just manually checked for and fixed some errors that were introduced recently.

mikevespi:
Dylan: We use revert in dev, but don't allow reverts past tags in production.

A discussion should be had on whether sqitch verify is actually helpful and worth maintaining.

pbastia:
we've had a few instances where some code was hastily pushed to dev without being checked if it runs properly - having an automated action that reverts and verifies would help a little with this as an additional check

dleard:
revert is more important than verify. This card has implemented the revert check. We can talk about verify again if we decide it is a useful check.

PR #1190, opened 2022-11-18
Chore: fix sqitch revert

dleard:
- removes a call to non-existent function cif_private.revoke_permisisons()
- changes a reworked table to follow the `table_001` migration pattern. This revert error could have been fixed by simply fixing the reworked revert file, but I don't think this kind of change, while _technically_ idempotent, is what `sqitch rework` was meant for.


schema/revert/tables/form_001.sql
```diff
@@ -0,0 +1,9 @@
+-- Deploy cif:tables/form to pg
+-- requires: schemas/main
+
+begin;
+
+-- No revert. Cannot change varchar back to regprocedure without error.
+-- Table will be dropped in original revert file.
```
BCerki, line 7: I don't completely understand this comment. Sqitch will go all the way back to the original revert file since there isn't a 001 revert?

BCerki (commented):
I was able to deploy and then `sqitch revert` everything. I got a couple of warnings/info in the console, concerns?

`- tables/emission_intensity_report_001 ............................................ psql:revert/tables/emission_intensity_report_001.sql:13: NOTICE:  column "adjusted_holdback_payment_amount" of relation "emission_intensity_report" does not exist, skipping
psql:revert/tables/emission_intensity_report_001.sql:13: NOTICE:  column "date_sent_to_csnr" of relation "emission_intensity_report" does not exist, skipping
ok`
`- tables/project_001 .............................................................. psql:revert/tables/project_001.sql:10: NOTICE:  column "score" of relation "project" does not exist, skipping
psql:revert/tables/project_001.sql:10: NOTICE:  column "project_type" of relation "project" does not exist, skipping
ok`

schema/revert/tables/form_001.sql
pbastia: Sqitch runs all revert files in the reverse order of the plan, as sql scripts. If we don't have any code in there, no actions will be run, and sqitch will proceed to the next revert file. Eventually, the revert file for `form.sql` will be hit and the table dropped.

tmastrom (commented):
Sqitch verify is failing with 7 errors for me. This is after running `make drop_db && make deploy_dev_data`

dleard:
> Sqitch verify is failing with 7 errors for me. This is after running `make drop_db && make deploy_dev_data`

Thanks for checking this. We don't use or test the "final state" `sqitch verify` command anywhere currently so I'm not surprised it is broken. That failure can be handled in a separate PR or by #1049, whichever comes first. This PR didn't introduce the verify failure and is just a quick fix for `sqitch revert`, which was broken by recent commits.

pbastia (approved):

pbastia (approved):

PR #1197, opened 2022-11-22
Feat: add contract number to project form

mikevespi:
Adds contract_number column to the project table, and project form schemas.


pbastia (approved):
Looks great 🥇

PR #1198, opened 2022-11-22
Fix/missing projects in list 2

BCerki:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/1106


schema/deploy/computed_columns/project_manager_names_concatenated.sql
```diff
@@ -0,0 +1,31 @@
...
+create or replace function cif.project_manager_names_concatenated(cif.project)
+returns text
+as $$
+declare
+  index int;
+  names text := '';
+begin
+  for index in 1..4
+  loop
+    select concat(names, (select cif.cif_user_full_name(
```
dleard, line 16: I think using `string_agg` instead of concat here will simplify things.
https://www.postgresqltutorial.com/postgresql-aggregate-functions/postgresql-string_agg-function/

schema/deploy/computed_columns/project_manager_names_concatenated.sql
```diff
@@ -0,0 +1,31 @@
...
+begin;
+
+create or replace function cif.project_manager_names_concatenated(cif.project)
+returns text
+as $$
+declare
+  index int;
+  names text := '';
+begin
+  for index in 1..4
```
dleard, line 14: hardcoding this index here handcuffs us a bit if we add/remove manager labels in the future. See below comment on how we can filter the results more dynamically.

schema/deploy/computed_columns/project_manager_names_concatenated.sql
```diff
@@ -0,0 +1,31 @@
...
+  loop
+    select concat(names, (select cif.cif_user_full_name(
+            (select row(cif_user.*)::cif.cif_user from cif.cif_user
+            where cif.cif_user.id =
+                (select cif_user_id from cif.project_manager
+                where cif.project_manager.project_id=$1.id
+                and cif.project_manager.project_manager_label_id=index))))::text);
```
dleard, line 21: We only want primary managers here. I'd suggest filtering on the id's where the label name contains `primary`
```suggestion
                and cif.project_manager.project_manager_label_id in (select id from cif.project_manager_label where label ilike '%primary%')))))::text);
```

schema/deploy/computed_columns/project_manager_names_concatenated.sql
```diff
@@ -0,0 +1,31 @@
...
+  end loop;
+  return names;
+end;
+$$ language plpgsql;
```
dleard, line 25: now that we don't need variables or loops, we can change this function to a pure `sql` function (and add `stable` so it recognizes it as a computed column
```suggestion
$$ language sql stable;
```

schema/deploy/computed_columns/project_manager_names_concatenated.sql
```diff
@@ -0,0 +1,31 @@
...
+
+grant execute on function cif.cif_user_full_name to cif_internal, cif_admin;
```
dleard, line 27:
```suggestion
grant execute on function cif.project_primary_managers to cif_internal, cif_admin;
```

schema/deploy/computed_columns/project_manager_names_concatenated.sql
```diff
@@ -0,0 +1,31 @@
+-- Deploy cif:computed_columns/project_manager_names_concatenated to pg
+
+-- instead returns text, concatenated, it'll be an empty string if there are no managers. Then it wouldn't need to take an argument, filter for empty string and not null (check filters for null or empty options). includes insensitive empty string (potentially consider refactoring to use tvector later)
+
+begin;
+
+create or replace function cif.project_manager_names_concatenated(cif.project)
```
dleard, line 7: I think `cif.project_primary_managers` is a more explicit name for this function
```suggestion
create or replace function cif.project_primary_managers(cif.project)
```

schema/deploy/computed_columns/project_manager_names_concatenated.sql
```diff
@@ -0,0 +1,31 @@
...
+                where cif.project_manager.project_id=$1.id
+                and cif.project_manager.project_manager_label_id=index))))::text);
+  end loop;
+  return names;
+end;
+$$ language plpgsql;
+
+grant execute on function cif.cif_user_full_name to cif_internal, cif_admin;
+
+comment on function cif.project_manager_names_concatenated is 'Returns a concatenated list of the project managers (used for filtering the projects table)';
```
dleard, line 29:
```suggestion
comment on function cif.project_primary_managers is 'Returns a concatenated list of the project managers (used for filtering the projects table)';
```

dleard (changes requested):

sam-warren (dismissed):
Looks good - might want to wait for @Sepehr-Sobhani to review before go-ahead as I haven't spent as much time on the project though 😄

Sepehr-Sobhani:
Looks great to me, Thank you @BCerki!
I just noticed two issues:

1. when I remove a user from being a primary manager, I still get that project when I filter projects based on that user:

https://www.loom.com/share/19f8611a8a904e0cb7eee95cc28c0943

2. When I add a project  I can't see the project in the project list if it doesn't have a primary manager (it's either without a manager or with a secondary manager) -> not sure if this is related to your changes.

BCerki:
> Looks great to me, Thank you @BCerki! I just noticed two issues:
> 
> 1. when I remove a user from being a primary manager, I still get that project when I filter projects based on that user:
> 
> https://www.loom.com/share/19f8611a8a904e0cb7eee95cc28c0943
> 
> 2. When I add a project  I can't see the project in the project list if it doesn't have a primary manager (it's either without a manager or with a secondary manager) -> not sure if this is related to your changes.

Thanks for catching this, @Sepehr-Sobhani! I fixed the first issue, but I wasn't able to reproduce the second. Do you have a Loom?

Sepehr-Sobhani:
> > Looks great to me, Thank you @BCerki! I just noticed two issues:
> > 
> > 1. when I remove a user from being a primary manager, I still get that project when I filter projects based on that user:
> > 
> > https://www.loom.com/share/19f8611a8a904e0cb7eee95cc28c0943
> > 
> > 2. When I add a project  I can't see the project in the project list if it doesn't have a primary manager (it's either without a manager or with a secondary manager) -> not sure if this is related to your changes.
> 
> Thanks for catching this, @Sepehr-Sobhani! I fixed the first issue, but I wasn't able to reproduce the second. Do you have a Loom?

@BCerki Please dismiss the second issue :) I was on branch`missing_projects-in-list` instead of version 2

dleard (approved):
Nice work! I can see projects in the list without any managers assigned. Awesome that the plpgsql function was able to be converted to pure sql.

PR #1200, opened 2022-11-24
feat: adding calculated values

gurjmatharu:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/980


app/data/jsonSchemaForm/projectMilestoneUiSchema.ts
```diff
@@ -10,10 +10,16 @@ const projectMilestoneUiSchema = {
     "certifierProfessionalDesignation",
     "submittedDate",
     "totalEligibleExpenses",
+    // "totalEligibleExpensesToDate",
+    // "totalGrossPaymentAmountToDate",
+    // "totalNetPaymentAmountToDate",
```
Sepehr-Sobhani, line 15: Leftover?!

Sepehr-Sobhani (commented):
Looks good! Just a general question?! do we need to have calculated values as a read-only field and not an input?!

app/data/jsonSchemaForm/projectMilestoneUiSchema.ts
gurjmatharu: Nope they'll need to be added later.  Just added them for ordering and sanity.

app/pages/cif/project-revision/[projectRevision]/create.tsx
```diff
@@ -51,7 +54,9 @@ export function ProjectRevisionCreate({
 }: RelayProps<{}, createProjectRevisionQuery>) {
   const { session, projectRevision, allRevisionTypes, allAmendmentTypes } =
     usePreloadedQuery(pageQuery, preloadedQuery);
-  const taskList = <TaskList projectRevision={projectRevision} mode={"view"} />;
+  const existingRevision = projectRevision?.project?.pendingProjectRevision?.id;
+  const mode = "view";
+  const taskList = <TaskList projectRevision={projectRevision} mode={mode} />;
```
Comment (pbastia) [inline app/pages/cif/project-revision/[projectRevision]/create.tsx:59]:
```suggestion
  const taskList = <TaskList projectRevision={projectRevision} mode={"view"} />;
```

app/components/Form/ProjectMilestoneReportForm.tsx
```diff
@@ -233,6 +236,12 @@ const ProjectMilestoneReportForm: React.FC<Props> = (props) => {
         </Button>
 
         {sortedMilestoneReports.map((node, index) => {
+          schema.dependencies.hasExpenses.oneOf[1].properties.calculatedGrossAmount.default =
+            Number(node.calculatedGrossAmountThisMilestone);
+          schema.dependencies.hasExpenses.oneOf[1].properties.calculatedNetAmount.default =
+            Number(node.calculatedNetAmountThisMilestone);
+          schema.dependencies.hasExpenses.oneOf[1].properties.calculatedHoldbackAmount.default =
+            Number(node.calculatedHoldbackAmountThisMilestone);
```
pbastia, line 244: Instead of modifying the schema (the default value is meant to be static) we should set these values in the form data, if it's not already set

app/pages/cif/project-revision/[projectRevision]/create.tsx
```diff
@@ -89,21 +94,25 @@ export function ProjectRevisionCreate({
...
-            uiSchema={projectRevisionUISchema}
-            onSubmit={handleCreateRevision}
-            ObjectFieldTemplate={EmptyObjectFieldTemplate}
-          >
-            <Button
-              type="submit"
-              size="small"
-              disabled={isCreatingProjectRevision}
+          {existingRevision ? (
+            <h1>There is an existing revision</h1>
```
Comment (pbastia) [inline app/pages/cif/project-revision/[projectRevision]/create.tsx:98]: I thought we would redirect the user? we can leave this for now, but we need to confirm with @nanyangpro what the behaviour should be.

pbastia (changes requested):
Just a few changes!

app/pages/cif/project-revision/[projectRevision]/create.tsx
Comment (gurjmatharu) [reply app/pages/cif/project-revision/[projectRevision]/create.tsx:98]: yup the other card still will need to address this.

app/components/Form/ProjectMilestoneReportForm.tsx
```diff
@@ -231,8 +234,17 @@ const ProjectMilestoneReportForm: React.FC<Props> = (props) => {
         >
           <FontAwesomeIcon icon={faPlusCircle} /> Add another milestone report
         </Button>
-
         {sortedMilestoneReports.map((node, index) => {
+          const localNewFormData = JSON.parse(JSON.stringify(node.newFormData));
```
pbastia, line 238: We've used this pattern when importing schema objects from their global definition, to make sure we weren't modifying the singleton in-place. In this case, the `node.newFormData` object is local to this file / form and is safe to modify in-place, and there is no concept of 'local' vs 'global'

```suggestion
          const formData = node.newFormData;
```

pbastia (changes requested):

app/components/Form/ProjectMilestoneReportForm.tsx
gurjmatharu: We get `Cannot add property calculatedGrossAmountThisMilestone, object is not extensible` I think we need a deep copy here.

pbastia: I see. then object destructuring will do what you want:
```suggestion
                const formData = {...node.newFormData};
```

pbastia (approved):

PR #1202, opened 2022-11-24
Feat/calculated total project value

Sepehr-Sobhani:
[ZH Card](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/942)

### DIFF
--- a/app/components/Form/ProjectFundingAgreementForm.tsx
+++ b/app/components/Form/ProjectFundingAgreementForm.tsx
@@ -232,19 +232,24 @@ const ProjectFundingAgreementForm: React.FC<Props> = (props) => {
     return additionalFundingSourceForms;
   }, [filteredAdditionalFundingSourceFormChanges]);
 
+  const nextAdditionalFundingSourceIndex =
+    filteredAdditionalFundingSourceFormChanges.length
+      ? sortedAdditionalFundingSourceForms[
+          sortedAdditionalFundingSourceForms.length - 1
+        ].newFormData.sourceIndex + 1
+      : 1;
+
   const [
     discardAdditionalFundingSourceFormChange,
     isDiscardingAdditionalFundingSourceFormChange,
   ] = useDiscardAdditionalFundingSourceFormChange();
 
-  const deleteAdditionalFundingSource = (sourceIndex, formChangeId) => {
+  const deleteAdditionalFundingSource = (formChangeId) => {
     discardAdditionalFundingSourceFormChange({
       variables: {
         input: {
-          revisionId: projectRevision.rowId,
-          sourceIndex,
+          formChangeId,
         },
-        connections: [projectRevision.additionalFundingSourceFormChanges.__id],
       },
       onCompleted: () => {
         delete formRefs.current[formChangeId];
@@ -401,7 +406,7 @@ const ProjectFundingAgreementForm: React.FC<Props> = (props) => {
                       className="additionalFundingSourceForm"
                       idPrefix="additionalFundingSource"
                       validateOnMount={formChange.changeStatus === "staged"}
-                      ref={(el) => (formRefs.current[formChange.id] = el)}
+                      ref={(el) => (formRefs.current[formChange.rowId] = el)}
                       schema={createAdditionalFundingSourceSchema(
                         allAdditionalFundingSourceStatuses
                       )}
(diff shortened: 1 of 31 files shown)


schema/test/unit/computed_columns/project_revision_total_project_value_test.sql
```diff
@@ -0,0 +1,157 @@
...
+
+/** SETUP **/
+
+truncate cif.operator restart identity cascade;
+truncate cif.project restart identity cascade;
+insert into cif.operator (legal_name) values ('test operator');
+
+select cif.create_project();
+select cif.create_project();
+select cif.create_project();
```
tmastrom, line 13: Why do we call this 3 times?

tmastrom (commented):
Is the `discard_additional_funding_sources` mutation for if the user removes a funding source during a project revision?

schema/deploy/mutations/discard_additional_funding_source_form_change.sql
```diff
@@ -0,0 +1,36 @@
...
+begin;
+
+create or replace function cif.discard_additional_funding_source_form_change(revision_id int, source_index int)
+returns setof cif.form_change
+as $discard_additional_funding_source_form_change$
+declare
+form_change_record record;
+
+begin
+  for form_change_record in select * from cif.form_change
```
dleard, line 14: I'm confused by the loop here. Shouldn't we only ever have one record being deleted by this function (unique on revision_id and source_index)?

app/components/Form/ProjectFundingAgreementForm.tsx
```diff
@@ -424,11 +422,7 @@ const ProjectFundingAgreementForm: React.FC<Props> = (props) => {
                       size="small"
                       className="removeButton"
                       onClick={() =>
-                        deleteAdditionalFundingSource(
-                          formChange.id,
-                          formChange.rowId,
-                          formChange.operation
-                        )
+                        deleteAdditionalFundingSource(index + 1, formChange.id)
```
dleard, line 425: deleting things by this generated index is going to cause a bug.
Scenario:
create 3 sources:
source 1 generated index 1. db source_index: 1
source 2 generated index 2, db source_index: 2
source 3 generated index 3, db_source_index: 3

delete source 2
you now have 
source 1 generated index 1, db_source_index 1
source 3 generated index 2, db_source_index 3 <- The discard will no longer be able to delete this one as the front-end generated index and the index saved in the db no longer match

schema/deploy/mutations/discard_additional_funding_source_form_change.sql
dleard: Can't we just update/delete based on a form_change id since each additional_funding_source is (currently) its own form_change? That would solve the delete by index bug I described

dleard (changes requested):

Sepehr-Sobhani:
> Is the `discard_additional_funding_sources` mutation for if the user removes a funding source during a project revision?

Yes! we have `useDiscardFormChange` hook which serves a generic job for discarding form changes but it doesn't specifically return the `ProjectFundingAgreementForm_projectRevision` fragment in the response so when I discard an additional funding source, I don't get the updated total project value. That's why I added a separate discard mutation.

schema/test/unit/computed_columns/project_revision_total_project_value_test.sql
Sepehr-Sobhani: I wanted to create 3 separate projects and test different scenarios on those projects.
1- funding parameter without max funding amount and proponent cost
2- funding parameter with max funding amount, proponent cost and additional funding sources
3- funding parameter with max funding amount and proponent cost but no additional funding sources

schema/deploy/mutations/discard_additional_funding_source_form_change.sql
Sepehr-Sobhani: True 💯! Thanks for raising that Dylan.

Sepehr-Sobhani: @dleard Do we need to do the same for `discard_funding_parameter_form_change` ?!

pbastia (approved):
🥇

pbastia (approved):

schema/deploy/mutations/discard_additional_funding_source_form_change.sql
Sepehr-Sobhani: I would say so if @dleard agrees!?

pbastia (approved):

pbastia (approved):
The calculated project value is not on the summary page anymore, we might want to check in with @nanyangpro, I think it's still relevant information to display it there

dleard (approved):

PR #1203, opened 2022-11-24
Feat: user lookup on login

pbastia:
Updating the 'session_sub' to the KC gold when the user first logs in


schema/deploy/mutations/update_or_create_user_from_session.sql
```diff
@@ -0,0 +1,38 @@
...
+
+  end if;
+
+  select * from cif.cif_user where session_sub = jwt.sub into result;
+  return result;
+end;
+$function$ language plpgsql strict volatile;
+
+grant execute on function cif.update_or_create_user_from_session to cif_internal, cif_external, cif_admin;
+
```
dleard, line 37: This function could use a comment

schema/deploy/tables/cif_user_004_allow_sub_update_flag.sql
```diff
@@ -0,0 +1,20 @@
...
+
+create trigger cif_user_session_sub_immutable_with_flag
+    before update of session_sub on cif.cif_user
+    for each row
+    execute function cif_private.cif_user_session_sub_immutable_with_flag_set();
+
+-- Allowing all the existing users to update the sub once.
+update cif.cif_user set allow_sub_update = true;
+
+
```
dleard, line 18: new column `allow_sub_update` needs a comment

dleard (changes requested):
Looks good, just missing a couple comments & the `allow_sub_update` column has broken the e2e setup. I think adding `update cif.cif_user set allow_sub_update = true;` to the top of the `schema/data/dev/001_cif_user.sql` should fix the issue.

Let's make sure to do some QA on this in the test environment to make sure the user sub-mapping is working as expected.

dleard (approved):

Issue #1194, opened 2022-11-19
Spurious happo diff


dleard:
Emisisons intensity report form - with errors is flip-flopping the page diff on submit button being disabled / not disabled. We probably need to wait for a mutation hook to finish in the test

BCerki:
Lindsay, there's nothing to see in dev on this one

LindsayMacfarlane:
Awesome - thanks @BCerki ! We can probably move into Accepted by PO/Dev column

BCerki:
This is still a problem, e.g. [see ](https://happo.io/a/336/p/1010/compare/861249a2e0f3d5ea118b6e3b75fae9689102c4ec/fe550bdf473c6bcb1e1afc012f650db23ca421a3)

BCerki:
Commented out problem diffs in `/home/briannacerkiewicz/cas-cif/app/cypress/integration/cif/project-revision/index.spec.js`:
- `cy.happoAndAxe("Project budgets Form", "empty", "main");`
- `// cy.happoAndAxe(
    //   "Emissions intensity report Form",
    //   "with errors",
    //   ".error-detail"
    // );`

pbastia:
An avenue to explore: we use cypress 12, but the happo package wants cypress 10

BCerki:
If we decide to address this by removing screenshots, I'd suggest adding some cypress assertions for calculated values. Sometimes happo is the only thing that catches changes to the calculations that use values from multiple forms

pbastia:
idea: remove iOS browsers from happo, they generate spurious diffs

PR #1214, opened 2022-11-29
chore: fix spurious happo diffs

BCerki:
Card: https://github.com/bcgov/cas-cif/issues/1194

### DIFF
--- a/app/cypress/integration/cif/project-revision/create-project-revision.spec.js
+++ b/app/cypress/integration/cif/project-revision/create-project-revision.spec.js
@@ -111,18 +111,8 @@ describe("when creating a project, the project page", () => {
 
     //add quarterly reports
     cy.url().should("include", "/form/6");
-    cy.addQuarterlyReport(
-      1,
-      "2020-01-01",
-      "2020-02-02",
-      "I am the first general comment"
-    );
-    cy.addQuarterlyReport(
-      2,
-      "2022-01-01",
-      "2022-02-02",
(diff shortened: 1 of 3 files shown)


app/cypress/integration/cif/project-revision/create-project-revision.spec.js
```diff
@@ -111,18 +111,8 @@ describe("when creating a project, the project page", () => {
 
     //add quarterly reports
     cy.url().should("include", "/form/6");
-    cy.addQuarterlyReport(
-      1,
-      "2020-01-01",
-      "2020-02-02",
-      "I am the first general comment"
```
BCerki, line 118: Cypress runs faster when there's less text to type

app/cypress/integration/cif/project-revision/index.spec.js
```diff
@@ -251,8 +251,12 @@ describe("the new project page", () => {
     cy.contains("Changes saved").should("be.visible");
 
     cy.findByRole("button", { name: /^submit/i }).click();
-    cy.wait("@gqlstageEmissionIntensityFormChangeMutation");
-    cy.wait("@gqlstageReportingRequirementFormChangeMutation");
+    cy.wait("@gqlstageEmissionIntensityFormChangeMutation")
+      .its("response")
+      .should("have.property", "body");
```
BCerki, line 256: Adding these to try to slow down the test

PR #1216, opened 2022-11-29
Chore/amendment revision status 2

BCerki:


schema/revert/mutations/create_project_revision.sql
```diff
@@ -1,18 +1,25 @@
...
-  insert into cif.project_revision(project_id, change_status)
-  values ($1, 'pending') returning * into revision_row;
+  insert into cif.project_revision(project_id, change_status, revision_type)
+  values ($1, 'pending', $2) returning * into revision_row;
+
+  foreach _amendment_type in array $3
+    loop
+      insert into cif.project_revision_amendment_type(project_revision_id, amendment_type)
+      values (revision_row.id, (select name from cif.amendment_type where cif.amendment_type.name = _amendment_type));
+    end loop;
```
pbastia, line 22: This could be refactored to a single query (for loops are frowned upon in SQL, the query engine is much better and efficient at retrieving and building sets). 

Something like this:

```suggestion
      insert into cif.project_revision_amendment_type(project_revision_id, amendment_type)
      values (
          select 
              revision_row.id as project_revision_id, 
              name as amendment_type 
           from cif.amendment_type 
           where cif.amendment_type.name in $3
       );
```

We could put this as a tech debt card to refactor later if it's too time consuming, since we only have 3 amendment types for now

pbastia (approved):
I'm approving this because the refactor is not critical, but would be good practice. We can do that in a separate tech debt card

pbastia (approved):
👌

PR #1233, opened 2022-12-02
Feat/set status for amendments and revisions

Sepehr-Sobhani:
[ZH Card](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/981)


app/cypress/integration/cif/project-revision-change-logs.spec.js
```diff
@@ -41,5 +41,7 @@ describe("the project amendment and revisions page", () => {
     cy.happoAndAxe("Amendment & Revision View", "view", "main", true);
     cy.findByRole("link", { name: /view minor revision 1/i }).should("exist");
     cy.get('input[value="Minor Revision"]').should("be.checked");
+    cy.get('[aria-label="Status"]').contains("Applied");
```
tmastrom, line 44: The card calls this "Approved" but I believe it was decided that this would be renamed to "Applied" throughout?

Sepehr-Sobhani:
@pbastia based on your comment here: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/1166
Do we need to commit the revision once the status has changed to `Applied/Approved` ?! if yes, do I need to add it to the AC of this card or make a new card for that?!

app/pages/cif/project-revision/[projectRevision]/view.tsx
```diff
@@ -75,17 +98,97 @@ export const ViewProjectRevisionQuery = graphql`
...
+          name
+          isAmendmentSpecific
+        }
+      }
+    }
   }
 `;
 
+// Custom widget to update the revision status
+export const RevisionStatusWidget = (props: WidgetProps) => {
```
Comment (pbastia) [inline app/pages/cif/project-revision/[projectRevision]/view.tsx:113]: This widget should be in its own file

app/pages/cif/project-revision/[projectRevision]/view.tsx
```diff
@@ -97,6 +200,15 @@ export function ProjectRevisionView({
...
   // Growthbook - amendments
   if (!useShowGrowthbookFeature("amendments")) return null;
+
+  // filtering to show only the amendment statuses that are allowed to be selected based on the revision type
+  const filteredRevisionStatuses = allRevisionStatuses.edges.filter(
+    ({ node }) =>
+      projectRevision.revisionType === "Amendment"
+        ? true
+        : !node.isAmendmentSpecific
+  );
```
Comment (pbastia) [inline app/pages/cif/project-revision/[projectRevision]/view.tsx:210]: no need to iterate over the statuses if the revision type is "Amendment", something like that might work better:

```suggestion
  const filteredRevisionStatuses = 
    projectRevision.revisionType === "Amendment" ? 
        allRevisionStatuses.edges :
        allRevisionStatuses.edges.filter(({ node }) => !node.isAmendmentSpecific );
```

app/pages/cif/project-revision/[projectRevision]/view.tsx
```diff
@@ -15,8 +15,17 @@ import readOnlyTheme from "lib/theme/ReadOnlyTheme";
...
+import { Button } from "@button-inc/bcgov-theme";
+import ReadOnlyWidget from "lib/theme/widgets/ReadOnlyWidget";
+import { WidgetProps } from "@rjsf/core";
+import SelectWidget from "lib/theme/widgets/SelectWidget";
+import { useState } from "react";
 
-const createProjectRevisionViewSchema = (allRevisionTypes) => {
+const createProjectRevisionViewSchema = (
+  allRevisionTypes,
+  allRevisionStatuses
```
Comment (pbastia) [inline app/pages/cif/project-revision/[projectRevision]/view.tsx:27]: the two parameters are inconsistent (we should type them to avoid it and increase readability)
- `allRevisionTypes` is an object `{ edges: { node: ... }[]}`
- `allRevisionStatuses` is an array of edges

I suggest making them either both an array of edges, and calling them `allRevisionTypesEdges` and `allRevisionStatusesEdges`

app/pages/cif/project-revision/[projectRevision]/view.tsx
```diff
@@ -111,7 +223,10 @@ export function ProjectRevisionView({
             tagName={"dl"}
             className="project-revision-view-form"
             schema={
-              createProjectRevisionViewSchema(allRevisionTypes) as JSONSchema7
+              createProjectRevisionViewSchema(
+                allRevisionTypes,
+                filteredRevisionStatuses
+              ) as JSONSchema7
```
Comment (pbastia) [inline app/pages/cif/project-revision/[projectRevision]/view.tsx:229]: casting here could be avoided if we declare the return type of `createProjectRevisionViewSchema(..)`

pbastia (changes requested):
Good stuff!!

app/pages/cif/project-revision/[projectRevision]/view.tsx
```diff
@@ -15,12 +15,27 @@ import readOnlyTheme from "lib/theme/ReadOnlyTheme";
...
+        readonly type: string;
+      };
+    }[];
+  },
+  allRevisionStatusesEdges: readonly {
+    readonly node: {
+      readonly name: string;
+      readonly isAmendmentSpecific: boolean;
+    };
+  }[]
```
Comment (pbastia) [inline app/pages/cif/project-revision/[projectRevision]/view.tsx:33]: - the `readonly` is unnecessary here
- we can remove the edges key to make both arguments consistent in type

```suggestion
const createProjectRevisionViewSchema = (
  allRevisionTypesEdges: {
     node: {
       type: string;
      };
  }[],
  allRevisionStatusesEdges: {
   node: {
     name: string;
     isAmendmentSpecific: boolean;
    };
  }[]
```

pbastia (changes requested):
I probably just confused you a little with my comment earlier :p

pbastia (approved):
🥇

PR #1252, opened 2022-12-09
Fix: sqitch migrations

dleard:
@pbastia This should fix our broken sqitch plan

### DIFF
--- a/schema/test/unit/tables/project_revision_003_set_amendment_statuses_test.sql
+++ b/schema/test/unit/tables/project_revision_003_set_amendment_statuses_test.sql
@@ -53,7 +53,7 @@ select has_column(
   'cif',
   'project_revision',
   'revision_status',
-  'project_revision table now has an amendment_status column'
+  'project_revision table now has a revision_status column'
 );
 
 select is (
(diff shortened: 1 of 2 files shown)


schema/test/unit/tables/project_revision_003_set_amendment_statuses_test.sql
```diff
@@ -0,0 +1,87 @@
...
+
+select is (
+  (
+    select revision_status from cif.project_revision where id = 2
+  ),
+  'Draft',
+  'A revision in committed change_status is correctly set to a revision_status of Applied'
+);
+
+select * from cif.project_revision;
```
Sepehr-Sobhani, line 83: Leftover!

schema/test/unit/tables/project_revision_003_set_amendment_statuses_test.sql
```diff
@@ -0,0 +1,87 @@
...
+  'Draft',
+  'A revision in staged change_status is correctly set to a revision_status of Draft'
+);
+
+select is (
+  (
+    select revision_status from cif.project_revision where id = 2
+  ),
+  'Draft',
+  'A revision in committed change_status is correctly set to a revision_status of Applied'
```
Sepehr-Sobhani, line 80: I think either the id should be 3 or the result of the test should be `Applied`

dleard:
closing in favour of a new PR which splits the postUprade / preUpgrade jobs into two separate jobs instead of one with conditionals to avoid confusion/errors if or when we need to add new post/pre upgrade functionality.

PR #1256, opened 2022-12-09
fix: re-order sqitch plan - debug gh actions

dleard:


dleard:
debug complete, main and gh actions are not properly set up for pull requests. closing debug branch

PR #1276, opened 2022-12-14
chore: persist data in test until after migrations are complete

dleard:


chart/cas-cif/values-test.yaml
```diff
@@ -5,12 +5,18 @@ app:
 hostName: test.cif.gov.bc.ca
 
 db:
-  preUpgradeCommand: |
+  postUpgrade: true
+  upgradeCommand: |
```
pbastia, line 9: This api doesn't allow us to do both pre- and post- upgrade, I think we should keep them separate just in case, to avoid confusion

dleard:
closing in favour of a new PR which splits the postUprade / preUpgrade jobs into two separate jobs instead of one with conditionals to avoid confusion/errors if or when we need to add new post/pre upgrade functionality.

Issue #1075, opened 2022-09-09
As a project coordinator, I want to select a funding stream and RFP as the first step of creating a project


matthieu-foucault:
#### Description:

[- Envisioned workflow for EP](https://www.figma.com/proto/PzOY8RyQnVXnjZJdSv2AqQ/CIF-Design?page-id=5936%3A294&node-id=5936%3A1172&viewport=-589%2C1434%2C0.25&scaling=scale-down&starting-point-node-id=5936%3A1172&show-proto-sidebar=1) (validated by CIF)
[- Envisioned workflow for IA](https://www.figma.com/proto/PzOY8RyQnVXnjZJdSv2AqQ/CIF-Design?page-id=5936%3A294&node-id=6008%3A2837&viewport=-589%2C1434%2C0.25&scaling=scale-down&starting-point-node-id=6008%3A3069&show-proto-sidebar=1) (validated by CIF)

This would prevent altering it later and having form records that are not relevant. 

#### Acceptance Criteria:

Given I am a CIF user
When I click 'create project'
Then I am prompted to enter the `stream` and `RFP year` before continuing to the rest of the forms

#### Development Checklist:

- [ ] The 'create project' button creates a project on click. This is already how it works, but it will still need to create the project before we can populate the values.
- [ ] User is prompted to enter `stream` and `RFP year` before continuing to the rest of the forms
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [x] User story is included
- [x] User role and type are identified
- [x] Acceptance criteria are included
- [x] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [x] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [x] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [x] Reviewed and approved by Product Owner

#### Notes:

-

LindsayMacfarlane:
[...]

PR #1288, opened 2022-12-16
Feat/3 year stream first

BCerki:
zenhub Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1075
github card: https://github.com/bcgov/cas-cif/issues/1075

Immutability to be completed in: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1100

Notes:
- I updated the `create_project` function so every test file that used it had to be updated (hence why there are 38 files changed in this PR)
- I don't think we need `cif.form_change_is_pristine` anymore. As far as I can tell, it was used to make the tasklist show "In Progress" instead of "Attention Required" when the user first arrives at the project overview form. (The form is only pristine when `newFormData` is null vs. {}). Overview form status is now handled by the routing function because initial `newFormData` now includes rpf info. I've made a tech debt issue to investigate if we still need `cif.form_change_is_pristine` . https://github.com/bcgov/cas-cif/issues/1299
- This PR revealed a happo bug/tech debt, recorded here: https://github.com/bcgov/cas-cif/issues/1298


app/pages/cif/project-revision/new.tsx
```diff
@@ -0,0 +1,127 @@
...
+  }
+`;
+
+export function ProjectRevisionNew({
+  preloadedQuery,
+}: RelayProps<{}, newProjectRevisionQuery>) {
+  const { query } = usePreloadedQuery(pageQuery, preloadedQuery);
+
+  const router = useRouter();
+  const [formState, setFormState] = useState({});
```
BCerki, line 56: If you don't pass `formData` into `FormBase`, [the form clears whenever the state changes](https://github.com/rjsf-team/react-jsonschema-form/issues/584) (in this case, most relevantly when you click `Submit`). We can't grab `formData` from a graphql query like we do in the other forms because the project doesn't get a `projectId` until it's created (you click `Confirm`).

app/components/TaskList/index.tsx
```diff
@@ -92,7 +92,7 @@ const TaskList: React.FC<Props> = ({
 
   const [stageDirtyFormChanges] = useStageDirtyFormChanges();
   useEffect(() => {
-    if (mode !== "view")
+    if (mode !== "view" && !router.query?.isRoutedFromNew)
```
BCerki, line 95: This is the best way I could think of to control the tasklist status. Previously, it was controlled by checking if the form was pristine (`newFormData` was null). Now that we're creating the project with rfp info (`newFormData: {fundingStreamRfpId: #}`, there's no way to tell if `newFormData` contains only rfp data because the project has just been created (in progress status), or because the user forgot to fill out other fields (attention required status).

app/cypress/integration/cif/project-revision/index.spec.js
```diff
@@ -191,10 +195,18 @@ describe("the new project page", () => {
...
+    cy.url().should("include", "/new");
+    cy.findByRole("button", { name: /^confirm/i }).click();
+    cy.happoAndAxe("Project New Form", "with errors", ".error-detail", true);
+    cy.fillAndCheckNewProjectForm("Emissions Performance", "2020");
+    cy.findByRole("button", { name: /^confirm/i }).click();
 
     // OVERVIEW
-    cy.get("button").contains("Add a Project").click();
     cy.url().should("include", "/form/0");
+    cy.findByText("Emissions Performance").should("be.visible");
```
BCerki, line 209: In the happo screenshots, the RFP info isn't filled out on the overview page. It works locally so I think it's a race conditions thing. I'm forcing it to check that the info is visible but it still doesn't doesn't show it in the screenshots. Thoughts?

tmastrom: It doesn't seem like this is a unique problem to this PR. The following test passes and cypress is able to find all options in the dropdown selector by text. 

```  
it.only("selects funding stream and year", () => {
    cy.mockLogin("cif_admin");
    cy.visit("/cif/projects");
    cy.get("button").contains("Add a Project").click();
    cy.url().should("include", "/new");
    cy.happoAndAxe("Project New Form", "unfilled", "main");
    cy.fillAndCheckNewProjectForm("Emissions Performance", "2020");
    cy.findByText("Emissions Performance").should("be.visible");
    cy.findByText("Select a Funding Stream").should("be.visible");
    cy.findByText("Innovation Accelerator").should("be.visible");
  });
```

I was also not seeing the correct dropdown selected in the DOM snapshots when running these tests locally

app/components/TaskList/index.tsx
tmastrom: This seems to be a valid use of next router based on their docs. I wonder if there's an argument for there to be a new field to hold the `newFormStatus` to distinguish between `in progress status` and `just created`. This also could be a discussion in `1299 do we need form change is pristine`

app/cypress/integration/cif/project-revision/index.spec.js
BCerki: Know cypress issue, made a ticket to check if they've fixed it in the future: https://github.com/bcgov/cas-cif/issues/1298

app/pages/cif/projects.tsx
```diff
@@ -152,8 +138,7 @@ export function Projects({ preloadedQuery }: RelayProps<{}, projectsQuery>) {
   ) : (
     <Button
       role="button"
-      onClick={handleCreateProject}
-      disabled={isCreatingProject}
+      onClick={() => router.push(getNewProjectRevisionPageRoute())}
```
mikevespi, line 141: We need to add this logic to the new project button on the home page too, right now it's erroring out for me.

app/cypress/support/commands.js
```diff
@@ -80,11 +80,23 @@ Cypress.Commands.add("clearMockedTime", () => {
   cy.clearCookie("mocks.mocked_timestamp");
 });
 
+Cypress.Commands.add(
+  "fillAndCheckNewProjectForm",
```
mikevespi, line 84: Consider adding params for which funding stream and year to fill and check, just be a little more flexible for future use.

mikevespi (changes requested):
I left a couple of comments but it's looking great! Can we also confirm with UX whether we want to let users change the values of the stream / year on the project form page? Letting them change it seems to defeat the purpose of forcing them to enter this before creating the project. All we'll need to do is change them to the `ReadOnlyWidget` if so.

Great work!

app/pages/cif/project-revision/new.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { newProjectRevisionQuery } from "__generated__/newProjectRevisionQuery.graphql";
+
+const schema: JSONSchema7 = {
+  $schema: projectSchema.$schema,
+  type: projectSchema.type as JSONSchema7TypeName,
+  required: ["fundingStreamRfpId"],
+  properties: {
+    fundingStreamRfpId: {
+      type: "number",
+      title: "RFP Year ID",
```
pbastia, line 24: `title` is unnecessary here, it renders an additional label on the screen

pbastia (changes requested):
A couple more minor things. Great work!

BCerki:
> I left a couple of comments but it's looking great! Can we also confirm with UX whether we want to let users change the values of the stream / year on the project form page? Letting them change it seems to defeat the purpose of forcing them to enter this before creating the project. All we'll need to do is change them to the `ReadOnlyWidget` if so.
> 
> Great work!

Ultimately users won't be able to change stream/year; immutability is a separate ticket: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1100

app/cypress/support/commands.js
BCerki: I'm not sure what you mean--different params from what's on line 85 below?

mikevespi:
> > I left a couple of comments but it's looking great! Can we also confirm with UX whether we want to let users change the values of the stream / year on the project form page? Letting them change it seems to defeat the purpose of forcing them to enter this before creating the project. All we'll need to do is change them to the `ReadOnlyWidget` if so.
> > Great work!
> 
> Ultimately users won't be able to change stream/year; immutability is a separate ticket: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1100

Got it!

app/cypress/support/commands.js
mikevespi: Nope... Sorry I was still shaking off vacation brain

mikevespi (approved):
Awesome!

pbastia (approved):
👍 👍

PR #1295, opened 2022-12-28
Use relay testing helper package

tmastrom:


pbastia:
@dleard @tmastrom before we merge this, we'll wait until we have clarity on maintenance plans for this library, and a proper release

Issue #944, opened 2022-08-08
As a CIF administrator, I want to see anticipated funding amounts per year for a project, so that I can budget the year accordingly


pbastia:
#### Description:

Each project should display its anticipated funding amount per year.

[Formula with examples](https://bcgov.sharepoint.com/:w:/t/00608-ScrumTeam/EcXm1DnetBZNgFOkakDR_IEB1RS8r51PTHqH2h4C5L6URw?e=RbiNV2) - Please open it with the Word app instead of browser for formatting and better readability.

[Updated design](https://www.figma.com/file/PzOY8RyQnVXnjZJdSv2AqQ/CIF-Design?node-id=2898%3A2863&t=0b38GJFZh0GzOQrz-1) in Figma

#### Acceptance Criteria:

Given I am a CIF administrator
When I create or edit a project
And I am on the "Budgets, expenses and payments" form
Then I can see, for each *Fiscal* year funded, a calculated "Anticipated Funding amount (fiscal year x)"

Given I am a CIF administrator
When I view the anticipated funding amount for a year
Then the value is calculated from either the amounts paid for each past milestone (including holdback), or the maximum funding amounts for the upcoming milestones

#### Development Checklist:

- [x] New entries in our `json_schema_form` 
- [x] Computed column that returns an array of `{year: 20YZ, anticipatedFundingAmount: 1234.565 }` object. This will need to coalesce the adjusted gross amount and the calculated gross amount (similar to the `cif.form_change_calculated_net_amount_this_milestone` implementation)
- [x] Matches the government way of displaying fiscal years  `First Year / Second Year` (official spec [here](https://www2.gov.bc.ca/assets/gov/government/services-for-government-and-broader-public-sector/information-technology-services/standards-files/date_and_time_standard.pdf))
- [x] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [x] User story is included
- [x] User role and type are identified
- [x] Acceptance criteria are included
- [x] Wireframes are included (if required)
- [x] Design / Solution is accepted by Product Owner
- [x] Dependencies are identified (technical, business, regulatory/policy)
- [x] Story has been estimated (under 13 pts)

**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
[...]

PR #1302, opened 2023-01-03
Anticipated Funding Amounts

BCerki:
Card: https://github.com/bcgov/cas-cif/issues/944


BCerki:
Continued in another pr

Issue #1167, opened 2022-11-08
As a CIF administrator, I want to notify the individual(s) an amendment is waiting for, so I can get their attention


pbastia:
#### Description:

This is part 2 of #982

[Prototype Example 1 (Notify Tech Team)](https://www.figma.com/proto/PzOY8RyQnVXnjZJdSv2AqQ/CIF-Design?page-id=3989%3A5985&node-id=3989%3A8493&viewport=-674%2C-451%2C0.24&scaling=scale-down&starting-point-node-id=3989%3A8493&show-proto-sidebar=1)
[Prototype Example 2 (Notify Director)](https://www.figma.com/proto/PzOY8RyQnVXnjZJdSv2AqQ/CIF-Design?page-id=3989%3A5985&node-id=3989%3A7927&viewport=-674%2C-451%2C0.24&scaling=scale-down&starting-point-node-id=3989%3A7927&show-proto-sidebar=1) 
[Prototype Example 3 (Notify Proponent)](https://www.figma.com/proto/PzOY8RyQnVXnjZJdSv2AqQ/CIF-Design?page-id=3989%3A5985&node-id=3989%3A7085&viewport=-674%2C-451%2C0.24&scaling=scale-down&starting-point-node-id=3989%3A7085&show-proto-sidebar=1)

When the user clicks the "Notify" button on an amendment view in the "pending actions from" control, a modal opens up, for the user to select which individuals to notify

#### Acceptance Criteria:

Given I am a CIF administrator
When I view an amendment from the revisions list
And I click on the "Notify" button from the "pending actions from" section
Then I can see a modal dialog allowing me to select who to notify, with a "Notify By Email" and "Cancel" buttons

Given I am a CIF administrator
And I view the pending actions modal
When I click the "Cancel" button
Then the modal closes and nothing else happens

Given I am a CIF administrator
When I am notifying people that an amendment requires their attention
And I am viewing the modal 
Then for each selected group, I can select the following contacts:
- Ops Team : Ops Team Primary and Secondary managers
- Tech Team: Tech Team Primary and Secondary managers
- Director: Program director
- Proponent: Project contacts

Given I am a CIF administrator
When I am notifying people that an amendment requires their attention
And I am viewing the modal 
And a Project Manager (Primary or secondary) or a Proponent contact is not defined
Then I can see a message telling me where to add it
And the associated checkbox is disabled

Given I am a CIF administrator
And I am viewing the notification modal
And I selected at least one 
[...]

PR #1303, opened 2023-01-03
Feat/amendment notification ENV

BCerki:
ENV variables to support card: https://github.com/bcgov/cas-cif/issues/1167

### DIFF
--- a/app/.env.example
+++ b/app/.env.example
@@ -19,5 +19,5 @@ ENABLE_MOCK_TIME=<true or false>
 ENABLE_MOCK_AUTH=<true or false>
 
 # These variables are for setting the CIF program director name and email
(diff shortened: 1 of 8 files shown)


app/.env.example
```diff
@@ -19,5 +19,5 @@ ENABLE_MOCK_TIME=<true or false>
 ENABLE_MOCK_AUTH=<true or false>
 
 # These variables are for setting the CIF program director name and email
-NEXT_PUBLIC_PROGRAM_DIRECTOR_NAME=
-NEXT_PUBLIC_PROGRAM_DIRECTOR_EMAIL=
```
BCerki, line 23: They needed prefixes when being used via `process.env`, but since it's convict now, I shortened the names for clarity

Sepehr-Sobhani (approved):
Thanks Brianna!

PR #1304, opened 2023-01-04
Chore: prod data as migration

dleard:


schema/deploy/migrations/insert_json_schema_form_data.sql
```diff
@@ -1,34 +1,40 @@
+-- Deploy cif:migrations/insert_json_schema_form_data to pg
```
pbastia, line 1: Should we put all data files in a `schema/deploy/data/<...>_data.sql` ? to separate it from migrations on existing user data in the database

schema/verify/migrations/insert_json_schema_form_data.sql
```diff
@@ -0,0 +1,13 @@
+-- Verify cif:migrations/insert_json_schema_form_data on pg
+
+begin;
+
+do $$
+  begin
+    assert (
+      (select count(*) from cif.form) > 0
```
pbastia, line 8: Should we count the exact number of rows we want to insert?

pbastia (approved):
a couple of comments that are more just code opinions 😛 nice removing of code

schema/deploy/migrations/insert_json_schema_form_data.sql
dleard: I _think_ putting something under deploy that isn't in the sqitch plan will make sqitch angry?

schema/verify/migrations/insert_json_schema_form_data.sql
dleard: We could, but this will cause `sqitch verify` to fail in the future if we add/remove rows

schema/deploy/migrations/insert_json_schema_form_data.sql
pbastia: I didn't express myself well, I just meant updating the path for this file to be in `schema/deploy/data/...` instead of `schema/deploy/migrations/...`

schema/verify/migrations/insert_json_schema_form_data.sql
pbastia: That would be to make sure we insert the proper number of rows and force us to update the verify script if we rework the migration

pbastia (approved):
🥇

Issue #1037, opened 2022-08-30
Milestones: "Professional Designation" is optional but doesn't have a blank option [7]


Sepehr-Sobhani:
Describe the Bug:

"Professional Designation" is optional, but defaults to "Professional Engineer" and there's no way to unset this (no blank option)

Probability (how likely the bug is to happen, scored from 1-5): 5
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 2
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

Create a new milestone report

Screenshots:
https://www.loom.com/share/6a21a94870cf4d3bb00159212a4f09e1

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

gurjmatharu:
@nanyangpro @suhafa for this issue what is the expected behaviour.  Should an option be `None` or `N/A`?  Thoughts?

nanyangpro:
Thanks for your question, @gurjmatharu!

In terms of design consistency, we can replicate the exact behaviour of the optional fields on the "Project Managers" form. However, I'm lacking context of this field, @Suha's input would be more appreciated.

suhafa:
Thanks @gurjmatharu and @nanyangpro! We could go with your suggestion of `N/A` here. This field is almost always either the `Professional Engineer` or `CPA`. `N/A` makes more sense in case there is none assigned (yet).

LindsayMacfarlane:
Thanks @gurjmatharu! 

A quick observation:
- The review form show's an error if submitting a project where the professional designation is NA. 
![image.png](https://images.zenhubusercontent.com/60f9db089e3a388facde96d1/9d431e5b-bf45-4c01-8fee-5c07ab80c6a8)  
I confirmed that this error is not displayed when the other two professional designations are selected. 

Moving to approved, but if you could create a bug ticket that would be fantastic.

LindsayMacfarlane:
Note: bug card #1352 has been created.

PR #1305, opened 2023-01-04
fix: adding N/A as an option to Professional Designation

gurjmatharu:
bugfix: #1037


schema/data/prod/json_schema/milestone.json
```diff
@@ -51,6 +51,11 @@
             "type": "string",
             "title": "Certified Professional Accountant",
             "enum": ["Certified Professional Accountant"]
+          },
+          {
+            "type": "string",
+            "title": "N/A",
```
BCerki, line 57: Maybe it should say "Select a Designation" instead of "N/A"? I think all the other dropdowns say "Select a XX" if blank

schema/data/prod/json_schema/milestone.json
```diff
@@ -51,6 +51,11 @@
             "type": "string",
```
BCerki, line 51: Sorry, can't comment on lines that weren't changed, but on 43, now that we have a blank option, should we be defaulting to that instead?

BCerki (commented):
⛵

schema/data/prod/json_schema/milestone.json
tmastrom: We want to allow users to select a blank option if they don't have a professional designation. An option could simply be "None" or "No Professional Designation"

BCerki (approved):

PR #1307, opened 2023-01-04
Chore: test sqitch revert

dleard:
Fixing broken sqitch revert and adding a CI test to ensure revert does not fail after changes.

I've removed two sets of leftover files:
commit_project_revision (version 1.0.0-rc.5) deploy/revert/verify
create_project (version 1.0.0-rc.5) deploy/revert/verify

It appears these versions were created [here](https://github.com/bcgov/cas-cif/commit/fdc75ecdb24cc1201e0247653d7da5e17637660f#diff-ae3b5ff04fcdd319c0ba8eb130d1bd3daa5975c74388d91f623f4f44cb0723d6) and then subsequently removed from the sqitch plan [here](https://github.com/bcgov/cas-cif/commit/6f212d62164fd56ba58041d2cf3ff2f41e3f228b#diff-ae3b5ff04fcdd319c0ba8eb130d1bd3daa5975c74388d91f623f4f44cb0723d6) in favor of a set of files tagged with version 1.0.1

I think this was likely from branching off of a previous branch that was started before the 1.0.1 release (rather than off of develop) and continuing work on the new branch after the 1.0.1 release.

The rc.5 tagged files were not removed when replaced by 1.0.1 tagged files, so they persisted with the actual function drop in rc.5, but were unreachable as they did not exist in the sqitch plan. So these functions never get dropped during revert.


mikevespi (approved):
Nice! Tested it locally too, everything looks good!

PR #1310, opened 2023-01-05
Anticipated Funding Amounts 

BCerki:
card: https://github.com/bcgov/cas-cif/issues/944. See the card's comments for some notes on the design/calculations


app/lib/theme/AnticipatedFundingPerFiscalYearArrayFieldTemplate.tsx
```diff
@@ -0,0 +1,64 @@
...
+    return <div>No array, ask the designers what they want here</div>;
+  }
+
+  return (
+    <div>
+      {formContext.anticipatedFundingArray.edges.map(({ node }) => {
+        console.log("node", node);
+        return (
+          <>
+            {node && node.fiscalYear && (
```
BCerki, line 20: If there's no fiscal year (the user hasn't filled out any of date sent to csnr, submitted date, or due date), then the amount won't appear in any of the anticipated funding fields. I don't think it can work any other way, because without having a date we don't know where to add it, but it could result in a missing milestone armount.

app/data/jsonSchemaForm/fundingAgreementSchema.ts
```diff
@@ -79,6 +102,20 @@ export const fundingAgreementUiSchema = {
...
+        calculatedValueFormContextProperty: "calculatedAnticipatedAmount",
+      },
+    },
+    "ui:options": {
+      addable: false,
+      orderable: false,
+      removable: false,
+      label: false,
+    },
+  },
```
BCerki, line 118: I don't think this is needed if we're using the `ArrayFieldTemplate` but I left it in for now in case this is a better approach.

app/lib/theme/AnticipatedFundingPerFiscalYearArrayFieldTemplate.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  if (areNodesNull) {
+    const fiscalYears = [1, 2, 3];
+    return fiscalYears.map((i) => (
+      <>
+        <div key={i}>
+          <label
+            htmlFor={`ProjectFundingAgreementForm_anticipatedFundingAmountByFiscalYear${i}`}
+          >
+            Anticipated Funding Amount (Fiscal Year {i})
```
BCerki, line 23: We could get the fiscal years from the rfp year

schema/deploy/computed_columns/project_revision_anticipated_funding_amount_by_fiscal_year.sql
```diff
@@ -0,0 +1,42 @@
...
+create or replace function cif.project_revision_anticipated_funding_amount_per_fiscal_year(project_revision cif.project_revision)
+returns setof cif.sum_by_fiscal_year
+as
+$computed_column$
+
+with form_changes as (
+    select * from cif.form_change
+    where project_revision_id = $1.id
+    and form_data_schema_name='cif'
+    and form_data_table_name='reporting_requirement'
```
BCerki, line 14: I used `reporting_requirement` instead of milestone because there are other reports that can have payments. Calculation doc specifically referred to milestones though

app/lib/theme/AnticipatedFundingPerFiscalYearArrayFieldTemplate.tsx
```diff
@@ -0,0 +1,91 @@
+import { ArrayFieldTemplateProps } from "@rjsf/core";
```
BCerki, line 1: I ~*think we need to use an `ArrayFieldTemplate` because [array types can use custom widgets for individual fields](https://react-jsonschema-form.readthedocs.io/en/latest/usage/arrays/#:~:text=document.getElementById(%22app%22))%3B-,Custom%20widgets,-%C2%B6) but entire-array widgets [don't seem to be suppported](https://github.com/rjsf-team/react-jsonschema-form/issues/573#:~:text=The%20current%20ArrayField%20implementation%20doesn%27t%20support%20a%20custom%20widget)

app/components/Form/ProjectFundingAgreementForm.tsx
```diff
@@ -91,6 +91,15 @@ const ProjectFundingAgreementForm: React.FC<Props> = (props) => {
         projectFormChange {
           formDataRecordId
         }
+        anticipatedFundingAmountPerFiscalYear {
+          edges {
+            # eslint-disable-next-line relay/unused-fields
```
BCerki, line 96: Not sure how to avoid this as the whole node needs to be passed via `formContext`

schema/deploy/computed_columns/project_revision_anticipated_funding_amount_by_fiscal_year.sql
```diff
@@ -0,0 +1,42 @@
...
+    and (new_form_data->>'reportType') in (select name from cif.report_type where has_expenses=true)
```
BCerki, line 15: @dleard, it does only need to be milestones in these calculations (TEIMP payment is holdbacks, so those were already paid to csnr). Would you suggest changing this function then?

app/components/Form/ProjectFundingAgreementForm.tsx
tmastrom: Could this be turned into it's own fragment?

app/lib/theme/AnticipatedFundingPerFiscalYearArrayFieldTemplate.tsx
```diff
@@ -0,0 +1,96 @@
...
```
pbastia, line 1: I don't see this file used anywhere (it's commented out in the `FormWithTheme`)

app/tests/unit/components/Form/ProjectFundingAgreementFormSummary.test.tsx
```diff
@@ -211,6 +221,7 @@ describe("The Project Funding Agreement Form Summary", () => {
       },
     });
     componentTestingHelper.renderComponent();
+    screen.logTestingPlaygroundURL();
```
pbastia, line 224: left behind
```suggestion

```

schema/test/unit/util_functions/get_fiscal_year_from_timestamp_test.sql
```diff
@@ -0,0 +1,19 @@
+begin;
+select plan(12);
+
+select is((select cif.get_fiscal_year_from_timestamp('2000-01-01 16:21:42.693489-07')), '1999/2000');
+select is((select cif.get_fiscal_year_from_timestamp('2000-02-01 16:21:42.693489-07')), '1999/2000');
+select is((select cif.get_fiscal_year_from_timestamp('2000-03-01 16:21:42.693489-07')), '1999/2000');
+select is((select cif.get_fiscal_year_from_timestamp('2000-04-01 16:21:42.693489-07')), '2000/2001');
```
pbastia, line 7: maybe we can test the edge condition with "March 31 at 23:59:59" and "Apr 1st at 00:00:00"

pbastia (changes requested):
Just a few small things. Nice!

app/lib/theme/AnticipatedFundingPerFiscalYearArrayFieldTemplate.tsx
```diff
@@ -0,0 +1,96 @@
...
+) => {
+  const { formContext } = props;
+
+  console.log("formContext", formContext);
+
+  const areNodesNull =
+    formContext.anticipatedFundingArray.edges.filter(({ node }) => node)
+      .length === 0;
+
+  if (areNodesNull) {
```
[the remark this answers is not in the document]
BCerki: This is an old file I forgot to delete. 

As far as the error message, @nanyangpro asked us to use "This field cannot be calculated due to lack of information now." for any missing calculated values and we'll get make more specific messages in an enhancement ticket later.
​​​​​​​

pbastia (approved):
👌 👌

PR #1321, opened 2023-01-09
Chore: pgo artifactory images

dleard:
Use images in artifactory for the Postgres Operator, fixes an image-pull error when using the crunchy hosted images
plus a couple of tweaks to the db backup tests


mikevespi (approved):
Looks good!

PR #1337, opened 2023-01-13
Feat: Immutable funding stream

mikevespi:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1100

Render the funding stream fields with the read only widget.
Updated the relevant integration and db tests.


app/components/Form/ProjectForm.tsx
```diff
@@ -123,6 +123,14 @@ const ProjectForm: React.FC<Props> = (props) => {
           projectRevisionByProjectRevisionId {
             rank
           }
+          asProject {
```
tmastrom, line 126: The console has optimisticResponse warnings when editing the project overview. I think these fragment updates need to be propagated to the project revision.

tmastrom (changes requested):
Looks good but the integration tests: `project-revision/index` and `undo-changes` are going to need updates due to changes in how the funding stream and year are displayed and cannot be interacted with on revision.

tmastrom (changes requested):
Project Status is not reset when the undo button is clicked. See `undo-changes.spec` line 28. Test passes if the previously entered value is used for the test and this can also be seen when manual testing.

tmastrom (approved):

PR #1339, opened 2023-01-13
Funding agreement diff

Sepehr-Sobhani:
[ZH CARD](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1043)


mikevespi (commented):
Looks good! I stumbled across one case where the text doesn't get the background colouring. When the only thing that changed in a revision is that I removed the funding agreement, it doesn't get the styling for `diffOld`. The easiest way to reproduce is to reset your database, go into a project, resume the edition, remove the funding agreement, and go to the summary page. Let me know if you'd like to hop on a call about it!

mikevespi (approved):
Awesome fix!

PR #1343, opened 2023-01-16
Chore: remove directory from uninstall command

dleard:
had an unnecessary directory in the command that was causing noise in the output


pbastia (approved):
👌

Issue #1151, opened 2022-10-26
Milestone report's next report due link is not aligned with the correct report[7]


Sepehr-Sobhani:
Describe the Bug:

Milestone report's next report due link is not aligned with the correct report. removing one report from the middle can cause the link to show the wrong number and anchor.

Probability (how likely the bug is to happen, scored from 1-5):3
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5):4
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:
create multiple milestone reports → make the last report as the next report due → remove the last report → next report due link is still referring to that last report which does not exist anymore

Screenshots:
https://www.loom.com/share/b7dd8e1b496f49ea9d98d81213a6f988

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

LindsayMacfarlane:
Amazing instructions and video @Sepehr-Sobhani! Thank you for tackling the bug @gurjmatharu!  I could no longer reproduce the bug while following the steps outlined above.

PR #1369, opened 2023-01-20
fix: milestone report due date updates according to milestones

gurjmatharu:
[Card](https://github.com/bcgov/cas-cif/issues/1151)

### DIFF
--- a/app/components/Form/ProjectAnnualReportForm.tsx
+++ b/app/components/Form/ProjectAnnualReportForm.tsx
@@ -72,6 +72,7 @@ const ProjectAnnualReportForm: React.FC<Props> = (props) => {
         upcomingAnnualReportFormChange: upcomingReportingRequirementFormChange(
           reportType: "Annual"
         ) {
+          id
           # eslint-disable-next-line relay/must-colocate-fragment-spreads
           ...ReportDueIndicator_formChange
           asReportingRequirement {
@@ -162,6 +163,14 @@ const ProjectAnnualReportForm: React.FC<Props> = (props) => {
     projectRevision.emissionIntensityReportingRequirementFormChange.edges[0]
       ?.node;
 
+  const upcomingReportFormChange =
+    projectRevision.upcomingAnnualReportFormChange;
+
+  var indexOfNextAnnualReport =
+    sortedAnnualReports.findIndex(
+      (x) => x.id === upcomingReportFormChange?.id
+    ) + 1;
+
   return (
     <div>
       <header>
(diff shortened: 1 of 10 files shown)


app/mutations/MilestoneReport/updateMilestone.ts
```diff
@@ -21,6 +21,7 @@ const mutation = graphql`
         }
         asReportingRequirement {
           ...CollapsibleReport_reportingRequirement
+          reportDueDate
```
BCerki, line 24: `reportDueDate` is already in `...CollapsibleReport_reportingRequirement` so I don't know that this solves it?

BCerki (commented):
The dates look okay but the milestone number is still off: https://www.loom.com/share/e0cba5f5dfac4c3dac8d7c91e3e0300a

Sepehr-Sobhani (commented):
@gurjmatharu I've tested this manually and the `Milestone Report [index]` link is not getting updated with the correct index. Here's what I did:
- Created 3 milestone reports
- First one with the latest due date
- Second one with the earliest due date
- Third one with the due date between the two other reports
- deleted the first report
- Link still shows `Milestone Report 2` while it should show `Milestone Report 1`

app/components/Form/ProjectAnnualReportForm.tsx
```diff
@@ -162,6 +163,14 @@ const ProjectAnnualReportForm: React.FC<Props> = (props) => {
     projectRevision.emissionIntensityReportingRequirementFormChange.edges[0]
       ?.node;
 
+  const upcomingReportFormChange =
+    projectRevision.upcomingAnnualReportFormChange;
+
+  var indexOfNextAnnualReport =
+    sortedAnnualReports.findIndex(
+      (x) => x.id === upcomingReportFormChange?.id
+    ) + 1;
```
Sepehr-Sobhani, line 172: Since this code is repeated across other forms as well, it would be great if we can make it a helper function but not a big deal.

Sepehr-Sobhani: Also, being nitpicky :) It would possibly be better if replace `var` with `let` or `const`

Sepehr-Sobhani (approved):
Great Work! just left some small comments.

app/components/Form/ProjectAnnualReportForm.tsx
dleard: Hm, I thought we had an eslint rule for this

Issue #1383, opened 2023-01-24
emission intensity performance calculation should be based off of the form_change values


dleard:
The calculation for emission intensity performance is currently deriving the values for the calculation from the emission_instensity_report table. This means that the calculated value from that function will not update until the project revision is submitted. The original work for this was done in the summer in card #766. I think sometime after this, we found a better way to do calculated values from form_change records as opposed to deriving them from the table and populating them on the front-end via the form_change_as_table computed columns.

This work is currently blocking #1210 as it needs to use the updated calculated values within another function.

What needs to be done:

- [x] deprecate the `emission_intensity_ei_performance_calculation` computed column
- [x] create a `form_change_ei_performance_calculation` computed column that uses the method described in #766, but derives the values needed for the calculation from the form_change record.
- [x] update `project_revision_teimp_payment_percentage` to use the new function in place of the old one
- [x] update the `ProjectEmissionIntensityReportForm` and the `ProjectEmissionIntensityReportFormSummary` components to use the new computed column (and remove the unnecessary asEmissionIntensityReport part of the fragment)
- [x] Write / update tests for the changes

BCerki:
To test this one, you can go to the emissions form and enter in BEI, TEI, and PEI, and the GHG calculation should recalculate every time you change anything.

LindsayMacfarlane:
Thank you @BCerki! Tested as per your instructions above and can confirm that when I adjust BEI, TEI, PEI then then GHG Emission Intensity Performance auto calculates.

PR #1389, opened 2023-01-24
Bugfix: fix EI report calculated values

BCerki:
card: https://github.com/bcgov/cas-cif/issues/1383


schema/deploy/computed_columns/form_change_calculated_ei_performance.sql
```diff
@@ -0,0 +1,25 @@
...
+returns numeric
+as
+$fn$
+
+  select
+    case
+      when (fc.new_form_data->>'baselineEmissionIntensity' is null)
+        or (fc.new_form_data->>'targetEmissionIntensity' is null)
+        or (fc.new_form_data->>'postProjectEmissionIntensity' is null)
+        or (fc.new_form_data->>'baselineEmissionIntensity' = fc.new_form_data->>'targetEmissionIntensity')
```
gurjmatharu, line 15: should the case `or (fc.new_form_data->>'baselineEmissionIntensity' = fc.new_form_data->>'targetEmissionIntensity')` return a non-none value?

BCerki: If baseline and target are equal, that produces dividing by 0, and I've always seen this handled using `null`. Is there something else that's better to return?

pbastia: Should we test
```
(fc.new_form_data->>'baselineEmissionIntensity')::numeric - (fc.new_form_data->>'targetEmissionIntensity')::numeric) != 0
```
instead?

pbastia (dismissed):

pbastia (dismissed):

pbastia (approved):
Very nice!

PR #1399, opened 2023-01-30
Project Summary Report

tmastrom:
Card: https://github.com/bcgov/cas-cif/issues/1061

PR #1400, opened 2023-01-30
Feat/ia funding

BCerki:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1053

As discussed in the Teams Developer channel, the approach is to create two separate funding schemas rather than two separate funding forms

~~The happo diff in the total project value calculation (additional funding sources not being added) comes from an error with the dev data. Card to fix that here: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1435.~~


app/mutations/FundingParameter/updateFundingParameterFormChange.ts
```diff
@@ -4,8 +4,7 @@ import useDebouncedMutation from "mutations/useDebouncedMutation";
 
 /**
  *
- * Specific mutation for a form_change for a project.
- * Includes whether the suggested proposal reference is unique in the system.
+ * Specific mutation for a form_change for a funding agreement.
```
BCerki, line 7: Not sure why we weren't using this mutation before, but saved me having to make it

schema/deploy/computed_columns/form_change_calculated_total_payment_amount_to_date.sql
```diff
@@ -0,0 +1,25 @@
...
+
+select (($1.new_form_data->>'maxFundingAmount')::numeric*.8) +
+  coalesce(
+    ((select new_form_data->>'projectSummaryReportPayment'
+      from cif.form_change
+      where project_revision_id = $1.project_revision_id
+      and json_schema_name = 'project_summary_report'
+      and operation != 'archive'
+      order by id desc
+      limit 1)::numeric),
```
BCerki, line 18: This is so we get the most recent one. Not sure why we haven't needed to do this in other calculation computed columns

schema/revert/migrations/002_separate_funding_schemas_for_EP_and_IA.sql
```diff
@@ -0,0 +1,7 @@
+-- Revert cif:migrations/002_separate_funding_schemas_for_EP_and_IA from pg
+
+begin;
+
+-- no revert
```
BCerki, line 5: I think this one actually is revert-able because we could set all the schema names back to `funding_parameter`, but migration 001 said it's one way so I followed suit

schema/verify/data/insert_json_schema_form_data_separate_EP_and_IA_funding_schemas.sql
```diff
@@ -0,0 +1,14 @@
+-- Verify cif:data/insert_json_schema_form_data_separate_EP_and_IA_funding_schemas on pg
+
+begin;
+
+do $$
+  begin
+    assert (
+      -- (select count(*) from cif.form) = 15
```
BCerki, line 8:
```suggestion
```

schema/deploy/data/insert_json_schema_form_data_separate_EP_and_IA_funding_schemas.sql
```diff
@@ -0,0 +1,20 @@
...
+begin;
+
+with rows as (
+insert into cif.form(slug, form_change_commit_handler, json_schema, description)
+values
+-- funding_parameter_IA and funding_parameter_EP to be removed when funding agreement form gets refactored to be one form change, just here to pass the fkey
+('funding_parameter_EP', default, '{}'::jsonb, 'OBSOLETE (see comment above) schema data relating to funding_parameter_EP'),
+('funding_parameter_IA', default, '{}'::jsonb, 'OBSOLETE (see comment above) schema data relating to funding_parameter_IA')
+-- ,
+-- ('project_summary_report', default, '{}'::jsonb, 'OBSOLETE (see comment above) schema data relating to project_summary_report')
```
BCerki, line 12:
```suggestion
```

schema/deploy/computed_columns/form_change_calculated_total_payment_amount_to_date.sql
```diff
@@ -0,0 +1,17 @@
...
+
+begin;
+
+create or replace function cif.form_change_calculated_total_payment_amount_to_date(cif.form_change)
+returns numeric
+as
+$fn$
+
+-- TODO: placeholder
+select 0;
```
BCerki, line 11: This calculation depends on the Project Summary Form (#1399 ) and will be handled in https://github.com/bcgov/cas-cif/issues/1419

Sepehr-Sobhani (changes requested):
Great job here 🤩 Just two minor issues, but I've found a bigger issue with IA stream projects. when I create an IA stream project and fill out funding agreement, everything seems to be fine but when I go back to edit the funding agreement, changes are not showing. I feel like this should be the issue with `create_project_revision` function.

schema/verify/data/insert_json_schema_form_data_separate_EP_and_IA_funding_schemas.sql
```diff
@@ -0,0 +1,13 @@
+-- Verify cif:data/insert_json_schema_form_data_separate_EP_and_IA_funding_schemas on pg
+
+begin;
+
+do $$
+  begin
+    assert (
+      (select count(*) from cif.form) = 14
```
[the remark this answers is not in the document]
BCerki: It's 14 in this verify file because the `funding_parameter` slug isn't deleted until `cif_private.funding_form_changes_to_separate_schemas` is called

BCerki:
> Great job here star_struck Just two minor issues, but I've found a bigger issue with IA stream projects. when I create an IA stream project and fill out funding agreement, everything seems to be fine but when I go back to edit the funding agreement, changes are not showing. I feel like this should be the issue with `create_project_revision` function.

Yes, it was a problem with `create_project_revision`. I made the same mistake in the dev data (cif.funding_stream.id is not the same thing as cif.funding_stream_rfp.id), and I've made a card to fix it: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1435. FYI @tmastrom

Sepehr-Sobhani:
> > Great job here star_struck Just two minor issues, but I've found a bigger issue with IA stream projects. when I create an IA stream project and fill out funding agreement, everything seems to be fine but when I go back to edit the funding agreement, changes are not showing. I feel like this should be the issue with `create_project_revision` function.
> 
> Yes, it was a problem with `create_project_revision`. I made the same mistake in the dev data (cif.funding_stream.id is not the same thing as cif.funding_stream_rfp.id), and I've made a card to fix it: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1435. FYI @tmastrom

Nice catch!

schema/verify/data/insert_json_schema_form_data_separate_EP_and_IA_funding_schemas.sql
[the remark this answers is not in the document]
Sepehr-Sobhani: hmmm 🤔 When you run `sqitch verify` is that happy?!

app/components/Form/ProjectFundingAgreementForm.tsx
```diff
@@ -139,6 +160,12 @@ const ProjectFundingAgreementForm: React.FC<Props> = (props) => {
     props.projectRevision
   );
 
+  const fundingStream =
+    projectRevision.projectFormChange.asProject
+      .fundingStreamRfpByFundingStreamRfpId.fundingStreamByFundingStreamId.name;
+
+  const isFundingStreamEP = fundingStream === "EP";
```
tmastrom, line 167: It could be good to verify that the funding stream is "IA" and not just some erroneous value (see `formStructure.tsx`). This could also catch errors later if another stream is added

app/components/Form/ProjectFundingAgreementForm.tsx
```diff
@@ -167,6 +194,10 @@ const ProjectFundingAgreementForm: React.FC<Props> = (props) => {
     projectRevision.additionalFundingSourceFormChanges.edges.filter(
       ({ node }) => node.operation !== "ARCHIVE"
     );
+  // putting the conditional directly in the mutation throws errors
+  const jsonSchemaName = isFundingStreamEP
+    ? "funding_parameter_EP"
+    : "funding_parameter_IA";
```
tmastrom, line 200: same comment as above about having an explicit check for "IA"

app/mutations/FundingParameter/updateFundingParameterFormChange.ts
tmastrom: Will we not need the generic `updateFormChange` once the funding page is refactored to a single form as noted in [insert_json_schema_form_data_separate_EP_and_IA_funding_schemas.sql](https://github.com/bcgov/cas-cif/pull/1400/files#diff-b21441d8fc81741690b9066a831d732af88890a3f8f80ffd8cd82fb7a1164ddc)?

app/components/Form/ProjectFundingAgreementForm.tsx
mikevespi: If we are concerned about it, an alternative is to have a check that it's either IA or EP, then keep the single `isFundingStreamEP` variable

app/mutations/FundingParameter/updateFundingParameterFormChange.ts
BCerki: I think we'll continue to use custom update form change functions even after the refactor. Whenever a form needs to update things that aren't in `updateFormChange` (so in this case, some funding-specific computed columns), we can use a custom function so we're not grabbing those extra things for other forms that use `updateFormChange` but don't need the extras. The milestone form has been refactored and still has a custom update.

tmastrom: The only component using the generic `updateFormChange` is the funding parameter. It's also used in the FormPageFactory. I'm wondering if it's a better pattern to use custom mutations for specific components so changes are decoupled.

app/components/Form/ProjectFundingAgreementForm.tsx
BCerki: I think I'm not concerned about erroneous funding stream values because the validation plugin and db foreign key errors would have gone off before this point (but feel free to disagree strongly and I can change). 

If we get any other funding streams in the future, this can be changed then

Sepehr-Sobhani (dismissed):
👌

Sepehr-Sobhani (approved):

Issue #726, opened 2022-06-09
As a CIF user, I want to sort the projects by milestone due date, so that I can see which project needs attention next


pbastia:
#### Description:

Missing piece from #469, we need to be able to sort by next requirement due date

![image.png](https://images.zenhubusercontent.com/5faafee70e671ac198e88bc7/604f9521-f999-45f7-b079-d3cb4325c432)

note: the search filed is not applicable here

#### Acceptance Criteria:

Given I am a CIF administrator
When I am on the projects lists page
Then I can sort the projects by next requirement due date

#### Development Checklist:

- [x] The table header should use a `SortableHeader` in combination with a `SortOnlyFilter`
- [x] a computed column `project_next_milestone_due_date` returning the next milestone due date for a project
- [x] `project_next_milestone_due_date` is configured with `sortable: true` in `postgraphile.tags.json5`
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

pbastia:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/726?planning-poker) with ZenHub @BCerki @dleard @gurjmatharu @matthieu-foucault @mikevespi @Sepehr-Sobhani

pbastia:
@LindsayMacfarlane we just need a priority tag on this one

LindsayMacfarlane:
[...]

PR #1418, opened 2023-02-09
Sortable Milestone Due Dates

sam-warren:
# Card: https://github.com/bcgov/cas-cif/issues/726

I would appreciate if someone could take a close look at `schema/deploy/computed_columns/project_next_milestone_due_date.sql` for correctness. The way I've implemented it is to just fetch the earliest milestone due date for a project. I initially tried to implement a solution where the query returned the earliest milestone due date for a project after the current timestamp (to only return upcoming milestone due dates) but ran into issues when a milestone's latest due date was in the past, in the case of a completed milestone. Open to recommendations on this!

### No sorting
Screenshot from 2023-02-08 20-40-34

### Next Milestone Due Date (soonest to latest)
Screenshot from 2023-02-08 20-40-22

### Next Milestone Due Date (latest to soonest)
Screenshot from 2023-02-08 20-40-16


tmastrom (changes requested):
It would be good to add a unit test to ensure the filter is working.

tmastrom (dismissed):

tmastrom (approved):

PR #1423, opened 2023-02-11
Feat/routinely updated field widget

Sepehr-Sobhani:
[ZH Card](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1382)

### DIFF
--- a/app/components/ProjectRevision/routinelyUpdatedFieldFactory.tsx
+++ b/app/components/ProjectRevision/routinelyUpdatedFieldFactory.tsx
@@ -1,0 +1,147 @@
+import { WidgetProps } from "@rjsf/core";
+import { Button } from "@button-inc/bcgov-theme";
+import React, { useEffect, useMemo, useState } from "react";
+// import { UseDebouncedMutationConfig } from "mutations/useDebouncedMutation"; //TODO: bring this back when we have the mutation
+// import { Disposable, MutationParameters } from "relay-runtime"; //TODO: bring this back when we have the mutation
+import { DateTime } from "luxon";
+
+/**
+ * A factory function that creates a Field/Widget/React component for a routinely updated field.
+ *
+ * @param Widget The widget to wrap like a TextWidget, SelectWidget, etc.
+ * @param mutationFnHook The hook that returns the mutation function and a boolean indicating whether the mutation is in progress
+ * @param fieldNameToUpdate The name of the field to update (e.g. "status")
+ * @param formDataTableName The name of the form data table to use for updating the field (e.g. "project")
+ * @returns a React component that wraps the widget and adds the update button and functionality
+ *
+ * How to use:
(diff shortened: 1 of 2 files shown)


app/components/ProjectRevision/routinelyUpdatedFieldFactory.tsx
```diff
@@ -0,0 +1,147 @@
...
+  // mutationFnHook: () => [
+  //   (config: UseDebouncedMutationConfig<MutationParameters>) => Disposable,
+  //   boolean
+  // ] //TODO: bring this back when we have the mutation
+  // fieldNameToUpdate: string, //TODO: bring this back when we have the mutation
+  // formDataTableName: string //TODO: bring this back when we have the mutation
+): React.FC<WidgetProps> => {
+  const WrappedWidget: React.FC<WidgetProps> = (props) => {
+    // helper function to compare the initial value to the current value
+    const valueUnchanged = useMemo(
```
BCerki, line 36: I'm not sure about the name `valueUnchanged` This is the value that shows up before the user touches anything in the widget? I think `initalValue` like you've got in the comment above might be clearer

BCerki (commented):
I see what you mean about wanting to have the mutation in place before merging, but there's [another card](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1392) for linking the widget and the mutation, and that will include tests. So I think you've done everything your card requires.

app/components/ProjectRevision/routinelyUpdatedFieldFactory.tsx
Sepehr-Sobhani: This is a mini helper function to compare the untouched value with the current value and does some job based on that. I think `initialValue` would conflict with the existing `initialValue` state but any other name would be acceptable :) I'm not good at naming vars :D

Sepehr-Sobhani:
> I see what you mean about wanting to have the mutation in place before merging, but there's [another card](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1392) for linking the widget and the mutation, and that will include tests. So I think you've done everything your card requires.

Oh true, I didn't consider that card. so I think this card can go forward.

app/components/ProjectRevision/routinelyUpdatedFieldFactory.tsx
BCerki: Oh right, `initialValue` is already taken. Maybe something like... `committedValue`? `previouslyCommittedValue`? But I don't think it's a big deal or a necessary change

BCerki (dismissed):

BCerki (approved):

Issue #1431, opened 2023-02-14
As a developer, I want an external applicant-specific layout component, so that I can use on all external applicant pages


BCerki:
#### Description:

The existing `DefaultLayout` component contains internal-specific elements. This card is about creating a default layout component specific to the external operator.

Wireframe: https://www.figma.com/file/84BpebYIR1iYetxJThLsW6/CIF---External-User-Portal?node-id=209%3A562&t=YNq955131rxLiI2K-0

#### Acceptance Criteria:

Given I am a developer
When I work on external operator cards
Then I have access to an external operator default layout component

Given I am using the external operator default layout component
When I view the component on the front-end
Then I see the same header as the internal operator layout
And a subheader with the links from the wireframe 
And a footer with the links from the wireframe

#### Development Checklist:

- [ ] Create a file containing the footer links (look at `data/dashboardLinks/footerLinks` for an example). For any unconfirmed links, just make a dead link (see below for link list)
- [ ] Create a file for the subheader links (look at `data/dashboardLinks/subHeaderLinks`). For unconfirmed links, dead link
- [ ] Create an external-specific default layout component (look at `DefaultLayout.tsx`). When using Subheader and Footer components, use the links objects created above as props
- [ ] If necessary, create an external-specific `Navigation` component (the existing one might have enough props to customize it for external operators)
- [ ] If any of the other first slice tickets are completed before this one, replace their `DefaultLayout` component with the external-specific one.
- [ ] Meets the DOD

#### Links Checklist (for the link locations of the sub-header and footer):

_**Sub-header**_
- [x] _Home_ - Leads to project list/dashboard
- [x] _Email Us_ - For now, let's have it point to a mailto function that opens up their Outlook, pointing to this email (will need to confirm this with program area): [CleanBCIndustryFund@gov.bc.ca](mailto:CleanBCIndustryFund@gov.bc.ca)
- [x] _Contact Information_ - For now this will have to be an empty link, meaning it appears as a link but doesn't do anything when you click on it.

_**Footer**_
- [x] _Program Details_ - https://alpha.gov.bc.ca/gov/content/environment/climate-change/industry/cleanbc-industry-fund (will need to confirm with program area if they prefer it to link to this page or something else)
- [x] _Disclaimer_ - https://www2.gov.bc.ca/gov/content/home/disclaimer
- [x] _Privacy_ - https://www2.gov.bc.ca/gov/content/home/privacy
[...]

PR #1451, opened 2023-02-23
feat: external layout component

gurjmatharu:
[card](https://github.com/bcgov/cas-cif/issues/1431)

### DIFF
--- a/app/components/Layout/ExternalLayout.tsx
+++ b/app/components/Layout/ExternalLayout.tsx
@@ -1,5 +1,5 @@
 import { graphql, useFragment } from "react-relay";
-import type { DefaultLayout_session$key } from "DefaultLayout_session.graphql";
+import type { ExternalLayout_session$key } from "ExternalLayout_session.graphql";
 import getConfig from "next/config";
 import Navigation from "components/Layout/Navigation";
 import Footer from "components/Layout/Footer";
@@ -8,18 +8,18 @@ import UserProfile from "components/User/UserProfile";
 import GlobalAlert from "./GlobalAlert";
 import { useContext } from "react";
 import { ErrorContext } from "contexts/ErrorContext";
-import footerLinks from "data/dashboardLinks/footerLinks";
-import subHeaderLinks from "data/dashboardLinks/subHeaderLinks";
 
+import footerLinks from "data/externalLinks/footerLinks";
+import subHeaderLinks from "data/externalLinks/subHeaderLinks";
 const runtimeConfig = getConfig()?.publicRuntimeConfig ?? {};
 
 interface Props {
   title?: string;
-  session: DefaultLayout_session$key;
+  session: ExternalLayout_session$key;
   leftSideNav?: React.ReactNode;
 }
 
-const DefaultLayout: React.FC<Props> = ({
+const ExternalLayout: React.FC<Props> = ({
   children,
   title,
   session: sessionFragment,
@@ -27,7 +27,7 @@ const DefaultLayout: React.FC<Props> = ({
 }) => {
   const session = useFragment(
     graphql`
-      fragment DefaultLayout_session on KeycloakJwt {
+      fragment ExternalLayout_session on KeycloakJwt {
         cifUserBySub {
           ...UserProfile_user
         }
@@ -108,4 +108,4 @@ const DefaultLayout: React.FC<Props> = ({
   );
 };
 
-export default DefaultLayout;
+export default ExternalLayout;
--- a/app/components/Layout/Footer.tsx
+++ b/app/components/Layout/Footer.tsx
@@ -1,15 +1,18 @@
 import Link from "next/link";
 import BCGovFooter from "@button-inc/bcgov-theme/Footer";
-import footerLinks from "data/dashboardLinks/footerLinks";
 import BCGovLink from "@button-inc/bcgov-theme/Link";
 
-const Footer = () => {
+interface Props {
+  links: { name: string; href: string }[];
+}
+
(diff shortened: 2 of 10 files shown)


app/components/Layout/ExternalLayout.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const session = useFragment(
+    graphql`
+      fragment ExternalLayout_session on KeycloakJwt {
+        cifUserBySub {
+          ...UserProfile_user
+        }
+      }
+    `,
+    sessionFragment
+  );
```
BCerki, line 37: I don't think we're doing any true auth for first slice, so I think it would be fine to leave this out and just set the user to null below (might be easier to unit test)

app/data/externalLinks/footerLinks.ts
```diff
@@ -0,0 +1,24 @@
...
+  {
+    name: "Privacy",
+    href: "https://www2.gov.bc.ca/gov/content/home/privacy",
+  },
+  {
+    name: "Accessibility",
+    href: "https://www2.gov.bc.ca/gov/content/home/accessible-government",
+  },
+  {
+    name: "Copywrite",
```
BCerki:
```suggestion
    name: "Copyright",
```

app/data/externalLinks/subHeaderLinks.ts
```diff
@@ -0,0 +1,14 @@
+const subHeaderLinks = [
+  {
+    name: "Email Us",
+    href: { pathname: "mailto:CleanBCIndustryFund@gov.bc.ca" },
+    highlightOn: ["/cif/operator(.*)"],
+  },
+  {
+    name: "Contact Information",
+    href: { pathname: "#" },
+    highlightOn: ["/cif/contact(.*)"],
```
BCerki: Possible to use an empty string (or some other kind of dummy value) here? Contacts is an existing internal page

app/tests/unit/components/Layout/Footer.test.tsx
```diff
@@ -1,9 +1,10 @@
 import Footer from "components/Layout/Footer";
 import { render, screen } from "@testing-library/react";
+import footerLinks from "../../../../data/dashboardLinks/footerLinks";
 
 describe("The Footer Component", () => {
   it("Renders footer links", () => {
```
BCerki, line 7: Might be good to make this a "renders default footer links" test, and then have an additional test for the external footer

app/tests/unit/components/Layout/SubHeader.test.tsx
```diff
@@ -2,12 +2,14 @@ import SubHeader from "components/Layout/SubHeader";
 import { useRouter } from "next/router";
 import { render, screen, cleanup } from "@testing-library/react";
 import { mocked } from "jest-mock";
+import subHeader from "../../../../data/dashboardLinks/subHeaderLinks";
+
 jest.mock("next/router");
 
 describe("The SubHeader Component", () => {
   it("Renders the home link", () => {
```
BCerki, line 11: Same comment as previous test

app/components/Layout/StaticLayout.tsx
```diff
@@ -18,6 +20,7 @@ const StaticLayout: React.FC<Props> = ({
   return (
```
BCerki, line 20: `StaticLayout` is used on the 400/500 and fallback pages, and it looks like the subheader is hidden, but the footer isn't. We could maybe do some conditional logic to show the appropriate footer based on where the user came from, but that seems complicated for a first slice. Maybe we can hide the footer in static layout, or talk to the designers about if the internal/external footers have to be different.

app/components/Layout/Footer.tsx
```diff
@@ -1,15 +1,18 @@
 import Link from "next/link";
 import BCGovFooter from "@button-inc/bcgov-theme/Footer";
-import footerLinks from "data/dashboardLinks/footerLinks";
 import BCGovLink from "@button-inc/bcgov-theme/Link";
 
-const Footer = () => {
+interface Props {
+  links: { name: string; href: string }[];
```
BCerki, line 6: Should this be the same type as the one in Navigation and subheader?

BCerki (commented):
Looking good! I think we could also use a test for `ExternalLayout.tsx` that's similar to `DefaultLayout.test.tsx` (or the static layout test if you decide to drop the user graphql query)

app/components/Layout/StaticLayout.tsx
gurjmatharu: Opened a new issue for this discussion: https://github.com/bcgov/cas-cif/issues/1456

app/tests/unit/components/Layout/ExternalLayout.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+    expect(screen.queryByText("Email Us")).toBeNull();
+    expect(screen.queryByText("Contact Information")).toBeNull();
+  });
+
+  it("should render the subheader links if the user is logged in", () => {
+    componentTestingHelper.loadQuery();
+    componentTestingHelper.renderComponent();
+
+    expect(screen.getByText("Home")).toBeVisible();
+    expect(screen.getByText("Email Us")).toBeVisible();
```
BCerki, line 68: "Contact Us" still shows up (just as text) even if there's no link, right?

gurjmatharu: Added a check for `Contact Information`

BCerki (approved):

PR #1473, opened 2023-02-28
chore: remove need for manually created secret

dleard:
This cronjob was relying on a secret that was created manually in openshift. Removing that dependency and using an helm-created secret.


Sepehr-Sobhani (approved):
What does `host` doing here?!

dleard:
> What does `host` doing here?!

This defines the host connection to the database in our tools namespace that we spin up to test

PR #1503, opened 2023-03-03
feat: external project overview

gurjmatharu:
#1431

### DIFF
--- a/app/tests/unit/pages/project-revision/[projectRevision]/form/[formIndex]/external.test.tsx
+++ b/app/tests/unit/pages/project-revision/[projectRevision]/form/[formIndex]/external.test.tsx
@@ -1,0 +1,87 @@
+import "@testing-library/jest-dom";
+import { screen, within } from "@testing-library/react";
+
+import PageTestingHelper from "tests/helpers/pageTestingHelper";
+import externalCompiledQuery from "__generated__/viewExternalProjectRevisionQuery.graphql";
+import ExternalProjectFormPage from "pages/cif-external/project-revision/[projectRevision]/view";
+import { FormIndexPageQuery } from "__generated__/FormIndexPageQuery.graphql";
+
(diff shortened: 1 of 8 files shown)


BCerki (commented):
There are a couple pieces missing from the AC:
And I see a next button (dead link)
And I see a changes indicator
And I see a return to dashboard link

@suhafa, just wanted to check about a couple of the above because they're different on the internal side:
- internally, changes indicator is only on the edit page, not the view page
- internally, there's no next button; people use the tasklist links to move between forms

BCerki:
> There are a couple pieces missing from the AC: And I see a next button (dead link) And I see a changes indicator And I see a return to dashboard link
> 
> @suhafa, just wanted to check about a couple of the above because they're different on the internal side:
> 
> * internally, changes indicator is only on the edit page, not the view page
> * internally, there's no next button; people use the tasklist links to move between forms

Per discussion with @suhafa , don't include changes indicator but do include next button

app/tests/unit/pages/project-revision/[projectRevision]/form/[formIndex]/external.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+describe("The Project Overview Page (external)", () => {
+  beforeEach(() => {
+    externalPageTestingHelper.reinit();
+    externalPageTestingHelper.setMockRouterValues({
+      pathname: "/cif-external/project-revision/[projectRevision]/view",
+      query: { projectRevision: "mock-id" },
+    });
+  });
+
+  it("renders the task list in the left navigation", () => {
```
Comment (BCerki) [inline app/tests/unit/pages/project-revision/[projectRevision]/form/[formIndex]/external.test.tsx:70]: Do we additionally need an `ExternalTaskList.test.tsx`? I think it would end up having mostly the same assertions as this test. Maybe as part of the new card you made?

BCerki (approved):
:cake:

PR #1507, opened 2023-03-06
fix: fix funding_form_changes_to_separate_schemas migration

Sepehr-Sobhani:
[ZH CARD 1505](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1505)


dleard (changes requested):
The changes make sense and the tests cover the change to EP/IA funding_parameter schemas.

This won't fix our deployment problem though. We don't actually want to rework the function, we'll need to apply the fix directly to the `funding_form_changes_to_separate_schemas` function deploy file that is behind the 1.5.0 tag, otherwise the migration uses the original problematic function when it runs.

schema/test/unit/functions/funding_form_changes_to_separate_schemas_test.sql
```diff
@@ -43,13 +79,13 @@ select is (
...
+  3::bigint,
   'All EP projects have had their funding_parameter json_schema_name changed to funding_parameter_EP'
 );
 
 select is(
   (select count(*) from cif.form_change where json_schema_name='funding_parameter_IA'),
-  3::bigint,
+  2::bigint,
   'All IA projects have had their funding_parameter json_schema_name changed to funding_parameter_IA'
 );
```
dleard, line 91: One more test needed here now that I think of it. We should ensure that no other form_change records were changed as a result of this function.

dleard (changes requested):

dleard (approved):

PR #1510, opened 2023-03-07
chore: use main bcgov certbot

dleard:


chart/cas-cif/values.yaml
```diff
@@ -39,7 +39,7 @@ hostName: ~
 
 certbot:
   image:
-    tag: "1.0"
+    pullPolicy: IfNotPresent
```
pbastia, line 42: Any reason to override this value from the original chart?

pbastia (changes requested):
Nice! We should update our own chart version too

chart/cas-cif/values.yaml
dleard: I see the default pullPolicy is 'Always' in the original chart. Which I figured was unnecessary

dleard:
> Nice! We should update our own chart version too

We haven't been updating our chart version for CIF since it's not a chart anybody will be pulling

pbastia (approved):

PR #1511, opened 2023-03-07
Update e2e testing in dev docs

tmastrom:
No Card

A very small update to the developer guide for running e2e tests in headless mode


pbastia (approved):
Nice!
for next time:
- Conventional commits defines a `docs` type, so your commit should have `docs: xyz` as a message
- We try to name our branches according to the conventional commits too, so `docs/e2e-testing` or something like that

pbastia (approved):

PR #1522, opened 2023-03-08
Chore zero and decimals for bei tei pei total life time

Sepehr-Sobhani:
[ZH CARD 1437](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1437)

### DIFF
--- a/app/lib/theme/widgets/ReadOnlyNumberWidget.tsx
+++ b/app/lib/theme/widgets/ReadOnlyNumberWidget.tsx
@@ -5,6 +5,7 @@ const ReadOnlyNumberWidget: React.FC<WidgetProps> = ({
   id,
   value,
   uiSchema,
+  options,
 }) => {
   // If we are using this widget to show numbers as money or percent, we can set `isMoney` or `isPercentage` to true in the uiSchema.
   const isMoney = uiSchema?.isMoney;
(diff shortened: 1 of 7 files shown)


app/lib/theme/widgets/ReadOnlyNumberWidget.tsx
```diff
@@ -18,24 +19,32 @@ const ReadOnlyNumberWidget: React.FC<WidgetProps> = ({
   return (
     <dd>
       {value ? (
```
dleard, line 21: This `value ? ` conditional is causing the summary (and any other readOnly view) to show 'Not Added' for values of 0

dleard (changes requested):
Looks good for the most part. Just a bug in the ReadOnlyNumberWidget where 0 values are being displayed as 'Not Added'. Once that is fixed, some tests should be added to the summary test file for Emission Intensity Report to ensure the readonly values are displaying as expected.

app/lib/theme/widgets/ReadOnlyNumberWidget.tsx
Sepehr-Sobhani: @dleard Thanks for catching this 🙏 This led to a bigger problem with saving zero value and our diffing.

dleard (approved):
:100: awesome work. I checked that zeroes and decimals can be set for the EI form fields, that the zero values are persisted to the database table on submit and that the read-only summary page shows the 0 values properly

Issue #1478, opened 2023-02-28
As a CIF administrator, I want to see the "Payment Percentage of Performance Milestone Amount" between 0 and 100%, so that the number makes sense from a payment perspective


nanyangpro:
#### Description
On the Emissions Intensity Report page, the "Payment Percentage of Performance Milestone Amount" field is a percentage between 0% and 100%. The proponent gets the full/reduced payment or nothing.

If the calculated value goes over 100%, display 100%. If the calculated value goes negative, display 0%.

#### Acceptance Criteria

Given I am a CIF administrator
When I edit the TEIMP report
Then the "Payment Percentage of Performance Milestone Amount" field is clamped between 0 and 100%

Given I am a CIF administrator
When I edit the TEIMP report
Then the "Actual Performance Milestone Amount" field is calculated based on the clamped payment percentage value

Dev checklist

- [x] update postgraphile computed columns to reflect that logic
- [x] make sure tests cover negative and very large calculated payment percentage values

#### Additional context

![image.png](https://images.zenhubusercontent.com/5e3b4990c0f4cb9e2c9b1b1e/18a75be6-957f-4160-b48d-1c2403536664)

Dianadec:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/1478?planning-poker) with Zenhub @pbastia @dleard @tmastrom @Sepehr-Sobhani @BCerki

Dianadec:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/1478?planning-poker) with Zenhub @sam-warren

LindsayMacfarlane:
Thank you @sam-warren. Tested in Dev and it appears correct; however, this will be one we want the program area to review.

PR #1545, opened 2023-03-15
Payment Percentage Computed Column Changes

sam-warren:
Card: https://github.com/bcgov/cas-cif/issues/1478


schema/test/unit/computed_columns/form_change_payment_percentage_test.sql
```diff
@@ -93,6 +111,34 @@ select is(
...
+    from cif.form_change where id=5
+    ) select cif.form_change_payment_percentage((select * from record))
+  ),
+  (
+    0::numeric
+  ),
+  'payments percentage is 0 from the formula: 100 - ((-1.5) x 50 + 145) = -1245 < 0 ==> return 0'
+);
+
 -- 200.00,
```
BCerki, line 142: leftover?

schema/deploy/computed_columns/form_change_payment_percentage.sql
```diff
@@ -9,16 +9,28 @@ create or replace function cif.form_change_payment_percentage(fc cif.form_change
...
+select case when ((fc.new_form_data->>'adjustedEmissionsIntensityPerformance')::numeric is null
+and cif.form_change_calculated_ei_performance(fc) is null)
+  then null
+  else
+    greatest(
+      least(
+        round(
+          100 - ((-1.5) *
+          coalesce(
+            (fc.new_form_data->>'adjustedEmissionsIntensityPerformance')::numeric,
```
BCerki, line 23: I just learned this from Pierre, if a value is coming from an argument, it's clearer to use $1/$2/etc. instead of an alias to avoid confusion with any aliases in the function. (I know the alias was in there before this PR, but since it's already a rework it's a convenient time to refactor)

schema/deploy/computed_columns/form_change_payment_percentage.sql
```diff
@@ -9,16 +9,28 @@ create or replace function cif.form_change_payment_percentage(fc cif.form_change
 returns numeric as
 $fn$
 
-  select
```
BCerki, line 12: Can't comment on unchanged lines, but comment on line 4 should be updated with the 0,100 constraint (or deleted, I don't think we usually put the formula in comments)

schema/test/unit/computed_columns/form_change_payment_percentage_test.sql
```diff
@@ -93,6 +111,34 @@ select is(
...
+  (
+    with record as (
+    select row(form_change.*)::cif.form_change
+    from cif.form_change where id=4
+    ) select cif.form_change_payment_percentage((select * from record))
+  ),
+  (
+    100::numeric
+  ),
+  'payments percentage is 100 from the formula: 100 - ((-1.5) x 500 + 145) = 705 > 100 ==> return 100'
```
BCerki, line 125: can't comment on non-changed lines but formula should be updated on 85 too

schema/test/unit/computed_columns/form_change_payment_percentage_test.sql
sam-warren: Included this comment to be consistent with line 85 of the same file:
`'payments percentage is 30 from the formula: 100 - ((-1.5) x 50 + 145) = 30'`
Can certainly remove if it's not useful!

BCerki: I meant the `-- 200.00` comment floating at the end of the test here

schema/test/unit/computed_columns/form_change_payment_percentage_test.sql
```diff
@@ -93,6 +111,34 @@ select is(
...
+  (
+    with record as (
+    select row(form_change.*)::cif.form_change
+    from cif.form_change where id=4
+    ) select cif.form_change_payment_percentage((select * from record))
+  ),
+  (
+    100::numeric
+  ),
+  'payments percentage is 100 from the formula: greatest(least(100 - ((-1.5) x 500 + 145), 100), 0) = 100'
```
BCerki, line 125: Nit: when giving the formula it might be better to give it in a more readable format than SQL (I've never used `least`/`greatest` before so I had to go look up the parameters and their order). The expected result is already explained on line 114 so I don't think this is a necessary change, but something like  `100 – ((-1.5) x GHG Emission Intensity Performance + 145), values capped at 0 and 100` or whatever in the test output would be clearer for me

BCerki (approved):

schema/test/unit/computed_columns/form_change_payment_percentage_test.sql
sam-warren: Good call - will update this :)

BCerki (approved):

PR #1547, opened 2023-03-15
fix: fetching manager schema from backend

gurjmatharu:
https://github.com/bcgov/cas-cif/issues/1358


app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx
```diff
@@ -325,11 +325,190 @@ describe("The Create Project page", () => {
     expect(
       screen.getByText(/test-funding-stream-description - 2020/i)
     ).toBeInTheDocument();
+    screen.logTestingPlaygroundURL();
+    // expect(screen.getByText(/test manager one/i)).toBeInTheDocument();
+    // expect(screen.getByText(/test manager two/i)).toBeInTheDocument();
+    // expect(screen.getByText(/test manager three/i)).toBeInTheDocument();
```
Comment (tmastrom) [inline app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:331]: Very weird that these commented out tests fail especially given what you can see from the testing playground. We should remove the logging.

app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx
```diff
@@ -325,11 +325,190 @@ describe("The Create Project page", () => {
...
+      screen.getByText(/test-funding-stream-description - 2020/i)
+    ).toBeInTheDocument();
+    screen.logTestingPlaygroundURL();
     expect(screen.getByText(/test manager one/i)).toBeInTheDocument();
     expect(screen.getByText(/test manager two/i)).toBeInTheDocument();
     expect(screen.getByText(/test manager three/i)).toBeInTheDocument();
-    expect(screen.getByText(/test primary contact/i)).toBeInTheDocument();
-    expect(screen.getByText(/test secondary contact/i)).toBeInTheDocument();
+    // expect(screen.getByText(/test primary contact/i)).toBeInTheDocument();
+    // expect(screen.getByText(/test secondary contact/i)).toBeInTheDocument();
```
Comment (tmastrom) [inline app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:511]: Same comment as above

tmastrom (changes requested):

BCerki (commented):
There are a couple more files to change:
- delete the old front-end project managers schema
- remove the front-end schema from `validationSchema`

app/server/middleware/graphql/databaseSchemaConfiguration.ts
```diff
@@ -8,6 +8,8 @@ const DATABASE_SCHEMAS = [
   "milestone",
   "reporting_requirement",
   "project_contact",
+  "project_manager",
+  "operator",
```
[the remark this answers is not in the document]
gurjmatharu: Yes - branch name is incorrect.

app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx
```diff
@@ -157,10 +157,27 @@ describe("The Create Project page", () => {
...
     pageTestingHelper.loadQuery({
-      Form() {
-        return {
-          jsonSchema: projectContactProdSchema,
-        };
+      Form(context) {
+        // If the form is the project manager form, return the project manager schema
+        if (
+          context.name === "formByJsonSchemaName" &&
+          context.path[2] == "projectManagerFormChangesByLabel"
```
Comment (BCerki) [inline app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:164]: == vs. === is necessary or typo?

app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx
```diff
@@ -157,10 +157,27 @@ describe("The Create Project page", () => {
...
+        if (
+          context.name === "formByJsonSchemaName" &&
+          context.path[2] == "summaryContactFormChanges"
+        ) {
+          return {
+            jsonSchema: projectContactProdSchema,
+          };
+        }
+        // Handle default case
+        return { jsonSchema: projectContactProdSchema };
```
Comment (BCerki) [inline app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:180]: Might be better to return `null` or `undefined` if the form isn't given?

app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx
```diff
@@ -157,10 +157,27 @@ describe("The Create Project page", () => {
 
   it("Renders the summary with the filled-out details", async () => {
     pageTestingHelper.loadQuery({
-      Form() {
-        return {
-          jsonSchema: projectContactProdSchema,
-        };
+      Form(context) {
```
Comment (BCerki) [inline app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:160]: This looks awesome

BCerki (commented):
There shouldn't be any happo diffs on this because it's a refactor. Could be flake, but one of them is on the manager page. I'm going to hit re-run and see if they appear again

app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx
Comment (gurjmatharu) [reply app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:164]: Not entirely necessary but this way it ensures the typing is correct as well as equality.

app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx
Comment (gurjmatharu) [reply app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:180]: Using null/undefined will cause the test to break.  Once we have context for all use cases we can adjust to a switch statement to eliminate this line.

Comment (gurjmatharu) [reply app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:180]: tried that too - no luck.  My suggestion is to create a tech debt to refactor the if/else to switch case once we've completed all the other forms.

Comment (gurjmatharu) [reply app/tests/unit/pages/project-revision/[projectRevision]/index.test.tsx:180]: Added an entire schema object - should do the trick for now.

gurjmatharu:
> There shouldn't be any happo diffs on this because it's a refactor. Could be flake, but one of them is on the manager page. I'm going to hit re-run and see if they appear again

I think is related to the default value in project manager schema. Cypress tests expect it to be there - think it's fine?

BCerki:
> > There shouldn't be any happo diffs on this because it's a refactor. Could be flake, but one of them is on the manager page. I'm going to hit re-run and see if they appear again
> 
> I think is related to the default value in project manager schema. Cypress tests expect it to be there - think it's fine?

Or maybe the stuff @Sepehr-Sobhani has been doing with the next update? (how would default value be related?)

BCerki (approved):

tmastrom (approved):

PR #1553, opened 2023-03-16
fix: fetching operator schema from backend

gurjmatharu:
https://github.com/bcgov/cas-cif/issues/1475


tmastrom (commented):
I would recommend removing `"default": "undefined"` from the json schema since the strings are showing up in the frontend when adding a new contact 
Screenshot 2023-03-27 at 3 42 44 PM

app/components/Operator/OperatorForm.tsx
```diff
@@ -1,19 +1,125 @@
...
+        </h2>
+        <SavingIndicator
+          isSaved={!isUpdatingFormChange && !isCommittingFormChange}
+        />
+      </header>
+      <FormBase
+        {...props}
+        id="operator-form"
+        formData={newFormData}
+        schema={props.schema}
```
tmastrom, line 105: why don't we pass a "uiSchema" prop?

gurjmatharu: not needed for operator, as none of the forms have any unique ui

tmastrom (approved):
LGTM

pbastia:
The "Changes Saved" indicator moved under the header instead of in-line on the right

PR #1578, opened 2023-03-30
Update cronjob cron expression

### LINKED ISSUE #1574: Update cronjob cron expression to be greater than "every minute"
Platform services is planning to add a policy that does not allow frequently scheduled cronjobs. Our cronjobs are suspended (will not run until we tell them to), but will still be affected by the policy. We need to increase our cron expressions to have a greater interval.

To do:
Change the cron expression in all cronjob schedules.
From `* * * * *` to `0 * * * * `

This will satisfy the upcoming policy as they will be scheduled to run every hour instead of every minute. (Though they still won't run until we tell them to via airflow because of `suspend: true`)

divyeshio:
Fixes #1574

### DIFF
--- a/chart/cas-cif/templates/cronJobs/cron-swrs-operator-import.yaml
+++ b/chart/cas-cif/templates/cronJobs/cron-swrs-operator-import.yaml
@@ -7,7 +7,7 @@ metadata:
 {{ include "cas-cif.labels" . | indent 4 }}
 spec:
   suspend: true # This cron job is intended to be triggered manually
-  schedule: "* * * * *"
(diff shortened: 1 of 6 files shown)


pbastia:
Hello @divyeshio, thank you for your contribution!
This repository is a work in progress from a software development team from the Climate Action Secretariat with the Government of British Columbia. We open-source all our work under the Apache-2.0 license in the interest of transparency and public accountability, but we're not setup to accept external contributions at this point.

I apologize for not having clear contribution guidelines laid out at this point, it is something we're actively working on.

Issue #1496, opened 2023-03-02
Project contact form errors if there are no contacts in the system [4]


BCerki:
Describe the Bug:

Probability (how likely the bug is to happen, scored from 1-5): 1
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 3
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:
- clear all the contacts from the database
- try to add a contact to a project
- receive error

Screenshots:
https://www.loom.com/share/833d5d31f7c548489b31a27f82b52951

Additional information:
- in `ProjectContactForm`, `filteredProjectContactFormChanges` and `allForms` don't handle empty arrays properly

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

pbastia:
@sam-warren thank you! There is no real way of testing this in -dev so I'll mark this as approved.

pbastia:
@sam-warren thank you! There is no real way of testing this in -dev so I'll mark this as approved.

PR #1604, opened 2023-04-14
Bug: No Contacts Error Fix

sam-warren:
Card: https://github.com/bcgov/cas-cif/issues/1496

Added check for schema.anyOf in widget, now if there are no entries for the search widget, it will render "No options" like below:

Screenshot from 2023-04-14 13-15-33

### DIFF
--- a/app/lib/theme/widgets/SearchDropdownWidget.tsx
+++ b/app/lib/theme/widgets/SearchDropdownWidget.tsx
@@ -23,13 +23,15 @@ const SearchDropdownWidget: React.FC<WidgetProps> = (props) => {
     return selectedValue;
   }, [schema, props.value]);
 
+  const options = schema && schema.anyOf ? schema.anyOf : [];
+
   if (readonly) return <Widgets.SelectWidget {...props} />;
 
   return (
     <Autocomplete
       disableClearable
       id={id}
-      options={schema.anyOf}
+      options={options}
       defaultValue={getSelected()}
       value={getSelected()}
       onChange={handleChange}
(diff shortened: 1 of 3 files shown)


app/tests/unit/components/ProjectRevision/routinelyUpdatedFieldFactory.test.tsx
```diff
@@ -219,3 +219,30 @@ describe("The routinely updated field factory", () => {
     expect(logSpy).toHaveBeenCalledWith("update mutation called");
   });
 });
+
+it("should return 'no options' when no schema.anyOf does not exist or is empty", async () => {
```
BCerki, line 223: We're not doing routinely updated fields anymore, so I believe we'll be removing this file at some point as tech debt (@Sepehr-Sobhani, can you confirm?). I'd suggest moving this test into `SearchDropdown.test.tsx` (and changing the filename to `SearchDropdownWidget.test.tsx` if you're already in there)

BCerki (commented):
Looking good!

app/tests/unit/components/ProjectRevision/routinelyUpdatedFieldFactory.test.tsx
Sepehr-Sobhani: Yes, we no longer need `routinelyUpdatedFieldFactory` and its related test.

sam-warren: Good catch! Thanks for clarifying :)

tmastrom (approved):
LGTM

Sepehr-Sobhani (dismissed):

PR #1610, opened 2023-04-19
Chore: remove redundant widget factory

Sepehr-Sobhani:
No card associated, just cleaning up the code.


tmastrom (approved):
Good housekeeping!

Issue #1068, opened 2022-09-03
Our custom update_form_change() sql mutation will not update archived records by default [8]


dleard:
Describe the Bug:

@Sepehr-Sobhani and I found this issue when looking at the fix for #1040.
When the update_form_change() mutation is passed an operation as part of the patch, then it will update the operation to 'update'. When not passed 'update' as the operation, the `coalesce` in the mutation defaults back to 'archive'.
Archived records are filtered out on the front-end, thus the bug in #1040.

Simply passing 'update' or defaulting the mutation to change it to 'update' is probably not what we want either, as that means that any form_change record with a 'create' operation will automatically be changed to 'update' (this is not the behaviour of the app currently, all form_change records from the first revision have 'create' as the operation).

A workaround has been put in place for now in #1062 where we are conditionally passing either 'create' or 'update' from the front-end depending on the `is_first_revision` column on `project_revision`.

This bug is related to #1066, as if we fix the project_contact form_changes to actually archive properly, then contacts will end up having the same issue as outlined in #1040.

This is only an issue for form_change records that can be archived and updated in the same revision (ie: ones that are derived from intersection tables like project_manager and project_contact...potentially additional_funding_source as well). 

Probability (how likely the bug is to happen, scored from 1-5): 4
Effect (how bad the bug is when it does happen, scored from 1-5): 4

Steps to reproduce the behaviour:

call update_form_change() on an archived form_change record without defining an operation.
because of our `coalesce` in the mutation, the operation will not change from 'archive' to ' update'.

pbastia:
I can see the code handling the null case. Thanks @gurjmatharu !

PR #1062, opened 2022-09-02
Fix/clear existing project manager

Sepehr-Sobhani:
ZH: https://github.com/bcgov/cas-cif/issues/1040

The project manager kept coming as `ARCHIVED` so when choosing a new manager, the form change stays as `ARCHIVED` and doesn't show up in the form.

- Code updated based on the conversation with Dylan. This bug revealed another bugs ([#1066](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/1066) and [#1068](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/1068))
- We are conditionally passing operation here but there's a possibility to change this down the road.

PR #1612, opened 2023-04-19
fix: update_form_change operation field would not update from `archive` to `update`

gurjmatharu:
Fix for #1068


tmastrom (commented):
We can now remove the workaround in [1066](https://github.com/bcgov/cas-cif/pull/1062/files)

Also it could be worthwhile to write a test confirming form_change record with a 'create' operation will NOT automatically be changed to 'update'

schema/test/unit/util_functions/import_swrs_operators_test.sql
```diff
@@ -51,7 +51,7 @@ select results_eq(
       (new_form_data->'swrsOrganisationId')::int,
       operation::text,
       change_status::text
-    from cif.form_change where form_data_table_name='operator'
+    from cif.form_change where form_data_table_name='operator' order by id
```
BCerki, line 54: I wonder why the changes affected this test

schema/deploy/mutations/update_form_change.sql
```diff
@@ -9,8 +9,10 @@ $$
 
   update cif.form_change set
     new_form_data = coalesce(form_change_patch.new_form_data, new_form_data),
-    operation = coalesce(form_change_patch.operation, operation),
-    validation_errors = coalesce(form_change_patch.validation_errors, validation_errors),
+    operation = case
+      when form_change_patch.operation is null and operation = 'archive' then 'update'
+      else coalesce(form_change_patch.operation, operation)
+    end,    validation_errors = coalesce(form_change_patch.validation_errors, validation_errors),
```
BCerki, line 15: I don't totally understand what the card is asking for, so I'm not sure if this change covers everything. In this sentence:
> When the update_form_change() mutation is passed an operation as part of the patch, then it will update the operation to 'update'. When not passed 'update' as the operation, the coalesce in the mutation defaults back to 'archive'.

Is that saying that regardless of what operation `update_form_change` receives, it will always default to "update"? In which case I think that needs addressing?

schema/test/unit/util_functions/import_swrs_operators_test.sql
gurjmatharu: > If sorting is not chosen, the rows will be returned in an unspecified order. The actual order in that case will depend on the scan and join plan types and the order on disk, but it must not be relied on. A particular output ordering can only be guaranteed if the sort step is explicitly chosen.

Not sure what exactly happened but it's something we can't rely on, best practice should be to include an `order by` clause.

schema/deploy/mutations/update_form_change.sql
gurjmatharu: My understanding is that quote is describing current behaviour.  Passing in `UPDATE` or `undefined`/`null` provide different results.

The expected behaviour is that regardless of explicitly stating we want an `UPDATE` or passing in nothing we should get the same result, which is an `UPDATE`.

This is due to the `coalesce` where the first parameter will be `NULL` and second will be `ARCHIVE` so we need to conditionally check for this unique case.

app/components/Form/ProjectManagerFormGroup.tsx
```diff
@@ -137,8 +136,7 @@ const ProjectManagerFormGroup: React.FC<Props> = (props) => {
           rowId,
           formChangePatch: {
             newFormData,
-            // a temporary fix to make sure the form change operation is set correctly
-            operation: projectRevision.isFirstRevision ? "CREATE" : "UPDATE",
+            operation: undefined,
```
dleard, line 139: we can remove this operation line now

app/components/Form/ProjectManagerFormGroup.tsx
```diff
@@ -148,7 +146,7 @@ const ProjectManagerFormGroup: React.FC<Props> = (props) => {
             id: formChangeId,
             changeStatus: "pending",
             newFormData,
-            operation: projectRevision.isFirstRevision ? "CREATE" : "UPDATE",
+            operation: undefined,
```
dleard, line 149: same here, this operation line can be removed

dleard (changes requested):
Looks great! 
I tested it and the mutation is properly assigning update to archived records when appropriate.
Just a couple unnecessary lines to clean up and then should be good to go

dleard (approved):

Issue #1564, opened 2023-03-22
Use stageMultipleReportingRequirementFormChanges function


tmastrom:
### Use stageMultipleReportingRequirementFormChanges function

`stageReportFormChanges` is deprecated. 
The Project Funding Agreement Form and Emission Intensity Report Form should use `stageMultipleReportingRequirementFormChanges` instead. 

The deprecated `stageReportFormChanges` accepts an optional parameter `updateFormChangeMutation` which conditionally stages changes to non-reporting requirement tables. 

The updated `stageMultipleReportingRequirementFormChanges` function relies on a custom commit handler in the database to handle the case where a single form adds data to multiple different tables (see `ProjectMilestoneReportForm`) thus reducing the number of graphql mutation calls. 

###Tech Debt Triage#
The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

Risk Value Scoring:

| Level  | Value |
| ------ | ----- |
| High   | 3     |
| Medium | 2     |
| Low    | 1     |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             |       |       |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     |       |       |
| System Fault Risk - Risk of system errors or application downtime                                                             |       |       |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? |       |       |
| Time Sink Risk - How much will this tech debt slow the development process down                                               |       |       |
| TOTAL SCORE:                                                                                                                  |       |       |

tmastrom:
The value of this is still unclear but could be a good first issue to get familiar with the form change process. Some exploratory work can be found [here](https://github.com/bcgov/cas-cif/pull/1625)

PR #1625, opened 2023-04-26
chore: replace deprecated reporting requirement staging function

tmastrom:
Exploratory work done regarding card [1564](https://github.com/bcgov/cas-cif/issues/1564)

Looks like the deprecated function can be swapped very simply.


tmastrom:
[Tech debt ticket](https://github.com/bcgov/cas-cif/issues/1637) created based on this work

Issue #1569, opened 2023-03-23
As a CIF administrator, I want to see clear context of the changes in Amendments/General Revisions, so that I know what I am about to override


nanyangpro:
#### Description:

As outcome of #1523 (with back-end work handled in #1528), the design of the diff section for an Amendment or General Revision is refined to provide clear context of the changes.

[- Design: Sequential Scenario 1](https://www.figma.com/file/PzOY8RyQnVXnjZJdSv2AqQ/CIF---Internal-User-Portal?node-id=7633%3A4258&t=9ZUtWsS7W7FVvhtv-1)
[- Design: Sequential Scenario 2](https://www.figma.com/file/PzOY8RyQnVXnjZJdSv2AqQ/CIF---Internal-User-Portal?node-id=7636%3A403&t=9ZUtWsS7W7FVvhtv-1)
[- Design: Sequential Scenario 3](https://www.figma.com/file/PzOY8RyQnVXnjZJdSv2AqQ/CIF---Internal-User-Portal?node-id=7636%3A669&t=9ZUtWsS7W7FVvhtv-1)

[- Design (Cleaner Version): preferred by users, nice-to-have, out of scope of this ticket, will be handled in #1573](https://www.figma.com/file/PzOY8RyQnVXnjZJdSv2AqQ/CIF---Internal-User-Portal?node-id=7647%3A912&t=9ZUtWsS7W7FVvhtv-1)

In summary, for the diff section, we'll show the following fields under their forms with help/contextual texts:
- The fields manually updated by the user
- The calculated and auto-populated fields impacted by the manual updates
- The fields changed by other
    - General Revision applied while the current Amendment is open (not approved)
    - Amendment approved while the current General Revision is open (not applied)

#### Acceptance Criteria:

Given I am a user
When I edit *non-calculated* fields on a form and another revision *hasn't* been committed since I opened the revision
Then when I view the "Forms updated" section
Then the old value is in black and struckout 
and if there was no old value (the field was blank) I see "Not entered"

Given I am a user
When I edit fields that affect *calculated* fields on a form and another revision *hasn't* been committed since I opened the revision
Then when I view the "Forms updated" section
Then the old value is in black and struckout 
and if there was no old value (the field was blank) I see "Not calculated"
and I see the help text (see wireframe)

Given I am a user
When I edit fields on a form and another revision *hasn't* been committed since I opened the revision
Then when I view the "Forms updated" section
Then the new value is in red
and there is an arrow between the old and new values

[...]

PR #1632, opened 2023-04-27
refactor: view revision changes in project revisions

gurjmatharu:
#1569

### DIFF
--- a/app/lib/theme/CustomDiffFields.tsx
+++ b/app/lib/theme/CustomDiffFields.tsx
@@ -4,9 +4,6 @@ import NumberFormat from "react-number-format";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faLongArrowAltRight } from "@fortawesome/free-solid-svg-icons";
 import { getLocaleFormattedDate } from "./getLocaleFormattedDate";
-// import Tooltip from "@mui/material/Tooltip";
-// import InfoRounded from "@mui/icons-material/InfoRounded";
-// import IconButton from "@mui/material/IconButton";
 
 const contentSuffixElement = (
   id: string,
@@ -31,7 +28,6 @@ const renderData = (isDate: boolean | undefined, data: string | undefined) => {
 };
 
 const renderTooltip = () => {
-  // TODO: implement this
   return <></>;
 };
 
--- a/app/lib/theme/CustomDiffFields.tsx
+++ b/app/lib/theme/CustomDiffFields.tsx
@@ -20,7 +20,7 @@ const contentSuffixElement = (
   );
 };
 
-const renderData = (isDate: boolean | undefined, data: string | undefined) => {
+const formatData = (isDate: boolean | undefined, data: string | undefined) => {
   if (isDate) {
     return getLocaleFormattedDate(data);
   }
@@ -66,19 +66,19 @@ const showStringDiff = (
               id={id && `${id}-${diffOldClsName}1`}
               className={diffOldClsName}
             >
-              {renderData(isDate, latestCommittedData)}
+              {formatData(isDate, latestCommittedData)}
             </span>
             {contentSuffix && contentSuffixElement(id, contentSuffix)}
             {renderArrow()}
           </>
         )}
         <span id={id && `${id}-${diffOldClsName}`} className={diffOldClsName}>
-          {renderData(isDate, oldData)}
+          {formatData(isDate, oldData)}
         </span>
         {contentSuffix && contentSuffixElement(id, contentSuffix)}
         {renderArrow()}
         <span id={id && `${id}-${diffNewClsName}`} className={diffNewClsName}>
-          {renderData(isDate, newData)}
+          {formatData(isDate, newData)}
         </span>
         {contentSuffix && contentSuffixElement(id, contentSuffix)}
         {renderTooltip()}
@@ -90,12 +90,12 @@ const showStringDiff = (
     return (
       <>
         <span id={id && `${id}-${diffOldClsName}`} className={diffOldClsName}>
-          {renderData(isDate, oldData)}
+          {formatData(isDate, oldData)}
         </span>
         {contentSuffix && contentSuffixElement(id, contentSuffix)}
         {renderArrow()}
         <span id={id && `${id}-${diffNewClsName}`} className={diffNewClsName}>
-          {renderData(isDate, newData)}
+          {formatData(isDate, newData)}
         </span>
         {contentSuffix && contentSuffixElement(id, contentSuffix)}
         {renderTooltip()}
@@ -130,12 +130,12 @@ const showStringDiff = (
     return (
       <>
         <span id={id && `${id}-${diffOldClsName}`} className={diffOldClsName}>
-          {renderData(isDate, latestCommittedData)}
+          {formatData(isDate, latestCommittedData)}
         </span>
         {contentSuffix && contentSuffixElement(id, contentSuffix)}
         {renderArrow()}
         <span id={id && `${id}-${diffNewClsName}`} className={diffNewClsName}>
-          {renderData(isDate, newData)}
+          {formatData(isDate, newData)}
         </span>
         {contentSuffix && contentSuffixElement(id, contentSuffix)}
         {renderTooltip()}
@@ -152,7 +152,7 @@ const showStringDiff = (
         </span>
         {renderArrow()}
         <span id={id && `${id}-${diffNewClsName}`} className={diffNewClsName}>
-          {renderData(isDate, newData)}
+          {formatData(isDate, newData)}
         </span>
         {contentSuffix && contentSuffixElement(id, contentSuffix)}
       </>
--- a/app/lib/theme/CustomDiffFields.tsx
+++ b/app/lib/theme/CustomDiffFields.tsx
@@ -125,6 +125,15 @@ const showStringDiff = (
             >
               {formatData(isDate, oldData)}
             </span>
+
+            {contentSuffix && contentSuffixElement(id, contentSuffix)}
+            {renderArrow()}
+            <span
+              id={id && `${id}-${diffOldClsName}`}
+              className={diffOldClsName}
+            >
+              {formatData(isDate, latestCommittedData)}
+            </span>
             {contentSuffix && contentSuffixElement(id, contentSuffix)}
           </>
         )}
--- a/app/lib/theme/CustomDiffFields.tsx
+++ b/app/lib/theme/CustomDiffFields.tsx
@@ -56,6 +56,7 @@ const showStringDiff = (
     "diffText",
   ];
 
+  // The numbers show the truth values of oldData, newData, latestCommittedData
   // Case 7 ->  111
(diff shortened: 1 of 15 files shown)


app/lib/theme/CustomDiffFields.tsx
```diff
@@ -20,193 +23,161 @@ const contentSuffixElement = (
...
+  isDate: boolean,
+  contentSuffix?: string
+): JSX.Element => {
+  const [diffOldClsName, diffNewClsName, diffTextClsName] = [
+    "diffOld",
+    "diffNew",
+    "diffText",
+  ];
+
+  // Case 7 ->  111
```
BCerki, line 63: suggest adding a comment that explains how the 111 system works

app/lib/theme/CustomDiffFields.tsx
```diff
@@ -20,193 +23,161 @@ const contentSuffixElement = (
...
+  if (oldData && newData && latestCommittedData) {
+    return (
+      <>
+        {latestCommittedData !== oldData && (
+          <>
+            <span
+              id={id && `${id}-${diffOldClsName}1`}
```
BCerki, line 70: Is the one at the end of id supposed to be there?

app/lib/theme/CustomDiffFields.tsx
```diff
@@ -20,193 +23,161 @@ const contentSuffixElement = (
...
+        )}
+        <span id={id && `${id}-${diffOldClsName}`} className={diffOldClsName}>
+          {renderData(isDate, oldData)}
+        </span>
+        {contentSuffix && contentSuffixElement(id, contentSuffix)}
+        {renderArrow()}
+        <span id={id && `${id}-${diffNewClsName}`} className={diffNewClsName}>
+          {renderData(isDate, newData)}
+        </span>
+        {contentSuffix && contentSuffixElement(id, contentSuffix)}
```
BCerki, line 87: This is pretty repetitive, maybe there could be functions for returning styled old, latestCommitted, and new components?

app/lib/theme/CustomDiffFields.tsx
```diff
@@ -20,193 +23,161 @@ const contentSuffixElement = (
...
   }
-  if (isDate) {
-    return `${displayString}${getLocaleFormattedDate(latestCommittedData)}`;
+
+  // Case 5 ->  101
+  if (oldData && !newData && latestCommittedData) {
+    return (
+      <>
+        <span id={id && `${id}-${diffOldClsName}`} className={diffOldClsName}>
+          {isDate ? getLocaleFormattedDate(oldData) : oldData}
```
BCerki, line 115: There are a few of these lines, can they be replaced with the `renderData` function?

app/lib/theme/CustomDiffFields.tsx
```diff
@@ -20,193 +23,161 @@ const contentSuffixElement = (
...
-      {
-        //   latestCommittedData &&
-        overwriteNotification({ latestCommittedData, isDate })
-      }
-    </>
-  );
-};
+  // Case 1 ->  001
+  if (!oldData && !newData && latestCommittedData) {
+    return <>ERROR SD: 1</>;
```
BCerki, line 168: Can this case happen?

app/lib/theme/CustomDiffFields.tsx
```diff
@@ -20,193 +23,161 @@ const contentSuffixElement = (
...
+        {contentSuffix && contentSuffixElement(id, contentSuffix)}
+        {renderTooltip()}
+      </>
+    );
   }
-  if (isDate) {
-    return `${displayString}${getLocaleFormattedDate(latestCommittedData)}`;
+
+  // Case 5 ->  101
+  if (oldData && !newData && latestCommittedData) {
```
BCerki, line 111: I think this one might need a conditional check for if old and latest are equal

app/lib/theme/CustomDiffFields.tsx
```diff
@@ -20,193 +23,161 @@ const contentSuffixElement = (
...
-        <NumberFormat
-          thousandSeparator
-          fixedDecimalScale={true}
-          decimalScale={decimalScale}
-          prefix={isMoney ? "$" : ""}
-          suffix={isPercentage ? " %" : ""}
-          displayType="text"
-          value={latestCommittedData}
-        />
+        <span id={id && `${id}-${diffNewClsName}`} className={diffNewClsName}>
```
BCerki, line 125: If there's no newData, that means something's been removed, right? So should it be diffOld classes?

app/lib/theme/CustomDiffFields.tsx
```diff
@@ -20,193 +23,161 @@ const contentSuffixElement = (
...
+          {isDate ? getLocaleFormattedDate(oldData) : oldData}
+        </span>
+        {contentSuffix && contentSuffixElement(id, contentSuffix)}
+      </>
+    );
+  }
+  // Case 3 ->  011
+  if (!oldData && newData && latestCommittedData) {
```
BCerki, line 133: I think there's supposed to be not entered/not calculated text if there's no old data.

app/lib/theme/CustomDiffFields.tsx
```diff
@@ -20,193 +23,161 @@ const contentSuffixElement = (
...
-        </>
-      )}
-      {
-        // latestCommittedData && oldData && oldData !== latestCommittedData &&
-        overwriteNotification({ latestCommittedData, isDate })
-      }
-    </>
-  );
+  // Case 0 ->  000
+  if (!oldData && !newData && !latestCommittedData) {
```
BCerki, line 172: If there's nothing then I don't think we need to show a diff at all

app/lib/theme/CustomDiffFields.tsx
```diff
@@ -20,193 +23,161 @@ const contentSuffixElement = (
...
+      <>
+        <span id={id && `${id}-${diffTextClsName}`} className={diffTextClsName}>
+          {"Not Entered"}
+        </span>
+      </>
+    );
+  }
 };
 
 const showNumberDiff = (
```
BCerki, line 183: I'll look at the NumberField in more detail in the next round of review--at this point it's pretty much all the same comments as the StringField's

app/lib/theme/CustomDiffFields.tsx
```diff
@@ -20,193 +23,161 @@ const contentSuffixElement = (
...
 
-interface OverwriteNotificationArgument {
-  latestCommittedData: any;
-  isNumber?: boolean;
-  isDate?: boolean;
-  isMoney?: boolean;
-  isPercentage?: boolean;
-  numberOfDecimalPlaces?: number;
-}
+const renderData = (isDate: boolean | undefined, data: string | undefined) => {
```
BCerki, line 26: maybe this could be called `formatData`? At first I thought data was a typo for date

BCerki (commented):
I had a lot of comments on the `CustomDiffFields` file that could affect the rest of the PR, so this is only a partial reveiw of the first half of that file. I'll comment on the rest of the file and tests, etc. in the next round of review.

I haven't done much manual testing yet, but I'm seeing some styling issues with the diffs:
image

In the wireframe there never seems to be strikeout if there's a latest committed value (which might be worth checking with Nan about), and the arrow's a lot smaller

app/lib/theme/CustomDiffFields.tsx
gurjmatharu: I went down this path and found that some elements had embedded spans and made it hard to style.

BCerki (changes requested):
I manually tested the combinations where there's no latest committed data, and then started on the ones with latest committed data, but I turned up comments and bugs that I think it's best to stop here for now and continue the manual testing in another round (sidenote, how are you manually testing the combinations with latest committed? I haven't thought of a good approach)

Bugs:

When we have old and new:
Removing a project manager doesn't display correctly
image

Removing the funding form shows old diff styling:
image

Removing reports shows the old diff styling:
image

When we have old, new, and latest committed:
Calculated diffs (rank) don't show (maybe need to rebase?):
image

Some display errors:
image
(I didn't change these values in any revision)

I'm missing "Not entered" in these diffs:
image

BCerki (commented):
To test this:
- created a project with the minimum info (general revision 1)
- created amendment and made changes but left revision open
- create general revision 2, made changes, and committed

A few things came up:
- diffs appear to display in the wrong order:
image
- I saw different results when I made revisions via the UI vs. cypress or test data. Via the UI I see three diffs, but via cypress or test data latest committed doesn't show. I'll make a branch with that code and details on how to reproduce it. Mike and I thought it might have something to do with the conditionality in CustomDiffFields

app/lib/theme/CustomDiffFields.tsx
```diff
@@ -216,22 +190,18 @@ const showNumberDiff = (
...
-  return (
-    <>
-      <span id={id && `${id}-${diffOldClsName}`} className={diffOldClsName}>
+  const [diffOldClsName, diffNewClsName, diffTextClsName] = [
+    "diffOld",
+    "diffNew",
+    "diffText",
+  ];
+  // case 7 ->  111
+  if (oldData && newData && latestCommittedData) {
```
BCerki, line 202: In every case that has both old and latestCommitted, we need to check if they're equal before doing anything else because if they are equal, then we're only dealing with two diffs.

This might end up covering case 111, 101, and 100

BCerki (commented):
Here are some initial comments from the pair review. We still need to look at the tests, so more comments to come later.

All of the string field comments also apply to the number fields.

app/lib/theme/CustomDiffFields.tsx
```diff
@@ -20,193 +20,167 @@ const contentSuffixElement = (
...
-        overwriteNotification({ latestCommittedData, isDate })
-      }
-    </>
-  );
-};
+  // Case 5 ->  101
+  if (oldData && !newData && latestCommittedData) {
+    return (
+      <>
+        {oldData == latestCommittedData ? (
```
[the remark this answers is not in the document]
BCerki: ===

app/lib/theme/CustomDiffFields.tsx
```diff
@@ -259,141 +270,227 @@ const showNumberDiff = (
...
+          fixedDecimalScale={true}
+          decimalScale={decimalScale}
+          prefix={isMoney ? "$" : ""}
+          suffix={isPercentage ? " %" : ""}
+          displayType="text"
+          value={oldData}
+          className={diffOldClsName}
+          id={id && `${id}-${diffOldClsName}`}
+        />
+        {newData === 0 ? (
```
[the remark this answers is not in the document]
gurjmatharu: see: https://github.com/bcgov/cas-cif/issues/1714

app/lib/theme/CustomDiffFields.tsx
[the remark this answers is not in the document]
gurjmatharu: added a comment to https://github.com/bcgov/cas-cif/issues/1714

gurjmatharu:
closing - N/A

PR #1638, opened 2023-05-01
Feat: 1231 Show calculated diffs on overview form

BCerki:
Card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1231

Notes: 
-`project_revision_rank` was using `project_revision` as an argument but already using the latest `form_change` data for the calculation, so the refactor was just a matter of changing the argument
-checked happo and happy with diffs


schema/deploy/computed_columns/form_change_rank.sql
```diff
@@ -0,0 +1,62 @@
...
+      (
+        select(new_form_data ->>'fundingStreamRfpId')::int from cif.form_change
+        where form_data_table_name='project'
+        and project_revision_id=$1.project_revision_id
+      )
+    -- As soon as there's a committed record for this project we want to exclude that record's score from the ranking calculation because it isn't as up-to-date as the new_form_data
+    and project.id!=(select form_data_record_id from cif.form_change
+        where form_data_table_name='project'
+        and project_revision_id=$1.project_revision_id
+      )
```
tmastrom, line 34: good catch

tmastrom (changes requested):
Rank functionality is working well with the new computed column!

### Frontend Comments: 
Is this what the diff comment should say for the Rank?
Screenshot 2023-05-03 at 2 36 01 PM

Nit: it could be nice to limit the size of the rank message before a value is entered to keep the width of the form consistent 

### Backend Comments: 
`sqitch verify` is failing with the following message 
```
  - tables/report_type_003 ............................................................................ 2023-05-03 14:44:12.077 PDT [28614] ERROR:  update or delete on table "report_type" violates foreign key constraint "reporting_requirement_report_type_fkey" on table "reporting_requirement" 
 ```

BCerki:
> Rank functionality is working well with the new computed column!
> 
> ### Frontend Comments:
> Is this what the diff comment should say for the Rank? Screenshot 2023-05-03 at 2 36 01 PM
> 
> Nit: it could be nice to limit the size of the rank message before a value is entered to keep the width of the form consistent
> 
> ### Backend Comments:
> `sqitch verify` is failing with the following message
> 
> ```
>   - tables/report_type_003 ............................................................................ 2023-05-03 14:44:12.077 PDT [28614] ERROR:  update or delete on table "report_type" violates foreign key constraint "reporting_requirement_report_type_fkey" on table "reporting_requirement" 
> ```

Thanks for catching the "none" latest committed diff! I'll update the rest of the calculated diff cards to include that.

Nit: The message came from the designers back when this widget was first implemented so I'll make a card/ask them

Sqitch verify: I couldn't reproduce this, though when I ran `sqitch verify` I got 17 other errors, so I'll bring that up with in Developers. I'm not sure it's related to this PR

tmastrom (approved):

PR #1641, opened 2023-05-03
Feat: 1561 Editing workflow

BCerki:
card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1561

Note: I decided to do this a little differently from the dev checklist. Instead of using a tasklist mode, I made separate edit and view pages. We've all been confused about this new workflow and where the code lives, and I think separate pages makes it clearer (it's also more consistent with the rest of forms, which have separate form and summary components).

### DIFF
--- a/app/components/ProjectRevision/RevisionStatusWidget.tsx
+++ b/app/components/ProjectRevision/RevisionStatusWidget.tsx
@@ -6,6 +6,8 @@ import SelectWidget from "lib/theme/widgets/SelectWidget";
 import { useState } from "react";
 import { graphql, useFragment } from "react-relay";
 import { useCommitProjectRevision } from "mutations/ProjectRevision/useCommitProjectRevision";
+import { getProjectRevisionViewPageRoute } from "routes/pageRoutes";
+import { useRouter } from "next/router";
 
 const RevisionStatusWidgetFragment = graphql`
   fragment RevisionStatusWidget_projectRevision on ProjectRevision {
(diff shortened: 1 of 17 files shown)


app/cypress/e2e/cif/project-revision/create-subsequent-project-revision.cy.js
```diff
@@ -197,5 +179,7 @@ describe("the project amendment and revisions page", () => {
     cy.findByText("General Comments (optional)")
       .next()
       .contains("test change reason");
+    // Screenshot below is commented out because of flakiness. Documented in ticket #1194.
+    // cy.happoAndAxe("Project Revision View", "view", "main");
```
BCerki, line 183: I'm assuming this will be flaky because it's the same page as one above that's commented out

app/components/ProjectRevision/RevisionStatusWidget.tsx
```diff
@@ -52,6 +56,7 @@ const RevisionStatusWidget: React.FC<WidgetProps> = (props) => {
           },
         },
       },
+      onCompleted: () => router.replace(getProjectRevisionViewPageRoute(id)),
```
tmastrom, line 59: Nice. I haven't used `replace` before but it makes sense here so the user can't use the back button to undo changes

tmastrom (approved):
Great work, I support the separation of the edit/view functionality into separate files to avoid spaghetti.

PR #1645, opened 2023-05-05
fix: show emission intensity data on view mode

Sepehr-Sobhani:
Addressing a bug (reported in Teams) -> "Missing" Emissions info

[ZH CARD 1647](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1647)


BCerki (approved):

BCerki:
I think we will need a card created for this so we can move it to PO approval so everyone knows it's in

Issue #1582, opened 2023-03-30
As a user, I want to see a Preview Revision page (no more Review and Submit page), so that I can use the new committing workflow


BCerki:
#### Description:

Currently, when users finish creating a revision, then they are directed to a "Review and Submit" page (`app/pages/cif/project-revision/[projectRevision]/index.tsx`) where they can see their diffed changes and click "Submit." This is an old workflow. To facilitate the new workflow, users should see a "Preview Revision" page (`app/pages/cif/project-revision/[projectRevision]/view.tsx`) before creating a revision and commit projects via a status dropdown and update button.

- Design: [Preview Revision (the last step of creating a new revision)](https://www.figma.com/proto/PzOY8RyQnVXnjZJdSv2AqQ/CIF---Internal-User-Portal?page-id=2289%3A2663&node-id=3811-3067&viewport=-470%2C1174%2C0.2&scaling=scale-down&starting-point-node-id=3169%3A2715&show-proto-sidebar=1)
- Design : [View Revision (after a revision is created, editable unless the status is changed to "Approved" / "Applied")](https://www.figma.com/proto/PzOY8RyQnVXnjZJdSv2AqQ/CIF---Internal-User-Portal?page-id=2289%3A2663&node-id=7283-630&viewport=-470%2C1174%2C0.2&scaling=scale-down&starting-point-node-id=3169%3A2715&show-proto-sidebar=1)

#### Acceptance Criteria:

Given I am a user
When I finish editing a form (I click "Submit XX Form" button at the end of any form page)
Then I am directed to the project revision page (the one with the radio boxes for amendment, pending actions from, forms updated, `app/pages/cif/project-revision/[projectRevision]/view.tsx`)
and I do not see a "Submit" button

Given I am a user
When I am on the project revision page
Then I can see a "Discard Amendment" or "Discard General Revision" (as applicable) button
And when I click it, I see a confirmation message

Given I am a user
When I view the tasklist
Then I should not see a "Submit Changes" section

Given I am a user
When I view the tasklist
Then I should see an Amendments & Revisions section
and a subsection with the text "New Revision (Preview)"

#### Development Checklist:

- [x] Make the individual form submit button link to the project revision page
- [x] Add the discard button from `app/pages/cif/project-revision/[projectRevision]/index.tsx` to the project revision page
- [x] Remove the yellow/green diff styling from `CustomDiffFields` (only using the strikeout/red styling from now on)
- [ ] Meets the DOD

[...]

PR #1650, opened 2023-05-09
1582 Preview Revision

tmastrom:
Card: https://github.com/bcgov/cas-cif/issues/1582

Note: check the card comments as well as AC as some things changed after discussion with designers

### DIFF
--- a/app/components/TaskList/index.tsx
+++ b/app/components/TaskList/index.tsx
@@ -243,7 +243,7 @@ const TaskList: React.FC<Props> = ({
         })}
 
         {/* Project Summary Section */}
-        {mode !== "view" && (
+        {mode !== "view" && mode !== "update" && (
           <TaskListSection
             defaultExpandedState={currentStep === "summary"}
             listItemNumber={String(
@@ -263,7 +263,7 @@ const TaskList: React.FC<Props> = ({
           </TaskListSection>
         )}
         {/* Amendments & Other Revisions section */}
-        {useShowGrowthbookFeature("amendments") && mode === "view" && (
+        {useShowGrowthbookFeature("amendments") && projectByProjectId && (
           <ProjectRevisionChangeLogsTaskListSection
             projectRevisionId={id}
             defaultExpandedState={[
--- a/app/components/TaskList/index.tsx
+++ b/app/components/TaskList/index.tsx
@@ -4,9 +4,9 @@ import { useStageDirtyFormChanges } from "mutations/FormChange/stageDirtyFormCha
 import {
   getProjectRevisionPageRoute,
   getProjectRevisionAttachmentsPageRoute,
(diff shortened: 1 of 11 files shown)


app/components/TaskList/index.tsx
```diff
@@ -263,14 +266,10 @@ const TaskList: React.FC<Props> = ({
...
-        {useShowGrowthbookFeature("amendments") && mode === "view" && (
+        {useShowGrowthbookFeature("amendments") && mode !== "create" && (
           <ProjectRevisionChangeLogsTaskListSection
             projectRevisionId={id}
-            defaultExpandedState={[
-              getProjectRevisionChangeLogsPageRoute(id).pathname,
-              projectRevisionViewPagePathName,
-              projectRevisionCreatePagePathName,
-            ].includes(router.pathname)}
+            defaultExpandedState={true}
```
BCerki, line 272: There's already conditional logic in TaskListItem to hide the subheading and caret when we don't want to see it, so this can always be expanded

app/components/TaskList/index.tsx
```diff
@@ -263,14 +270,10 @@ const TaskList: React.FC<Props> = ({
...
```
Sepehr-Sobhani, line 276: Since we are not using `ProjectRevisionChangeLogsTaskListSection` anywhere else, I think we can safely remove the `defaultExpandedState` from this component and set the `isExpanded` state always true.

app/cypress/e2e/cif/project-revision/edit-project-revision.cy.js
```diff
@@ -184,79 +187,86 @@ describe("when editing a project, the project page", () => {
...
-    )
-      .next()
-      .next()
-      .should("have.text", "ADDED");
+      "#root_teimpReporting_measurementPeriodStartDate-diffAmendmentsAndOtherRevisionsNew"
+    ).should("have.text", "Jan 1, 2022");
 
-    cy.get("#root_comments-diffReviewAndSubmitInformationOld").should(
+    cy.contains("Jan 1, 2022").next().should("have.text", "(ADDED)");
+    cy.pause();
```
Sepehr-Sobhani, line 257: Cool, I didn't know about this command but I think it's a leftover, right?

Sepehr-Sobhani (changes requested):
Looks great 🤩 Just some minor changes are required.
Also, the AC requires the subsection with the text "New Revision (Preview)" but we don't have the `(Preview)` part, should we add it?

BCerki:
> Looks great star_struck Just some minor changes are required. Also, the AC requires the subsection with the text "New Revision (Preview)" but we don't have the `(Preview)` part, should we add it?

No, just `New Revision` is good per conversation with @nanyangpro

Sepehr-Sobhani (approved):

PR #1654, opened 2023-05-11
Fix: 1651 not using hardcoded rfp ID

Sepehr-Sobhani:
[ZH CARD 1651](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1651)


pbastia (approved):
🥇

PR #1688, opened 2023-05-24
Feat: 979 reusable tooltip component

Sepehr-Sobhani:
[ZH CARD 979](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/979)

NOTE: we need to check accessibility using axe once we render the tooltip in any fields.

### DIFF
--- a/app/lib/theme/colors.ts
+++ b/app/lib/theme/colors.ts
@@ -2,5 +2,5 @@
 export const BC_GOV_LINKS_COLOR: string = "#1A5A96";
 export const BC_GOV_PRIMARY_BRAND_COLOR_BLUE: string = "#003366";
 // App specific colors
-export const TASKLIST_GREY_BG_COLOR: string = "#E5E5E5";
-export const MAIN_BG_COLOR: string = "#fafafc";
+export const DARK_GREY_BG_COLOR: string = "#E5E5E5";
+export const LIGHT_GREY_BG_COLOR: string = "#fafafc";
--- a/app/lib/theme/colors.ts
+++ b/app/lib/theme/colors.ts
@@ -1,0 +1,6 @@
+// BC GOV colors
+export const BC_GOV_LINKS_COLOR: string = "#1A5A96";
+export const BC_GOV_PRIMARY_BRAND_COLOR_BLUE: string = "#003366";
+// App specific colors
+export const TASKLIST_GREY_BG_COLOR: string = "#E5E5E5";
+export const MAIN_BG_COLOR: string = "#fafafc";
--- a/app/lib/theme/widgets/ContextualHelp.tsx
+++ b/app/lib/theme/widgets/ContextualHelp.tsx
@@ -3,6 +3,7 @@ import { faExclamationCircle } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import Tooltip from "@mui/material/Tooltip";
 import { BC_GOV_LINKS_COLOR, DARK_GREY_BG_COLOR } from "../colors";
+import parse from "html-react-parser";
 
 /**
  * @param {string} text - The text of the tooltip to be displayed (can be HTML)
@@ -17,6 +18,7 @@ import { BC_GOV_LINKS_COLOR, DARK_GREY_BG_COLOR } from "../colors";
  *   "icon": faExclamationCircle,
  *   "placement": "top"
  * }
+ * NOTE: we are using html-react-parser package to parse the text of the tooltip to allow for HTML tags and avoid using dangerouslySetInnerHTML
  */
 
 interface ContextualHelpProps {
@@ -49,7 +51,7 @@ const ContextualHelp: React.FC<ContextualHelpProps> = ({
   return (
     <>
       <Tooltip
-        title={<span dangerouslySetInnerHTML={{ __html: text }} />}
+        title={parse(text)}
         arrow
         placement={placement}
         role="tooltip"
--- a/app/lib/theme/widgets/ContextualHelp.tsx
+++ b/app/lib/theme/widgets/ContextualHelp.tsx
@@ -2,7 +2,7 @@ import { IconDefinition } from "@fortawesome/fontawesome-svg-core";
 import { faExclamationCircle } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import Tooltip from "@mui/material/Tooltip";
-import { BC_GOV_LINKS_COLOR, TASKLIST_GREY_BG_COLOR } from "../colors";
+import { BC_GOV_LINKS_COLOR, DARK_GREY_BG_COLOR } from "../colors";
 
 /**
  * @param {string} text - The text of the tooltip to be displayed (can be HTML)
@@ -71,12 +71,12 @@ const ContextualHelp: React.FC<ContextualHelpProps> = ({
           cursor: pointer;
         }
         :global(.MuiTooltip-tooltip) {
-          background-color: ${TASKLIST_GREY_BG_COLOR};
+          background-color: ${DARK_GREY_BG_COLOR};
           box-shadow: 5px 5px 5px #888888;
           color: black;
         }
         :global(.MuiTooltip-tooltip .MuiTooltip-arrow) {
-          color: ${TASKLIST_GREY_BG_COLOR};
+          color: ${DARK_GREY_BG_COLOR};
         }
         :global(.MuiTooltip-tooltip a) {
           color: ${BC_GOV_LINKS_COLOR};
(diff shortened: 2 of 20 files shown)


app/lib/theme/widgets/ContextualHelp.tsx
```diff
@@ -0,0 +1,89 @@
+import { IconDefinition } from "@fortawesome/fontawesome-svg-core";
+import { faExclamationCircle } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import Tooltip from "@mui/material/Tooltip";
+import { BC_GOV_LINKS_COLOR, TASKLIST_GREY_BG_COLOR } from "../colors";
+
+/**
+ * @param {string} title - The title of the tooltip to be displayed (can be HTML)
```
BCerki, line 8: I find "title" misleading because some of our tooltips are going to show sentences and bullets. I know we can't change the MUI prop names, but here we could say "the text the tooltip will display" (or whatever)

Also these params are great documentation

app/tests/unit/lib/theme/widgets/ContextualHelp.test.tsx
```diff
@@ -0,0 +1,36 @@
+import { render, screen } from "@testing-library/react";
+import { faExclamationCircle } from "@fortawesome/free-solid-svg-icons";
+import ContextualHelp from "lib/theme/widgets/ContextualHelp";
+
+describe("ContextualHelp component", () => {
+  const titleMock = "Tooltip Title";
```
BCerki, line 6: If we move away from the word "title", could change to whatever the replacement is here too

app/tests/unit/lib/theme/widgets/ContextualHelp.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    const componentUnderTest = render(
+      <ContextualHelp
+        title={titleMock}
+        icon={faExclamationCircle}
+        label={labelMock}
+      />
+    );
+    const tooltipElement = screen.getByRole("tooltip");
+    const tooltipIcon = screen.getByRole("img", {
+      hidden: true,
```
BCerki, line 19: Why are they hidden?

app/lib/theme/colors.ts
```diff
@@ -0,0 +1,6 @@
+// BC GOV colors
+export const BC_GOV_LINKS_COLOR: string = "#1A5A96";
+export const BC_GOV_PRIMARY_BRAND_COLOR_BLUE: string = "#003366";
+// App specific colors
+export const TASKLIST_GREY_BG_COLOR: string = "#E5E5E5";
+export const MAIN_BG_COLOR: string = "#fafafc";
```
BCerki, line 6: Love this refactor. I'm not sure about the naming of LINKS and MAIN_BG_COLOR though--links is used on things besides links, and bg-color is used in the tasklist not just main. Maybe there are more accurate names? (Unless these are the BC design system names, in which case disregard.)

app/lib/theme/widgets/ContextualHelp.tsx
```diff
@@ -0,0 +1,89 @@
...
+ * @param {IconDefinition} icon - The icon to display in the tooltip; defaults to faExclamationCircle
+ * @param {string} placement - The placement of the tooltip; defaults to "top"
+ * @param {string} label - The label of the tooltip to be used as the aria-labelledby attribute; this gets passed by the parent component(FieldLabel)
+ * @returns {ReactElement} - The tooltip component
+ * To use this component, simply add "ui:tooltip" object to the uiSchema of the field you want to add a tooltip to.
```
BCerki, line 13: This component will also be used on the diffs page, which doesn't use rjsf. It looks to me like we can use it like any react component (`<ContextualHelp />`), but want to check if you had any concerns with it being used that way

app/lib/theme/widgets/FieldLabel.tsx
```diff
@@ -17,16 +19,36 @@ const FieldLabel: React.FC<Props> = ({
     return null;
   }
 
+  // Show the tooltip if the uiSchema has a tooltip object
+  const ContextualHelpWidget = () => {
+    const tooltipObj = uiSchema?.["ui:tooltip"];
+    if (
+      !tooltipObj ||
+      typeof tooltipObj !== "object" ||
+      (typeof tooltipObj === "object" && Array.isArray(tooltipObj))
```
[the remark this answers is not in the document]
Sepehr-Sobhani: Just as a sanity check, we don't have to have this check since we already get front-end errors if we don't pass proper params.

app/lib/theme/widgets/ContextualHelp.tsx
```diff
@@ -0,0 +1,89 @@
...
+    | "right-start"
+    | "top-end"
+    | "top-start";
+  label: string;
+}
+
+const ContextualHelp: React.FC<ContextualHelpProps> = ({
+  title,
+  icon = faExclamationCircle,
+  placement = "bottom",
```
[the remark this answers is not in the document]
Sepehr-Sobhani: I think the same, just wanted to make it more configurable in case we wanted to use it anywhere else.

app/lib/theme/colors.ts
Sepehr-Sobhani: The link is coming from BC GOV design system: https://developer.gov.bc.ca/Design-System/Colour-Palette
But for the other one I really didn't have a better name and I'm terrible at naming variables :D please suggest some!

app/lib/theme/widgets/ContextualHelp.tsx
```diff
@@ -0,0 +1,89 @@
...
+  label,
+}) => {
+  // generate a unique id for the tooltip based on the label
+  const tooltipId = `${label.replace(/\s+/g, "-").toLowerCase()}-tooltip`;
+  return (
+    <>
+      <Tooltip
+        title={<span dangerouslySetInnerHTML={{ __html: title }} />}
```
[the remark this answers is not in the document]
Sepehr-Sobhani: I tried it but didn't work since we want to pass HTML code as the text. Also, I think it's safe to use `dangerouslySetInnerHTML` here since we are not getting user input here and simply passing the text from the UI schema to the client side.

app/lib/theme/widgets/ContextualHelp.tsx
Sepehr-Sobhani: If I understood your point correctly, We have this tooltip on the diffs page too:
Screenshot 2023-05-29 at 4 26 06 PM

app/tests/unit/lib/theme/widgets/ContextualHelp.test.tsx
Sepehr-Sobhani: The icon has an `img` role but when we want to get it using `getByRole` it doesn't find an accessible element with that role, but if we pass `hidden: true` since the SVG has the `aria-hidden=true` by default, we can select that element for testing purposes.
Also why `aria-hidden=true` on `img` role?: https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion

app/lib/theme/widgets/ContextualHelp.tsx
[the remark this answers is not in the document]
BCerki: I read that thing about it being safe if you never get user input, but I think I'd still like a second opinion on this one since it's a potential security thing. @mikevespi or @dleard, could you please weigh in?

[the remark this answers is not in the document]
Sepehr-Sobhani: I would love to change this if there's a workaround to this. I've found a way to parse HTML as text by adding the below package to the code but I'm not sure if it's really necessary:
https://www.npmjs.com/package/html-react-parser

app/lib/theme/colors.ts
mikevespi: It is generally a good idea to listen to Chris Coyier about css ^

app/tests/unit/lib/theme/widgets/ContextualHelp.test.tsx
```diff
@@ -0,0 +1,37 @@
+import { render, screen } from "@testing-library/react";
+import { faExclamationCircle } from "@fortawesome/free-solid-svg-icons";
+import ContextualHelp from "lib/theme/widgets/ContextualHelp";
+
+describe("ContextualHelp component", () => {
+  const textMock = "Tooltip Text";
```
BCerki, line 6: I'd suggest adding html to the text mock and checking that things look right after it gets parsed. There could also be another test with malformed html to see if the parser handles that well

Sepehr-Sobhani: That's a good point I've added an extra test for HTML text, Thanks 👍 I think the second test is not needed since we don't want to test the package itself.

BCerki (approved):
🪂

Issue #1562, opened 2023-03-21
Project Table Filter Field Inconsistencies


tmastrom:
Describe the Bug:

In the Projects table on the Projects page, the Primary Project Managers filter is missing the arrow indicating the field is a dropdown. The dropdown should look like the status dropdown (see screenshots 1 and 2).

Placeholder text in dropdown filters is serif font whereas other filters use a sans serif font (see screenshot 3).

Probability: 5

Effect: 2 (UX is slightly unclear)

Steps to reproduce the behaviour:
Go to the Projects page

Screenshots:

![Screenshot 2023-03-21 at 9.54.49 AM.jpg](https://images.zenhubusercontent.com/62f2b9c7cb78a05765516ca0/5a83bd11-083a-42ae-ba76-be9d68315c5d)

![Screenshot 2023-03-21 at 9.54.34 AM.jpg](https://images.zenhubusercontent.com/62f2b9c7cb78a05765516ca0/151b58db-c7ac-4e01-b6d5-0990748d2e7b)

![Screenshot 2023-03-21 at 9.57.56 AM.jpg](https://images.zenhubusercontent.com/62f2b9c7cb78a05765516ca0/7664236a-6db9-458c-aabb-6f3f43409f99)

Additional information:

- OS: MacOS Ventura
- Browser: chrome

pbastia:
Thanks @sam-warren ! The font looks good, did we decide not to implement the sorting arrows on that field?

pbastia:
I see. Thank you @sam-warren !

PR #1693, opened 2023-05-25
Primary Project Manager Table Filter Fix

sam-warren:
Card: https://github.com/bcgov/cas-cif/issues/1562

### DIFF
--- a/app/components/Table/Filters/SearchableDropdownFilter.tsx
+++ b/app/components/Table/Filters/SearchableDropdownFilter.tsx
@@ -67,7 +67,6 @@ export default class SearchableDropdownFilter extends TableFilter<string> {
                   ...params.InputProps,
                   disableUnderline: true,
                   style: {
-                    fontFamily: "BC Sans",
                     padding: "0",
                     fontSize: "14.4px",
                   },
--- a/app/pages/cif/projects.tsx
+++ b/app/pages/cif/projects.tsx
@@ -118,7 +118,7 @@ export function Projects({ preloadedQuery }: RelayProps<{}, projectsQuery>) {
         "Primary Project Managers",
         "primaryProjectManager",
         allCifUsers.edges.map((e) => e.node.fullName),
-        { allowFreeFormInput: true, sortable: false }
+        { allowFreeFormInput: false, sortable: false }
       ),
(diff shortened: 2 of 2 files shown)


app/pages/cif/projects.tsx
```diff
@@ -118,7 +118,7 @@ export function Projects({ preloadedQuery }: RelayProps<{}, projectsQuery>) {
         "Primary Project Managers",
         "primaryProjectManager",
         allCifUsers.edges.map((e) => e.node.fullName),
-        { allowFreeFormInput: true, sortable: false }
+        { allowFreeFormInput: false, sortable: false }
```
Sepehr-Sobhani, line 121: We can simply remove whole `allowFreeFormInput: false` as it's false by default.

app/components/Table/Filters/SearchableDropdownFilter.tsx
```diff
@@ -67,7 +67,6 @@ export default class SearchableDropdownFilter extends TableFilter<string> {
                   ...params.InputProps,
                   disableUnderline: true,
                   style: {
-                    fontFamily: "BC Sans",
```
Sepehr-Sobhani, line 70: Could you kindly provide some context on why this was removed?

Sepehr-Sobhani (changes requested):
Looks great 👍 just a minor change and a rebase is required!

app/pages/cif/projects.tsx
sam-warren: Good catch! Thank you

app/components/Table/Filters/SearchableDropdownFilter.tsx
sam-warren: The inclusion of this `fontFamily` property caused inconsistencies in the typeface used for the filters - part of the ticket's requirement was to have the affected filter fields' placeholder text consistent with the rest of the filter fields.

Sepehr-Sobhani: Oh sorry, I missed that part of the ticket's context. Thank you for the clarification.

sam-warren: All good! Thanks for your quick review :)

Sepehr-Sobhani (approved):

PR #1697, opened 2023-05-26
fix: showing `not added` when project status and operator not filled out

Sepehr-Sobhani:
[ZH CARD 805](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/805)

Note: Operator ID had the same issue which is not mentioned in the card itself but got fixed in this PR.


BCerki (approved):
Looks great! I found an unrelated bug while testing an operator with no bcRegistryId: https://github.com/bcgov/cas-cif/issues/1701

PR #1699, opened 2023-05-29
Cypress and dev data for 1569

BCerki:
PR for sharing code only, copy/paste files, don't merge

We tried testing the diffs two ways. 

1. programmatically adding the data that should trigger diffs: 

- in `psql` from `schema/data`:
\i e2e/dbReset.sql
 \i dev/001_cif_user.sql 
 \i dev/002_cif_operator.sql
 \i dev/003_cif_contact.sql
 \i dev/011_diffs.sql;

- go to the open amendment, change something that the general revision also changed, see no diff

2. Cypress tests: `diffs.cy.js` has three tests. Comment them in and out depending on what you want to test. The diffs don't show. Note (010 and 011 can't be used as part of `make drop_db make test_db`. If you need to use it, comment out everything in the 010 and 011, reset the database, then comment back in before running tests)


BCerki:
Just for sharing code, closing

PR #1705, opened 2023-05-31
Fix: fix default commit handler

Sepehr-Sobhani:
[ZH CARD 1702](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1702)


schema/deploy/functions/handle_default_form_change_commit.sql
```diff
@@ -38,12 +39,31 @@ begin
...
+  We don't want this to happen when operation is create since we need to use the default values provided by the table
+  **/
+  if fc.operation = 'update' then
+    for table_column_name in (
+        select column_name::text
+        from information_schema.columns
+        where table_schema = fc.form_data_schema_name
+          and table_name = fc.form_data_table_name
+          and (select cif_private.snake_to_camel_case(column_name)) not in (select key from jsonb_each(json_data))
+          and column_name::text not in ('id', 'created_at', 'updated_at', 'archived_at')
```
BCerki, line 60: I think we also don't want to wipe `created_by`, `updated_by`, `archived_by`. I'm trying to think if there's a way to continuously check for columns like this. If we add a column to a table in the future but don't remember to update this function, we'll unintentionally delete data

schema/deploy/functions/handle_default_form_change_commit.sql
```diff
@@ -13,6 +13,7 @@ declare
   record_id int;
   json_key text;
   json_data jsonb;
+  table_column_name text;
```
BCerki, line 16: I think the custom form change commit handlers might also have this bug

schema/test/unit/functions/handle_default_form_change_commit_test.sql
```diff
@@ -171,6 +170,32 @@ select results_eq(
   'All proper values were inserted when form_data includes a key that does not correspond to a column in the target table'
 );
 
+-- function updates the table with null values when the form_data does not include a key that corresponds to a column in the target table
+truncate mock_schema.mock_table restart identity;
+insert into mock_schema.mock_table (text_col, int_col, bool_col, required_col, defaulted_col)
```
BCerki, line 175: I think we additionally need tests to make sure it doesn't wipe anything it's not supposed to, e.g. `archived_at`.

BCerki (commented):
I haven't totally thought this through yet, but another way to fix this might be to adjust how we handle `newFormData`. If instead of deleting the whole key-value pair, we changed the value to `null` but kept the key, then the commit handler might put null in for us. (Advantages: don't have to worry about accidentally wiping data, disadvantages: side effects?).

I'm not sure how that key-value deletion happens. In the screenshot below, the first `console.log` is the form data before I touched anything. The next two logs happened after I deleted BC Registry ID, and there's a moment where it's undefined before it disappears entirely:
image

schema/deploy/functions/handle_default_form_change_commit.sql
```diff
@@ -38,12 +39,31 @@ begin
...
       json_data = json_data - json_key;
     end if;
   end loop;
 
+  /**
+  we need to add null values for any columns that are not in the form_data object
+  this is necessary because we may remove an optional field from the form but we don't update the data in the table because it is not in the form_data object
+  We don't want this to happen when operation is create since we need to use the default values provided by the table
+  **/
+  if fc.operation = 'update' then
```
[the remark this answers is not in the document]
Sepehr-Sobhani: Yes, but we want to manipulate `json_data` before generating `keys` on line 68.

schema/deploy/functions/handle_default_form_change_commit.sql
Sepehr-Sobhani: Fortunately, those commit handlers act differently. They look for specific fields in the form data and if they can't find it, set the value as null. 
This could also be bad I think, if we add a new field to those forms, we need to update its commit handler too.

Sepehr-Sobhani:
> I haven't totally thought this through yet, but another way to fix this might be to adjust how we handle `newFormData`. If instead of deleting the whole key-value pair, we changed the value to `null` but kept the key, then the commit handler might put null in for us. (Advantages: don't have to worry about accidentally wiping data, disadvantages: side effects?).
> 
> I'm not sure how that key-value deletion happens. In the screenshot below, the first `console.log` is the form data before I touched anything. The next two logs happened after I deleted BC Registry ID, and there's a moment where it's undefined before it disappears entirely: image

This is a valid point and I need to investigate this option more since It can have unintentional side-effects.

dleard:
@Sepehr-Sobhani This definitely looks more complicated than just checking against the set columns for a table can handle. Exploring how new_form_data is updated is definitely a good place to look. We _might_ be able to do something with the json schemas, but that may not work since not all schemas are being pulled from the database yet.

Sepehr-Sobhani:
Going to create a new clean branch for this issue with the new approach.

Issue #1616, opened 2023-04-20
As a CIF user, I want to see the differences on Milestone calculated fields since the last revision, so that I can focus on what did change


Sepehr-Sobhani:
#### Description:

Calculated fields are handled by computed columns, which our current diff logic can't display. For example, diff for calculated gross, net and holdback amounts for a milestone isn't shown on the summary page (page with red/black strikeout). This is likely because the calculated values are not actually part of the schema. 

This card is about implementing diffs in the Milestone form. The calculated values on the Milestone form are:
- Gross Payment Amount This Milestone
- Net Payment Amount This Milestone
- Holdback Amount This Milestone

#### Acceptance Criteria:

Given I have added enough values to complete the calculation for the first time
When I view the summary page
Then the diff displays the newly calculated value

Given I have changed values that affect the calculated value
When I view the summary page
Then the diff displays changed values properly for calculated values

Given I have removed enough values to prevent the calculation
When I view the summary page
Then the diff displays removed calculated values properly

#### Development Checklist:

- [ ] Any front-end calculated values are refactored into computed_columns:
    - N/A for Milestone form
-  [ ] Refactor calculations that calculate values based on info in `project_revision` to calculate based on `form_change` instead (this will also make the calculation dynamic)
    - [ ] project_revision_gross_payment_amount_this_milestone
    - [ ] project_revision_net_payment_amount_this_milestone
    - [ ] project_revision_holdback_amount_this_milestone
- [ ] Query the previously calculated values under `formChangeByPreviousFormChangeId`
- [ ] Pass the previously calculated value to `CustomDiffFields` via `formContext.oldData`
- [ ] Query the latest committed calculated value under `latestCommittedFormChangesFor`
- [ ] Pass the latest committed value to `CustomDiffFields` via `formContext.latestCommittedData`
- [ ] Pass the current calculated value to `CustomDiffFields` via `getSchemaAndDataIncludingCalculatedValues` 
- [ ] update summary form tests
- [ ] update any other mutations or tests as necessary
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [x] User story is included
- [x] User role and type are identified
- [x] Acceptance criteria are included
- [x] Wireframes are included (if required)(N/A)
- [ ] Design / Solution is accepted by Product Owner
- [x] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

[...]

PR #1706, opened 2023-05-31
feat: calculated milestone values diffing

gurjmatharu:
#1616

### DIFF
--- a/app/components/Form/ProjectMilestoneReportFormSummary.tsx
+++ b/app/components/Form/ProjectMilestoneReportFormSummary.tsx
@@ -194,7 +194,6 @@ const ProjectMilestoneReportFormSummary: React.FC<Props> = ({
       const clonedMilestoneFormDiffObject = JSON.parse(
         JSON.stringify(milestoneFormDiffObject)
       );
-      console.log("GURJ", clonedMilestoneFormDiffObject);
       Object.entries(titleMap).forEach(([key, title]) => {
         if (
           clonedMilestoneFormDiffObject.formData[key] !== null &&
--- a/app/components/Form/ProjectMilestoneReportFormSummary.tsx
+++ b/app/components/Form/ProjectMilestoneReportFormSummary.tsx
@@ -49,9 +49,6 @@ const ProjectMilestoneReportFormSummary: React.FC<Props> = ({
               formDataRecordId
               isPristine
               newFormData
-              calculatedHoldbackAmountThisMilestone
-              calculatedGrossAmountThisMilestone
-              calculatedNetAmountThisMilestone
               operation
               formChangeByPreviousFormChangeId {
                 newFormData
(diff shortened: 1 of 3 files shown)


BCerki (commented):
A few overall notes:
- I didn't fully understand how the milestone form changes work; if that's awkward to go through on github we could pair
- I think we could use assertions for calculated fields in the subsequent revision cypress test as well as the create one
- The ticket mentioned refactoring some computed columns and I don't see that work in this PR

gurjmatharu:
> * I think we could use assertions for calculated fields in the subsequent revision cypress test as well as the create one
I think the unit tests should take care of this. I've added assertions to the create project revision e2e test as well.

app/components/Form/ProjectMilestoneReportFormSummary.tsx
```diff
@@ -105,9 +105,31 @@ const ProjectMilestoneReportFormSummary: React.FC<Props> = ({
       // Set the formSchema and formData based on showing the diff or not
       const milestoneFormDiffObject = renderDiff
         ? getMilestoneFilteredSchema(
-            milestoneReport.formByJsonSchemaName.jsonSchema
-              .schema as JSONSchema7,
-            milestoneReport
+            {
+              ...(milestoneReport.formByJsonSchemaName.jsonSchema
+                .schema as JSONSchema7),
+              properties: {
```
BCerki, line 111: If you change the milestone type to Reporting, which doesn't have expenses, then the adjusted values still show up in the diffs: https://www.loom.com/share/8a62e10419a14496b6a7dcd67ffdf8fa. Do you think that was introduced by this cloning strategy? Also possible old things have always been hanging out in the new form data and we just didn't know until know because we never displayed diffs

BCerki (commented):
Looking good!

app/components/Form/ProjectMilestoneReportFormSummary.tsx
gurjmatharu: https://github.com/bcgov/cas-cif/issues/1745

app/tests/unit/components/Form/ProjectMilestoneReportFormSummary.test.tsx
```diff
@@ -141,5 +148,258 @@ describe("The Project Milestone Report Form Summary", () => {
...
+          },
+        ],
+      },
+    };
+    return result;
+  },
+};
+
+describe("Project Milestone Summary with diffs, without latestCommittedData", () => {
+  it("Displays diffs of the data fields that were updated and the old values", () => {
```
BCerki, line 353: This seems to be testing the same thing as `it("Displays diffs of the the data fields that were updated"`?  Do we need this test?

app/tests/unit/components/Form/ProjectMilestoneReportFormSummary.test.tsx
```diff
@@ -141,5 +148,258 @@ describe("The Project Milestone Report Form Summary", () => {
...
+        testQuery: testQuery,
+        compiledQuery: compiledProjectMilestoneReportFormSummaryQuery,
+        getPropsFromTestQuery: (data) => ({
+          query: data.query,
+          projectRevision: data.query.projectRevision,
+        }),
+        defaultQueryResolver: modifiedMockQueryPayload,
+        defaultQueryVariables: {},
+        defaultComponentProps: defaultComponentProps,
+      });
```
BCerki, line 251: Do we need a new describe block and component testing helper here? I think it could work to `loadQuery({...new mock here...})

app/tests/unit/components/Form/ProjectMilestoneReportFormSummary.test.tsx
gurjmatharu: I just added it for more coverage, the first one shows "Not Entered" this one shows actual diffs.  Happy to remove it.

app/tests/unit/components/Form/ProjectMilestoneReportFormSummary.test.tsx
gurjmatharu: I thought it would be easier to read, we don't explicitly need it.

JoshLarouche (commented):
All the code looks good to me. One thing I noticed is that when you are submitting a report without enough values to calculate the fields, or removing enough values so the fields can't be calculated, It's maybe a ambiguous what is happening with these values where I think if the summary page showed a message for each field saying something like "not enough values were entered to calculate/recalculate this field" it would be more clear to the general user. But this is more of a design question, I'd confirm with Nan if it's good as is and if so, happy to give my approval.

Sepehr-Sobhani (changes requested):
LGTM, just a couple of minor changes are needed.

Sepehr-Sobhani (commented):
I think it would be better to change `fix` to `chore` on this commit. the whole issue is a feature and having a `fix` in the changelog for this issue would be confusing.

Sepehr-Sobhani (approved):

Issue #1698, opened 2023-05-26
Performance Milestone Holdback Percentage does not allow for decimal places in the form input


dleard:
The Performance Milestone Holdback Percentage does not allow the user to enter any decimal points in the input

Probability (how likely the bug is to happen, scored from 1-5): 4

Effect (how bad the bug is when it does happen, scored from 1-5): 4

Steps to reproduce the behaviour:
Attempt to add decimal points to the Performance Milestone Holdback Percentage field. They are not accepted.

dleard:
@nanyangpro just want to confirm that this is not expected behaviour

Sepehr-Sobhani:
Possible solution:
adding `numberOfDecimalPlaces: 2` to `holdbackPercentage` object in `fundingParameterEPUiSchema` file

nanyangpro:
Thanks for your question, @dleard. I've seen 10% and 20% entered as the Performance Milestone Holdback Percentage in the past. However, you are correct, decimal places should be allowed for this field, users shouldn't be blocked to do so.

pbastia:
I can enter decimal places! Thanks a lot @sam-warren !

PR #1713, opened 2023-06-05
fix: holdback percentage not allowing decimals

sam-warren:
Card: https://github.com/bcgov/cas-cif/issues/1698


app/data/jsonSchemaForm/fundingParameterEPUiSchema.ts
```diff
@@ -34,6 +34,7 @@ export const fundingParameterEPUiSchema = {
   holdbackPercentage: {
     "ui:widget": "NumberWidget",
     isPercentage: true,
+    numberOfDecimalPlaces: 2,
```
BCerki, line 37: Is it confirmed that it should be 2 places?

BCerki (commented):
Nice! I'd suggest adding decimals to the mock data in the jest funding form test and summary tests, and one of the cypress tests, to make sure this displays as expected.

app/data/jsonSchemaForm/fundingParameterEPUiSchema.ts
sam-warren: Added two decimal places per the comment by @Sepehr-Sobhani on the ticket - @nanyangpro can you confirm what the expected number of decimal places are?

sam-warren: Update: Confirmed with Nan that 2 decimal places is acceptable.

BCerki (approved):

PR #1716, opened 2023-06-08
1361 Chore: refactor emissions intensity schema

BCerki:
card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1361

A couple notes:
- There's a previously existing bug in the calculated value widget: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/736. I was able to improve the display a little as part of required work for this ticket (removed 0/0/Not added), but bug remains
- Turned up another bug: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1740


schema/data/prod/json_schema/emission_intensity_report.json
```diff
@@ -62,7 +61,7 @@
         "title": "GHG Emission Intensity Performance",
         "type": "number"
       },
-      "calculatedPaymentPercentage": {
+      "paymentPercentage": {
```
BCerki, line 64: As far as I can tell, `paymentPercentage` and `actualPerformanceMilestoneAmount` aren't saved in any tables

app/components/Form/ProjectQuarterlyReportForm.tsx
```diff
@@ -106,6 +107,8 @@ const ProjectQuarterlyReportForm: React.FC<Props> = (props) => {
     props.projectRevision
   );
 
+  console.log("rev", projectRevision);
+
```
BCerki, line 111:
```suggestion

```

app/tests/unit/components/Form/ProjectEmissionIntensityReportFormSummary.test.tsx
```diff
@@ -235,6 +211,7 @@ describe("the emission intensity report form component", () => {
     };
     componentTestingHelper.loadQuery(customPayload);
     componentTestingHelper.renderComponent();
+    screen.logTestingPlaygroundURL();
```
BCerki, line 214:
```suggestion

```

schema/deploy/functions/handle_emission_intensity_form_change_commit.sql
```diff
@@ -0,0 +1,116 @@
...
+      ($1.new_form_data->>'targetEmissionIntensity')::numeric,
+      ($1.new_form_data->>'postProjectEmissionIntensity')::numeric,
+      ($1.new_form_data->>'totalLifetimeEmissionReduction')::numeric,
+      ($1.new_form_data->>'adjustedEmissionsIntensityPerformance')::numeric,
+      ($1.new_form_data->>'holdbackAmountToDate')::numeric,
+      ($1.new_form_data->>'dateSentToCsnr')::timestamptz
+    );
+
+
+    update cif.form_change set form_data_record_id = reporting_requirement_record_id where id = $1.id;
```
Sepehr-Sobhani, line 71: unrelated question: why do we get the value of `reporting_requirement_record_id` from the form data on line 22 and then set the same value here (line 71)?

Sepehr-Sobhani (changes requested):
Looks great! just a couple of minor changes are required. Also, I wasn't able to see the emission intensity data when I create a new revision:
Screenshot 2023-06-26 at 3 15 21 PM

Still, there are some files I need to go through but this was my major observation.

app/lib/theme/GroupedObjectFieldTemplateWrapper.tsx
```diff
@@ -0,0 +1,36 @@
...
+              key={group.title}
+              {...childProps}
+              title={group.title}
+            />
+          </>
+        );
+      })}
+    </>
+  );
+};
```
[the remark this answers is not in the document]
BCerki: How do I see the key prop error?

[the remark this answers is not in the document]
Sepehr-Sobhani: I think we can use a `div` or `Fragment` instead of `<>` and put the `key` there.

app/tests/unit/components/Form/ProjectQuarterlyReportForm.test.tsx
```diff
@@ -248,4 +248,62 @@ describe("The ProjectQuarterlyReportForm", () => {
       },
     });
   });
+
+  it("renders correct contract start date and TEIMP end date", () => {
```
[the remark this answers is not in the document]
BCerki: I added it because at one point in the refactor, I broke the start/end dates and none of the existing tests failed.  I can't remember exactly what I did, but I think I had the dates coming through as undefined (maybe a misplaced snake or camel case). However, now that you mention it, this test still uses dummy dates, so I don't think it's actually testing what I wanted. I'll look into adding an assertion to the cypress test. Thanks!

schema/deploy/functions/handle_emission_intensity_form_change_commit.sql
BCerki: Because the milestone form change commit does that and I copied it 😛  (I think because it gets updated line 41)

BCerki:
> Looks great! just a couple of minor changes are required. Also, I wasn't able to see the emission intensity data when I create a new revision: Screenshot 2023-06-26 at 3 15 21 PM
> 
> Still, there are some files I need to go through but this was my major observation.

I wasn't able to reproduce this with the following flow: https://www.loom.com/share/7e9c241840ff44bbbf0fb814442e76e2?sid=e9e10e23-837e-427a-9f6e-3c5a537a0bf1. Did you get there some other way?

app/lib/theme/GroupedObjectFieldTemplateWrapper.tsx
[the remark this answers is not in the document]
BCerki: Oh, I was thinking it might be on the children. I can remove the <> entirely, maybe that will help. Where do you see the error?

app/lib/theme/schemaFilteringHelpers.ts
```diff
@@ -12,18 +12,27 @@ export const getSchemaAndDataIncludingCalculatedValues = (
...
+    properties: {
+      ...JSON.parse(JSON.stringify(formSchema)).properties,
+      ...additionalSchemaProperties,
+    },
   };
 
   const newDataObject = {};
 
   for (const key of Object.keys(filteredSchema.properties)) {
+    // null new data with undefined old data occurs when an optional form (e.g. budgets) is created but not filled
```
BCerki, line 24: This is what causes the bug where we see labels but no diffs: 
Screenshot from 2023-06-27 15-18-19
We could also handle this in `CustomDiffs` and I can make a tech debt card if people think that would be a better approach. The milestone form uses a custom schema filtering helper, so it might have the same bug

Sepehr-Sobhani: That makes sense to me. I think I faced that issue before when I was implementing the schema filtering function for the Emissions Intensity Report.

app/lib/theme/GroupedObjectFieldTemplateWrapper.tsx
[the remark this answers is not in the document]
Sepehr-Sobhani: I think that would fix the issue. I got that on the console when I was working with the EI form.

app/components/Form/ProjectEmissionIntensityReportForm.tsx
```diff
@@ -82,61 +81,61 @@ export const createEmissionIntensityReportUiSchema = (
...
-  uiSchemaCopy["ui:FieldTemplate"] = ReadOnlyEmptyFieldTempalte;
-  uiSchemaCopy["ui:ObjectFieldTemplate"] = EmptyObjectFieldTemplate;
-
   return uiSchemaCopy;
 };
 
 const ProjectEmissionsIntensityReport: React.FC<Props> = (props) => {
   const formRefs: MutableRefObject<{}> = useRef({});
 
+  const { eiFormBySlug } = useFragment(
```
BCerki, line 128: If I get the schema via the formChanges instead, sometimes I get the emission_intensity schema, but sometimes I get the reporting_requirement schema (this is what caused some of the fields to disappear; when I got the reporting_requirement schema instead). Not sure why this was happening, but querying for the schema we want should fix it

BCerki:
Continuing in new PR after hotfix, closing

PR #1717, opened 2023-06-08
feat: 1663 remove read only revision type

Sepehr-Sobhani:
[ZH CARD 1663](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1663)

### DIFF
--- a/app/pages/cif/project-revision/[projectRevision]/edit.tsx
+++ b/app/pages/cif/project-revision/[projectRevision]/edit.tsx
@@ -53,15 +53,6 @@ export const EditProjectRevisionQuery = graphql`
         }
       }
     }
-    allRevisionTypes {
-      # type is passed to the helper function that builds the schema
-      # eslint-disable-next-line relay/unused-fields
-      edges {
-        node {
-          type
-        }
-      }
-    }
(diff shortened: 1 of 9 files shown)


app/pages/cif/project-revision/[projectRevision]/edit.tsx
```diff
@@ -83,8 +80,7 @@ export function ProjectRevisionEdit({
 }: RelayProps<{}, editProjectRevisionQuery>) {
   const router = useRouter();
   const query = usePreloadedQuery(EditProjectRevisionQuery, preloadedQuery);
-  const { session, projectRevision, allRevisionTypes, allRevisionStatuses } =
```
Comment (JoshLarouche) [inline app/pages/cif/project-revision/[projectRevision]/edit.tsx:86]: allRevisionTypes seems to be being removed everywhere else, shouldn't it be removed from the query above on line 56 as well?

Comment (Sepehr-Sobhani) [reply app/pages/cif/project-revision/[projectRevision]/edit.tsx:86]: Good call, It's a leftover! Thanks Josh.

JoshLarouche (approved):
Looks great! Nice work!

PR #1719, opened 2023-06-10
Fix: add null and undefined values to optional fields in json schemas

Sepehr-Sobhani:
[ZH CARD 1702](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1702)

NOTES:
- If a form consumes its own commit handler we don't need to add default values (like milestone form)
- On the forms with default null values, we get an extra mutation call when creating/updating/discarding that form
- We don't save some fields in tables: `holdback amount` in milestone form and `payment note` in project summary report form (Need a bug card?!)
- One observation: project summary report doesn’t show the latest value when creating a new revision [Bug Card](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1802)
- We need to discuss how to remove/archive records in project contact and project manager forms as they both have FK relations and are not nullable in DB


JoshLarouche (approved):
looks good to me!

schema/data/prod/json_schema/emission_intensity_report.json
```diff
@@ -4,7 +4,6 @@
     "type": "object",
     "title": "Emissions Intensity Report",
     "required": [
-      "reportDueDate",
       "measurementPeriodStartDate",
       "measurementPeriodEndDate",
       "emissionFunctionalUnit",
```
dleard, line 9: why was 'reportDueDate' removed as a requirment here?

schema/data/prod/json_schema/contact.json
```diff
@@ -2,7 +2,7 @@
   "schema": {
     "$schema": "http://json-schema.org/draft-07/schema",
     "type": "object",
-    "required": ["givenName", "familyName", "email", "phone"],
+    "required": ["givenName", "familyName", "email"],
     "properties": {
       "givenName": {
         "type": "string",
```
dleard, line 8: Why was 'phone' removed as a requirement here?

dleard (commented):
Awesome work @Sepehr-Sobhani! This was definitely a complicated fix to implement.

I manually tested all the optional fields by creating a project with all fields filled, committing it and then removing all optional fields. They were all properly removed from the table data :100: 

I just have a couple questions around some removal of required fields.

schema/data/prod/json_schema/emission_intensity_report.json
Sepehr-Sobhani: At first, I removed it to align with the `emission_intensity_reporting_requirement.json` file. However, I decided not to bring it back as we no longer utilize the `emission_intensity_report.json` file, which has been replaced by `emission_intensity.json`.

schema/data/prod/json_schema/contact.json
Sepehr-Sobhani: Currently, the `contactSchema.ts` is being used as the schema for the contact form instead of `contact.json`. To ensure consistency, I made them identical. Upon reviewing the PROD data, I noticed multiple null values for the phone field. I believe that it was not necessary from the start. However, please correct me if my assumption is incorrect.

JoshLarouche (approved):

dleard (approved):

PR #1721, opened 2023-06-11
refactor: custom diff fields

gurjmatharu:
#1569 
- this refactor uses a stack implementation
- explicitly checks for undefined/null
- decimal values check added (note: some decimal values still seem incorrect from the different forms: tiemp)
- zero's are tested comprehensively

### DIFF
--- a/app/tests/unit/lib/theme/CustomDiffFields.test.tsx
+++ b/app/tests/unit/lib/theme/CustomDiffFields.test.tsx
@@ -135,6 +135,7 @@ describe("The Object Field Template", () => {
 
     expect(componentUnderTest.getByText("stringTest NEW")).toBeInTheDocument();
     expect(componentUnderTest.getByText("stringTest OLD")).toBeInTheDocument();
+    // unicorn -- shouldn't see latest committed when old=latest
     expect(
       componentUnderTest.queryByText(/^(.*?)stringTest LAST COMMITTED/)
     ).not.toBeInTheDocument();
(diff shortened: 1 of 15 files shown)


app/tests/unit/lib/theme/CustomDiffFields.test.tsx
```diff
@@ -127,73 +125,85 @@ describe("The Object Field Template", () => {
...
       componentUnderTest.getByText("I replaced the OLD ID")
     ).toBeInTheDocument();
+    expect(
+      componentUnderTest.queryByText(/I am the last committed value/i)
+    ).not.toBeInTheDocument();
   });
-
-  it("shows data has been added when there is newData, the operation is 'CREATE' & there is no old data", () => {
+  // 101
+  it("shows data has been removed when there is old data and latest committed data but no new data", () => {
```
BCerki, line 155: this probably needs two cases, one for if old === latest and one for if it doesn't

app/tests/unit/lib/theme/CustomDiffFields.test.tsx
```diff
@@ -77,7 +77,8 @@ const latestCommittedData = {
 };
 
 describe("The Object Field Template", () => {
```
BCerki, line 79: Test name could use an update

I think we're missing a few tests still:
- if there have been no changes to anything, we should see no diffs
- if we have old and new but not latest committed
- arrows show up correctly

BCerki (commented):
This version's a lot easier to follow, nice!

I'm going to hold off on manual testing for now; I think there are going to be some display errors.

dleard:
@gurjmatharu Looks like there are a couple bugs around the placement of the arrows.
In some places where there are old, latest and new changes, the arrows are not in between each value, but two of them are grouped together.

Screenshot from 2023-06-19 15-01-14

There's another similar issue with the emission intensity form where there are only two changes: old and new but there are two arrows being shown. It looks like maybe the units CO2e/tonnes is being parsed here as a value?

Screenshot from 2023-06-19 15-01-19

dleard (approved):
@gurjmatharu Nice work! This diff format is much cleaner to read.
A tech debt card will be created to add some more tests and explore some more edge cases around this feature. The UI and test coverage in this PR is better than the current implementation so no reason to hold this back.

PR #1725, opened 2023-06-14
1644 login redirect functionality

JoshLarouche:
[zenhub 1644](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1644)

### DIFF
--- a/app/components/Layout/Navigation.tsx
+++ b/app/components/Layout/Navigation.tsx
@@ -5,6 +5,9 @@ import LogoutForm from "components/Session/LogoutForm";
 import SubHeader from "./SubHeader";
 import { useRouter } from "next/router";
 
+import LoginForm from "components/Session/LoginForm";
+import useShowGrowthbookFeature from "lib/growthbookWrapper";
+
 interface Props {
   isLoggedIn?: boolean;
   alwaysShowSubheader?: boolean;
@@ -25,10 +28,20 @@ const Navigation: React.FC<Props> = ({
   userProfileComponent,
   links,
 }) => {
-  let rightSide = isLoggedIn && (
+  // Growthbook - external-operators
+  const showExternalOperatorsLogin =
+    useShowGrowthbookFeature("external-operators");
+
+  let rightSide = isLoggedIn ? (
     <>
       {userProfileComponent}
       <LogoutForm />
+    </>
+  ) : (
+    <>
+      {userProfileComponent}
+      <LoginForm />
+      {showExternalOperatorsLogin && <LoginForm isExternal={true} />}
     </>
   );
   const router = useRouter();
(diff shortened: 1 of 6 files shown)


app/components/Layout/Navigation.tsx
Sepehr-Sobhani, line 1: I think we can do a bit of refactoring on lines 37 through 54, something like:
```
  let rightSide = isLoggedIn ? (
    <>
      {userProfileComponent}
      <LogoutForm />
    </>
  ) : (
    router.pathname !== "/" && (
      <div
        style={{
          display: "flex",
          flexDirection: "row",
          gap: "1.25em",
        }}
      >
        <LoginForm />
        {showExternalOperatorsLogin && <LoginForm isExternal={true} />}
      </div>
    )
  );
```

app/components/Session/LoginForm.tsx
Sepehr-Sobhani, line 1: we can also simplify things here a little bit and have less code to maintain:
```
      <form id="login-buttons" action={loginURI} method="post">
        <Button
          type="submit"
          variant={
            router.pathname.includes("login-redirect")
              ? "secondary-inverse"
              : isExternal && "secondary"
          }
        >
          {isExternal ? "External User Login" : "Administrator Login"}
        </Button>
      </form>
```

Sepehr-Sobhani, line 1: Also, we need to tweak `loginURI` too, currently, this part of the AC is not in place:
`Then I am redirected back to the page I was trying to get to.`
I've played with the code and something like this could be a solution:
```
  let loginURI = "/login";

  if (router.query.redirectTo)
    loginURI += `?redirectTo=${encodeURIComponent(
      router.query.redirectTo as string
    )}`;
  else if (isExternal)
    loginURI += `?redirectTo=${encodeURIComponent(
      getExternalUserLandingPageRoute().pathname as string
    )}`;
```

app/tests/unit/pages/login-redirect.test.tsx
Sepehr-Sobhani, line 1: I believe it would be advantageous to conduct a test here on the redirect functionality for external users.

Sepehr-Sobhani (changes requested):
Great Job 👍 Just a couple of changes are required.

app/components/Layout/Navigation.tsx
JoshLarouche: Good idea! Changed!

app/components/Session/LoginForm.tsx
JoshLarouche: Good call! Changed!

JoshLarouche: Good catch I tested mostly with the admin login! Changed!

app/tests/unit/pages/login-redirect.test.tsx
JoshLarouche: After discussion with @Sepehr-Sobhani  it was determined that it does not make sense to test this for now until we add more external user functionality.

Sepehr-Sobhani (approved):

PR #1732, opened 2023-06-19
Chore: Tweak 1582 preview revision

BCerki:
Card #1582


BCerki:
Work finished in new branch

PR #1753, opened 2023-06-28
Feat: 1703 hard delete attachments

BCerki:
card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1703

The card is only about hard deleting attachments in the first revision. I think there may be some other cases in which we'd like to hard delete (e.g., deleting an attachment that was uploaded in the same form change) and made a tech debt ticket to investigate: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1776

Also, hard deleting attachments as part of discarding a project revision was split into another card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1798


app/server/middleware/attachmentDeleteRouter.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const response = await storageClient.bucket(bucketName).file(file).delete();
+
+    return res.sendStatus(response[0].statusCode);
+  } catch (error) {
+    next(error);
+  }
+};
+
+attachmentDeleteRouter.get("/delete/:attachmentId", handleDelete);
```
pbastia, line 64: Your router is expecting a `GET` request, but the request uses a `DELETE` method in `AttachmentTableRow.tsx`

```suggestion
attachmentDeleteRouter.delete("/delete/:attachmentId", handleDelete);
```

pbastia (changes requested):

BCerki:
Continuing in https://github.com/bcgov/cas-cif/pull/1817

PR #1754, opened 2023-06-28
Fix missing project attachment form changes

Sepehr-Sobhani:


schema/deploy/migrations/007_create_form_changes_for_existing_project_attachments.sql
```diff
@@ -0,0 +1,13 @@
+-- Deploy cif:migrations/007_create_form_changes_for_existing_project_attachments to pg
+-- requires: tables/form_change
+-- requires: tables/project_attachment
+
+begin;
+
+  alter table cif.form_change disable trigger _100_timestamps;
```
dleard, line 7: I think we will need to add the following triggers to be disabled as well:
- commit_form_change
- _set_previous_form_change_id
- _100_committed_changes_are_immutable

schema/deploy/functions/migration_create_form_changes_for_existing_project_attachments.sql
```diff
@@ -0,0 +1,67 @@
...
+          'projectId', record.project_id,
+          'attachmentId', record.attachment_id
+        ),
+        'committed',
+        record.created_by,
+        record.created_at,
+        record.updated_by,
+        record.updated_at
+      );
+    else
```
dleard, line 55: We can remove this `else`. It was just there for sanity check purposes when we were debugging.

schema/deploy/functions/migration_create_form_changes_for_existing_project_attachments.sql
```diff
@@ -0,0 +1,67 @@
...
+      raise notice 'I EXIST. PROJECT_ID: %, ATTACHMENT_ID %', record.project_id, record.attachment_id;
+    end if;
+  end loop;
+end;
```
dleard, line 59: After the first loop is done here, we'll need to add something to programatically update the form_change records for revisions that were created after this bug was introduced

dleard (changes requested):

dleard (approved):

PR #1771, opened 2023-07-06
Refactor: 1361 emissions intensity schema

BCerki:
card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1361

A couple notes:
- There's a previously existing bug in the calculated value widget: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/736. I was able to improve the display a little as part of required work for this ticket (removed 0/0/Not added), but bug remains
- Turned up another bug: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1740


dleard (changes requested):
I found a few things that caused the migration to break when deploying over a proper dataset.

schema/deploy/mutations/create_project_revision.sql
dleard, line 1: There is an issue in this file [here](https://github.com/bcgov/cas-cif/blob/db7b89ef4a3eded800f4d9f61a9980541a51f114/schema/deploy/mutations/create_project_revision.sql#L67) where the reportType of 'TEIMP' needs be be ignored like we're ignoring 'Project Summary Report'.
This is causing the json_schema_name to be set to 'reporting_requirement' when it should be 'emission_intensity', which in turn causes the data not to propagate to the emission intensity report table on commit and the diff to be missing data.

dleard (changes requested):

schema/data/prod/json_schema/emission_intensity.json
```diff
@@ -0,0 +1,80 @@
...
+      "measurementPeriodStartDate",
+      "measurementPeriodEndDate",
+      "emissionFunctionalUnit",
+      "baselineEmissionIntensity",
+      "targetEmissionIntensity"
+    ],
+    "properties": {
+      "reportDueDate": {
+        "type": "string",
+        "title": "Report Due Date"
```
BCerki, line 16: add `defaultValue: null` for all optional fields
type should `type: string[] || null` for optional fields

BCerki: to test--clear an optional field, commit, and go check db to make sure it's null in the db (if the previously committed value is still there, it didn't work)

BCerki: for date widgets, might not need to add defaultValue because the widget handles it, check this

app/data/jsonSchemaForm/emissionIntensityUiSchema.ts
```diff
@@ -0,0 +1,130 @@
...
+    "ui:widget": "DateWidget",
+    isDueDate: true,
+  },
+  submittedDate: {
+    "ui:widget": "DateWidget",
+    isReceivedDate: true,
+  },
+  comments: {
+    "ui:widget": "TextAreaWidget",
+  },
```
BCerki, line 99: I deleted these two and they're gone from the db: 
image

app/data/jsonSchemaForm/emissionIntensityUiSchema.ts
```diff
@@ -0,0 +1,130 @@
...
+  baselineEmissionIntensity: {
+    "ui:widget": "NumberWidget",
+    numberOfDecimalPlaces: 8,
+  },
+  targetEmissionIntensity: {
+    "ui:widget": "NumberWidget",
+    numberOfDecimalPlaces: 8,
+  },
+  postProjectEmissionIntensity: {
```
BCerki, line 43: deleted this and it's gone from db: 
image

dleard (approved):
:100: This was a big task. Nice work!
I've checked:
- The migration does not fail
- It creates the correct number of form_change records
- The form properly creates/updates EI report form_change records
- Table data is properly populated on commit
- Optional values are removed from the table data if removed on the form

This looks good to go, just needs a rebase

Issue #1692, opened 2023-05-25
As a CIF Administrator, I want to see help text of form fields, so that I understand what it is and how it's calulated


nanyangpro:
#### Description:

The rules for some calculated fields are getting too complex even for someone familiar with the CIF program. Tooltips are prepared for fields on the forms below to provide more clarity to users.

- Budgets, Expenses, & Payments
- Milestone Reports
- Emissions Intensity Report 

#### Acceptance Criteria:

Given I am a CIF administrator
When I hover over the tooltip icon next to a field listed in the [Tooltips for Calculated Fields](https://bcgov.sharepoint.com/:w:/t/00608-ScrumTeam/EfLRtSfmMwxJl5B8DGj2I2cBgFyPnX_bxFF8SZiIoZ30yA?e=e7z0T3) document
Then I see contextual help text for the field

#### Development Checklist:

- [x] Add `ui:tooltip` object to the UISchema of each of these forms and pass the related context to it
- [x] Add tests
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

Sepehr-Sobhani:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-cif/1692?planning-poker) with Zenhub @BCerki @dleard @gurjmatharu @JoshLarouche @mikevespi @sam-warren

BCerki:
[...]

PR #1774, opened 2023-07-06
Feat: 1692 tooltips

BCerki:
card #1692 

continued from https://github.com/bcgov/cas-cif/pull/1771/files

### DIFF
--- a/app/data/jsonSchemaForm/tooltipText.ts
+++ b/app/data/jsonSchemaForm/tooltipText.ts
@@ -1,0 +1,65 @@
+export const fundingTooltips = {
+  totalProjectValue:
+    "<div><ul><li>Total Project Value = Maximum Funding Amount + Proponent Cost + Additional Funding Amount (Source 1) + Additional Funding Amount (Source 2) + ... + Additional Funding amount (Source N).</li></ul></div>",
+
+  maxFundingAmount:
+    "<div><ul><li>The maximum payment amount that the proponent can receive from CIF.</li></ul></div>",
+
+  provinceSharePercentage:
+    "<div><ul><li>Typically, 50% at most.</li></ul></div>",
+
+  holdbackPercentage:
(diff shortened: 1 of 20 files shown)


app/data/jsonSchemaForm/tooltipText.ts
Sepehr-Sobhani, line 1: I love the idea of keeping all tooltip text in one place.

Sepehr-Sobhani (commented):
looks great 🤩 I'm just not sure about creating snapshots from form pages; we need to change them every time we change a single line in those form pages. what do you think?!

BCerki:
> looks great star_struck I'm just not sure about creating snapshots from form pages; we need to change them every time we change a single line in those form pages. what do you think?!

I'm not attached to the snapshot tests. I added them based on your comment in the old PR, `It would be nice to test the text of the tooltips in unit tests and make sure we won't change accidentally later. ` (jest assertions seemed like overkill for accidental text changes). Thoughts? Now that the tooltip text is separate from the ui schemas, I think it would be hard to accidentally change the text.

Sepehr-Sobhani:
> > looks great star_struck I'm just not sure about creating snapshots from form pages; we need to change them every time we change a single line in those form pages. what do you think?!
> 
> I'm not attached to the snapshot tests. I added them based on your comment in the old PR, `It would be nice to test the text of the tooltips in unit tests and make sure we won't change accidentally later. ` (jest assertions seemed like overkill for accidental text changes). Thoughts? Now that the tooltip text is separate from the ui schemas, I think it would be hard to accidentally change the text.

My main concern was inadvertently changing certain texts, but a single unit test would ensure that they are not altered by mistake. However, the new approach of keeping them in a separate file minimizes the risk of accidental changes. We are now good to proceed.

sam-warren (approved):
Changes look great :smiley:

PR #1775, opened 2023-07-06
Fix: 1743 false attention required status

dleard:


schema/test/unit/util_functions/get_form_status_test.sql
```diff
@@ -109,6 +109,27 @@ select results_eq(
...
+update cif.form_change set new_form_data='{}', change_status='staged'
+  where project_revision_id=1
+    and form_data_table_name='project_contact';
+
+/* test 7 */
+
+update cif.form_change set operation='archive', change_status='staged'
+  where project_revision_id=1
+    and validation_errors != '[]'
+    and form_data_table_name= 'project_contact';
```
BCerki, line 121: I had a little trouble following what the test data looked like at this point. I think it might be clearer to do something like test 5 and set everything including errors just before the assertion

schema/deploy/util_functions/get_form_status_001.sql
```diff
@@ -26,6 +26,7 @@ $function$
         )
```
BCerki, line 26: Got confused by the filename initially--this PR is a rework of a _001 file?

BCerki (commented):
Looks cool!

When I tried to undo changes,  I got stuck on Attention Required again: 
image

I also found a diffs bug: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1780

schema/test/unit/util_functions/get_form_status_test.sql
dleard: I'll try and make it clearer

schema/deploy/util_functions/get_form_status_001.sql
dleard: yeah, it looks like a _001 file got created at one point instead of a rework, so now I'm just reworking the new file.

dleard:
> Looks cool!
> 
> When I tried to undo changes, I got stuck on Attention Required again: https://www.loom.com/share/16cbcc812cfb45fd92896e535f520365?sid=30b7907a-6ef8-425c-bb54-9a59a3ad04fc
> 
> I also found a diffs bug: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1780

Hm, it appears that the undo changes preserves the validation errors when restored, which I suppose makes sense except in this specific scenario. I think this is a separate issue, so I'll create a new card to investigate it. I don't think it should block this fix since it's a much less likely and impactful bug than what this PR fixes.

I also wrote up a card that's semi-related to the milestone diffs bug yesterday during a chat with Nan, #1777

BCerki (approved):

PR #1781, opened 2023-07-07
1039 submit button funded option

gurjmatharu:


app/components/Form/ProjectFundingAgreementForm.tsx
Sepehr-Sobhani, line 1: Based on this article I usually try to avoid using `useEffect` when possible: https://react.dev/learn/you-might-not-need-an-effect

Here we might be able to do so by following these steps (part of these steps are also suggestions for a refactor):
- Remove `useEffect`
- Adding `const [isFunded, setIsFunded] = useState(null);` (Just for making the state more meaningful)
- Adding `setIsFunded("no");` to `handleDiscard` function
- Changing all `onClick` props on radio buttons to `onChange`
- Adding `setIsFunded("yes");` to radio buttons with Yes label and `setIsFunded("no");` to those with No label
- Adding `checked={isFunded === "yes"}` to radio buttons with Yes label and `checked={isFunded === "no"}` to those with No label
- Adding `onChange={() => setIsFunded("yes")}` to radio buttons with Yes label and `onChange={() => setIsFunded("yes")}` to those with No label
- Setting `disabled={!isFunded}` and `onClick={() => props.onSubmit()}` for submit button (This is what we do in other forms)
- Adding `setIsFunded("yes");` to `onCancel` prop of `DangerAlert`

This is just my suggestion, feel free to ignore :)

Sepehr-Sobhani (commented):
LGTM, just added a suggestion but feel free to ignore it and I will approve the PR.

app/components/Form/ProjectFundingAgreementForm.tsx
gurjmatharu: Do you think we can create a tech debt for this?

Sepehr-Sobhani: I don't believe it's worthwhile to create a tech debt card and invest time in explaining the tech debt. However, I already have the code ready and can share it with you if you're interested.

Sepehr-Sobhani: One little thing I forgot: since this is a bugfix it might be better to replace the `chore` on the first commit with `fix`

Sepehr-Sobhani (approved):
🤩

PR #1795, opened 2023-07-11
Fix: 1794 - adjusted holdback amount capitalization

dleard:


schema/test/unit/migrations/008_replace_bad_field_capitalization_test.sql
```diff
@@ -0,0 +1,170 @@
...
+        json_build_object(
+          'projectId', 1,
+          'reportDueDate', '2000-12-12 16:21:42.693489-07',
+          'submittedDate', '2000-12-12 16:21:42.693489-07',
+          'reportType', 'General Milestone',
+          'reportingRequirementIndex', 2,
+          'description', 'general milestone report description ' ,
+          'adjustedGrossAmount', 10,
+          'adjustedNetAmount', 9,
+          'adjustedHoldBackAmount', 1,
```
andrea-williams, line 56: have you tested this on dummy data that has an attribute called `adjustedHoldbackAmount` to confirm there are no adverse effects on existing data with the correct capitalization?

andrea-williams (approved):
Code LGTM (to the limited extent that I can review it given I don't know the app and haven't set things up to run on my local yet...)

schema/test/unit/migrations/008_replace_bad_field_capitalization_test.sql
dleard: There won't be any in the data, all current records have the bad capitalization. But you reminded me that I should probably add another test for idempotency (which will also cover the scenario you brought up)

PR #1815, opened 2023-07-20
Fix:1802 project summary bug

BCerki:
card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1802


schema/deploy/functions/handle_project_summary_report_form_change_commit.sql
```diff
@@ -6,18 +6,21 @@ begin;
...
+  -- If there is no change in the form data, return the form_change record and do not touch the associated table.
+  if (fc.new_form_data = '{}') then
+    return fc.form_data_record_id; -- can be null if creating with empty form data...problem?
+  end if;
+
+  if (fc.change_status = 'committed') then
+    raise exception 'Cannot commit form_change. It has already been committed.';
+  end if;
+
+  reporting_requirement_record_id := fc.form_data_record_id;
```
BCerki, line 22: This line was missing and that's what created the bug. This function returns `reporting_requirement_record_id`, so if we don't assign here, then the update and archive cases return null.

schema/test/unit/mutations/create_project_revision_test.sql
```diff
@@ -177,28 +241,75 @@ select is_empty(
...
+      "totalLifetimeEmissionReduction": 44.4224
+      }'::jsonb, 'emission_intensity'::varchar),
+    (1,3,'{
       "projectId":1,
       "reportType": "General Milestone",
       "reportingRequirementIndex": 1
       }'::jsonb, 'milestone'::varchar),
-    ('{
+      -- brianna change this later to 4 I think
+    (4,6,'{
```
BCerki, line 278: If the commit handler returns null, then the `form_data_record_id` is also null. So when `create_project_revision` makes a copy of the existing forms, it just copies the original one over and over again

schema/test/unit/mutations/create_project_revision_test.sql
```diff
@@ -177,28 +241,74 @@ select is_empty(
...
+      "targetEmissionIntensity": 23.2357,
+      "totalLifetimeEmissionReduction": 44.4224
+      }'::jsonb, 'emission_intensity'::varchar),
+    (1,3,'{
       "projectId":1,
       "reportType": "General Milestone",
       "reportingRequirementIndex": 1
       }'::jsonb, 'milestone'::varchar),
-    ('{
+    (4,6,'{
```
BCerki, line 277: If the commit handler returns null, then the `form_data_record_id` is also null. So when `create_project_revision` makes a copy of the existing forms, it just copies the original one over and over again instead of getting the latest one

schema/test/unit/mutations/create_project_revision_test.sql
```diff
@@ -242,5 +352,37 @@ select lives_ok(
...
+  form_data_table_name,
+  form_data_record_id ,
+  project_revision_id ,
+  change_status ,
+  json_schema_name,
+  validation_errors ,
+  previous_form_change_id from cif.form_change where json_schema_name='project_summary_report';
+select '======';
+
+
```
Sepehr-Sobhani, line 386: Leftover!

Sepehr-Sobhani (commented):
This looks great! However, I do have one concern. There are currently three projects in production that have an associated project summary report, but the values in the database are incorrect. To resolve this issue, I believe we should create a migration and commit the form changes related to those projects.
Project IDs:
- 202 (2021-RFP-2-SKNA-018)
- 224 (2021-RFP-2-PETR-130)
- 226 (2021-RFP-2-FTBC-144)

schema/deploy/functions/migration_rebuild_project_summary_report_history.sql
```diff
@@ -0,0 +1,26 @@
...
+        on fc1.previous_form_change_id=fc2.id
+    )
+
+    update cif.form_change set form_data_record_id=
+        (
+        select fdrid from records where prfid=id2 limit 1
+        )
+        where form_data_record_id is null
+        and change_status='committed'
+        and json_schema_name='project_summary_report';
```
dleard, line 22: We need to connect the form_change being updated with the fc1 form_change in the records query
```suggestion
    with records as (
        select fc1.id as id1,fc1.previous_form_change_id as prfid, fc2.form_data_record_id as fdrid, fc2.id as id2
        from cif.form_change as fc1
        join cif.form_change as fc2
        on fc1.previous_form_change_id=fc2.id
        and fc1.json_schema_name='project_summary_report'
    )

    update cif.form_change fc set form_data_record_id=
        (
        select fdrid from records where fc.id=id1
        )
        where form_data_record_id is null
        and change_status='committed'
        and json_schema_name='project_summary_report';
```

dleard (changes requested):
Looks good, just one suggestion to make and we need a migration to call the migration function. Make sure to disable these 3 triggers in the migration file:

_100_committed_changes_are_immutable
_set_previous_form_change_id,
_100_timestamps;

BCerki:
I think I fixed the bug about the summary report summary not showing in view. I found another related one that wasn't straightforward so I made a card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1844

app/components/Form/ProjectEmissionIntensityReportFormSummary.tsx
```diff
@@ -175,7 +175,10 @@ const ProjectEmissionsIntensityReportFormSummary: React.FC<Props> = ({
 
   if (isOnAmendmentsAndOtherRevisionsPage && allFormChangesPristine)
     return null;
-
+  console.log("emissions");
+  console.log("allFormChangesPristine", allFormChangesPristine);
+  console.log("viewOnly", viewOnly);
+  console.log("summaryReportingRequirement", summaryReportingRequirement);
```
Sepehr-Sobhani, line 181: Leftover!

app/components/Form/ProjectEmissionIntensityReportFormSummary.tsx
```diff
@@ -223,7 +226,7 @@ const ProjectEmissionsIntensityReportFormSummary: React.FC<Props> = ({
         },
       },
     };
-
+  console.log("allFormChangesPristine", allFormChangesPristine);
```
Sepehr-Sobhani, line 229: Leftover!

app/components/Form/ProjectSummaryReportFormSummary.tsx
```diff
@@ -24,7 +24,22 @@ const ProjectSummaryReportFormSummary: React.FC<Props> = ({
   viewOnly,
   isOnAmendmentsAndOtherRevisionsPage,
   setHasDiff,
+  query,
 }) => {
+  console.log("query", query);
```
Sepehr-Sobhani, line 29: Leftover!

schema/test/unit/functions/migration_rebuild_project_summary_report_history_test.sql
```diff
@@ -0,0 +1,126 @@
...
+
+-- create has no previous_form_change_id and id=1
+select cif.create_form_change('create', 'project_summary_report', 'cif', 'project_summary_report', '{}', 123, null, 'committed', '[]');
+
+-- all updates have the same previous_form_change_id=1 and form_data_record_id=123 (bug=form_data_record_id clears on commit)
+select cif.create_form_change('update', 'project_summary_report', 'cif', 'project_summary_report', '{"description": "value"}', 123, 1, 'pending', '[]');
+
+select cif.create_form_change('update', 'project_summary_report', 'cif', 'project_summary_report', '{"description": "most recent"}', 123, 1, 'pending', '[]');
+
+select cif.commit_project_revision(1);
```
Sepehr-Sobhani, line 41: Now that you've resolved the problem with the summary report handler, I'm wondering if we can be certain that the commit handler will clear the form_data_record_id upon commit. Do you think it would be better to simply pass null values for form_data_record_id instead?

Sepehr-Sobhani (changes requested):
I did notice a few leftovers and have one small question. Can you please help me with that?
Otherwise, everything seems to be in order 🤩

schema/test/unit/functions/migration_rebuild_project_summary_report_history_test.sql
BCerki: Great point, thanks for catching this!

Sepehr-Sobhani (approved):
Awesome 🤩

PR #1818, opened 2023-07-24
fix: 1787 calculated values total project and max performance amount

gurjmatharu:
#1787


app/cypress/e2e/cif/project-revision/index.cy.js
```diff
@@ -217,16 +217,13 @@ describe("the new project page", () => {
       "$1.00",
       "50.00 %",
       "$777.00",
-      "99.87%",
+      "This field cannot be calculated due to lack of information now.",
       /Jun(\.)? 10, 2020/,
       /Jun(\.)? 10, 2020/,
       "$1.00",
-      "$778.00",
+      "This field cannot be calculated due to lack of information now.",
```
JoshLarouche, line 224: I agree that we need to test that the fields are showing correctly when not enough information is entered, but is this removing confirming that the funding source is showing correctly when values are entered from the e2e tests?

gurjmatharu: The funding source calculations should handle this, and I've added more db unit tests for better coverage.

JoshLarouche (approved):

schema/deploy/computed_columns/form_change_total_project_value.sql
```diff
@@ -17,25 +13,37 @@ with additional_funding_sources as
...
         )::jsonb
       ) as x(source text, amount numeric, status text)
-    )
-  select
+    ),
+additional_funding_sources_sum as
+    (
+        select sum(amount::numeric) as total from additional_funding_sources where status = 'Approved'
+    ),
+additional_funding_sources_exist as
```
dleard, line 21: What do we want to show for `total_project_value` when an additional_funding_source status is `awaiting_approval` or `denied`?
Currently in both cases `total_project_value` will show the "cannot calculate.." message. Is this what we want?
Scenario:
An additional funding source is denied, until it is removed the `total_project_value` will never calculate. I think at least for `denied` it should return 0 as this may be data that we won't want to remove, but if we don't remove it, total_project_value is permanently unable to calculate.

There is also a bit of a behavioural inconsistency because of the `select 1 from cif.form_change...` block below on line 23. The calculation will work and consider `denied` and `awaiting_approval` as 0 when there is at least 1 approved source in the set, but will not calculate if the set only includes `denied` and `awaiting_approval` sources.

dleard (changes requested):

schema/deploy/computed_columns/form_change_total_project_value.sql
nanyangpro: As discussed with @gurjmatharu & @dleard:
- As long as there's one or more `awaiting_approval`, `total_project_value` cannot be calculated.
- As long as the rest are `approved`, when there are `denied`, `total_project_value` can be calculated.

dleard:
closed in favour of #1858

PR #1820, opened 2023-07-26
fix: 1788 show budgets diffs triggered by milestone changes

BCerki:
card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1788

Note: Anticipated Funding Per Fiscal Year is no longer in the happo shot. That's correct--the cypress test doesn't change anything that would create a diff in anticipated funding.

### DIFF
--- a/app/lib/theme/schemaFilteringHelpers.ts
+++ b/app/lib/theme/schemaFilteringHelpers.ts
@@ -1,4 +1,5 @@
 import type { JSONSchema7 } from "json-schema";
+import { isEqual } from "lodash";
 
 // The two functions below filter out fields from the formSchema that have not changed from the previous revision so the summary ignores these fields
 // This is mainly used when the form has multiple fields within it and we want to check each field data with the previous revision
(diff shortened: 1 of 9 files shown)


JoshLarouche (approved):

app/lib/theme/schemaFilteringHelpers.ts
```diff
@@ -22,14 +23,17 @@ export const getSchemaAndDataIncludingCalculatedValues = (
...
       oldFormDataIncludingCalculatedValues?.[key] === undefined
     ) {
       delete filteredSchema.properties[key];
     } else if (
-      formDataIncludingCalculatedValues?.[key] ===
-      oldFormDataIncludingCalculatedValues?.[key]
+      isEqual(
+        formDataIncludingCalculatedValues?.[key],
+        oldFormDataIncludingCalculatedValues?.[key]
+      )
```
Sepehr-Sobhani, line 36: No suggestions here, I'd like to know why we need to use the lodash here. Is it because we want to compare array objects?

Sepehr-Sobhani (approved):
Nice work! 🤩

app/lib/theme/schemaFilteringHelpers.ts
BCerki: Yes, the anticipated funding per fiscal year and additional funding are arrays. If we just use `===`, they're never equal and we get a diff even when nothing has changed

PR #1829, opened 2023-07-31
fix: 1767

gurjmatharu:


BCerki:
Finished in https://github.com/bcgov/cas-cif/pull/1864

PR #1831, opened 2023-08-01
Fix: 1744 show the Generate Reports button if we have an archived report

Sepehr-Sobhani:
[ZH CARD 1744](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1744)


BCerki (approved):
Nice, simple solution!

PR #1841, opened 2023-08-04
chore: 1772 emissions intensity e2e tests

Sepehr-Sobhani:
[ZH CARD 1772](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1772)

### DIFF
--- a/app/cypress/e2e/cif/project-revision/create-first-project-revision.cy.js
+++ b/app/cypress/e2e/cif/project-revision/create-first-project-revision.cy.js
@@ -250,36 +250,23 @@ describe("when creating a project, the project page", () => {
       .should("have.text", "Bob004 Loblaw004");
 
     // TEIMP section
-    cy.findByText(/^TEIMP start date/i)
-      .next()
-      .contains(/Jan(\.)? 1, 2022/)
-      .should("be.visible");
-    cy.findByText(/TEIMP end date/i)
-      .next()
-      .next()
-      .contains(/Duration: 1 month, 1 day/i);
-    cy.findByText(/^Functional Unit/i)
-      .next()
-      .should("have.text", "tCO");
-    cy.findByText(/^Production Functional Unit/i)
-      .next()
-      .should("have.text", "G");
-    // GHG performance uses the ReadOnlyAdjustableCalculatedValueWidget, which has a different HTML structure than the other fields
-    cy.findByText("GHG Emission Intensity Performance")
-      .next()
-      .should("have.text", "200.00%");
-    cy.findByText("GHG Emission Intensity Performance (Adjusted)")
-      .next()
-      .should("have.text", "100%");
-    cy.findByText("Payment Percentage of Performance Milestone Amount (%)")
-      .next()
-      .should("have.text", "100.00%");
-    cy.findByText("Maximum Performance Milestone Amount")
-      .next()
-      .should("have.text", "$12.00");
-    cy.findByText("Actual Performance Milestone Amount")
-      .next()
-      .should("have.text", "$12.00");
+    cy.checkEmissionIntensityReportSummaryForm(
+      /Jan(\.)? 1, 2022/,
+      /Feb(\.)? 2, 2022/,
+      "1 month, 1 day",
+      "tCO",
+      "G",
+      1.23,
+      2.34,
+      3.45,
+      "Not added",
+      "200.00%",
+      "100%",
+      null,
+      "100.00%",
+      "$12.00",
+      "$12.00"
+    );
 
     // summary
     cy.happoAndAxe("Project summary Form", "filled", "main", true);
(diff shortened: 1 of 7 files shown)


app/lib/theme/CustomDiffFields.tsx
```diff
@@ -11,7 +11,8 @@ const contentSuffixElement = (
 ): JSX.Element => {
   return (
     <span
-      id={id && `${id}-contentSuffix`}
+      // Add random number to id to avoid duplicate ids when multiple fields have the same contentSuffix
+      id={id && `${id}-contentSuffix-${Math.random().toFixed(2)}`}
```
BCerki, line 15: Even if fields have the same `contentSuffix`, shouldn't the `id`s be different? Are we accidentally repeating a field somewhere? I had this problem when both funding and emissions were displaying `holdbackPercentageToDate`, but that should be fixed now

app/cypress/e2e/cif/project-revision/create-first-project-revision.cy.js
```diff
@@ -122,13 +122,13 @@ describe("when creating a project, the project page", () => {
...
+    cy.addOrEditEmissionIntensityReport(
       "2022-01-01",
       "2022-02-02",
       "tCO",
-      "1",
-      "2",
-      "3",
+      "1.23",
+      "2.34",
+      "3.45",
```
BCerki, line 131: good call adding decimals

app/cypress/e2e/cif/project-revision/edit-project-revision.cy.js
```diff
@@ -151,22 +151,19 @@ describe("when editing a project, the project page", () => {
...
 
     // edit teimp
-
     cy.findByRole("button", { name: /Emissions Intensity Report/i }).click();
     cy.findByText(/Edit emissions intensity report/i).click();
-    cy.findByRole("button", {
-      name: /Add Emissions Intensity Report/i,
-    }).click();
-    cy.addEmissionIntensityReport(
+    cy.addOrEditEmissionIntensityReport(
```
BCerki, line 156: I think we additionally need a few assertions down by line 192ish to check these diffs like we're checking the other diffs

app/cypress/e2e/cif/project-revision/index.cy.js
```diff
@@ -227,7 +227,36 @@ describe("the new project page", () => {
...
+      .next()
+      .contains(/Jun(\.)? 10, 2020/);
+    cy.contains("Received Date")
+      .next()
+      .contains(/Jun(\.)? 10, 2020/);
+    cy.contains("General Comments")
+      .next()
+      .should("have.text", "TEIMP report comments 1");
+
+    cy.checkEmissionIntensityReportSummaryForm(
```
BCerki, line 243: Why do we need assertions in addition to the check function?

app/cypress/support/commands.js
```diff
@@ -349,16 +349,16 @@ Cypress.Commands.add(
...
     measurementPeriodEndDate,
     emissionFunctionalUnit,
     baselineEmissionIntensity,
     targetEmissionIntensity,
     postProjectEmissionIntensity,
-    productionFunctionalUnit = undefined,
-    adjustedEmissionsIntensityPerformance = undefined
+    productionFunctionalUnit,
+    adjustedEmissionsIntensityPerformance
   ) => {
```
BCerki, line 362: nit: I wonder about doing the arguments as an object? Then it wouldn't matter what order they were entered in. (I've entered them in the wrong order before and confused myself.) If you do make this change, I'd suggest only doing it for this new function--changing all the old ones would be a lot of work

BCerki (commented):
Nice work!

app/lib/theme/CustomDiffFields.tsx
Sepehr-Sobhani: Yes, In this scenario, we seem to be duplicating the content suffix (`Duration: X month`) on the same line when the TEIMP start and end dates are already filled in.

app/cypress/e2e/cif/project-revision/index.cy.js
Sepehr-Sobhani: I did not include checks for the fields specific to reporting requirements in the "checkEmissionIntensityReportSummaryForm" function. The reason behind this is that I intended to use this function again on the project summary view. As a result, I had to include additional checks in this section.

app/lib/theme/CustomDiffFields.tsx
Sepehr-Sobhani: Screenshot 2023-08-08 at 10 05 07 AM

BCerki: Gotcha. That looks like a bug to me--if the end date changes, then the duration would also change. Unless someone changed both the start and end dates and ended up with the exact same amount of time, I guess. Maybe we can put to remove the random number in the bug card?

BCerki (approved):
I think we need a bug card for duration, but the PR itself looks great!

app/lib/theme/CustomDiffFields.tsx
Sepehr-Sobhani: In my opinion, it would be better to display the duration on one side only and avoid duplicating it. This would eliminate the need for a random ID generator.

PR #1843, opened 2023-08-08
1703 hard delete attachments feature

BCerki:


Sepehr-Sobhani (changes requested):
Awesome 🤩 Good job! You just need to remove the sneaky `yarn-error.log` file :)

Sepehr-Sobhani (approved):

PR #1845, opened 2023-08-08
fix: 1759 multiple attachment diffs issue

JoshLarouche:
[zenhub 1759](https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-cif/1759)

### DIFF
--- a/app/components/Form/ProjectAttachmentsFormSummary.tsx
+++ b/app/components/Form/ProjectAttachmentsFormSummary.tsx
@@ -1,10 +1,10 @@
 import AttachmentTableRow from "components/Attachment/AttachmentTableRow";
 import Table from "components/Table";
 import { NoHeaderFilter, TextFilter } from "components/Table/Filters";
-import { useMemo } from "react";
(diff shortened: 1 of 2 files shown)


JoshLarouche:
Does a test need to be or should be added to ProjectAttachmentsFormSummary.test.tsx? I'm not sure.

BCerki (commented):
Looks great, bug is fixed! 
Yes, it would be good to add a test. You could do something like `Only displays the data fields that have changed` from `ProjectAnnualReportFormSummary.test.tsx`, but instead of checking for label names, just check the correct filenames show up/don't show up.

JoshLarouche:
> Looks great, bug is fixed! Yes, it would be good to add a test. You could do something like `Only displays the data fields that have changed` from `ProjectAnnualReportFormSummary.test.tsx`, but instead of checking for label names, just check the correct filenames show up/don't show up.

Test added!

PR #1846, opened 2023-08-09
1809 feat: filter milestone status

BCerki:
card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/board?repos=402862646


schema/data/dev/005_cif_reporting_requirement.sql
```diff
@@ -23,16 +26,16 @@ do $$
       update cif.form_change set new_form_data =
       json_build_object(
         'projectId', temp_row.form_data_record_id,
-        'reportDueDate', now(),
-        'submittedDate', now(),
+        'reportDueDate', end_of_day_today,
+        'submittedDate', end_of_day_today,
```
BCerki, line 30: In real data, the report due date has an end-of-day timestamp. If we use `now()` here, then any reports due today have `Late` status

app/pages/cif/projects.tsx
```diff
@@ -113,7 +113,12 @@ export function Projects({ preloadedQuery }: RelayProps<{}, projectsQuery>) {
         allProjectStatuses.edges.map((e) => e.node.name),
         { orderByPrefix: "PROJECT_STATUS_BY_PROJECT_STATUS_ID__NAME" }
       ),
-      new SortOnlyFilter("Milestone Due", "nextMilestoneDueDate"),
+      new SearchableDropdownFilter(
+        "Milestone Due",
+        "milestoneStatus",
+        ["Complete", "None", "On track", "Late"],
```
BCerki, line 119: I considered putting the statuses into a table like `project_status`, but the forms handle report status differently, so the table didn't seem very resuable

schema/data/dev/005_cif_reporting_requirement.sql
Sepehr-Sobhani: Using `end_of_day_today` was causing some end-to-end tests to fail. If we continue to use it, the value will always change and we won't be able to test it properly. Additionally, it was making happo unhappy :)

Sepehr-Sobhani (commented):
🥇

Sepehr-Sobhani (approved):

PR #1858, opened 2023-08-25
Fix/1787 2

dleard:


app/cypress/e2e/cif/project-revision/edit-project-revision.cy.js
```diff
@@ -85,7 +85,7 @@ describe("when editing a project, the project page", () => {
     cy.contains(/Budgets, Expenses & Payments/i).click();
     cy.findByText(/Edit budgets/i).click();
     cy.url().should("include", "/form/3");
-    cy.findByRole("button", { name: /submit/i }).should("not.exist");
+    // cy.findByRole("button", { name: /submit/i }).should("not.exist");
```
BCerki, line 88:
```suggestion
   
```

app/cypress/e2e/cif/project-revision/index.cy.js
```diff
@@ -217,46 +217,14 @@ describe("the new project page", () => {
...
-      124.35,
-      44.4224,
-      "66.41%",
-      "98%",
-      /Jun(\.)? 10, 2020/,
-      "100.00%",
-      "$0.10",
-      "$0.10"
-    );
+    //TODO: TEIMP Agreement, when fixture is added
```
BCerki, line 227: accidental deletion?

schema/deploy/computed_columns/form_change_total_project_value.sql
```diff
@@ -17,16 +13,40 @@ with additional_funding_sources as
...
+select
+  case
+    when
+      (
+        ($1.new_form_data ->> 'proponentCost')::numeric IS NULL
+        OR
+        ($1.new_form_data ->> 'maxFundingAmount')::numeric IS NULL
+        OR
+        (select awaiting_approval from check_awaiting_approval) = 1
+      ) THEN NULL
```
BCerki, line 40: lowercase SQL

schema/deploy/computed_columns/form_change_total_project_value.sql
```diff
@@ -17,16 +13,40 @@ with additional_funding_sources as
...
+    else
+      (
+        coalesce(($1.new_form_data ->> 'proponentCost')::numeric, 0)
+        +
+        coalesce(($1.new_form_data ->> 'maxFundingAmount')::numeric, 0)
+        +
+        coalesce((select total from additional_funding_sources_sum), 0)
```
BCerki, line 47: Do we still need to coalesce here, since we've already checked it isn't `null` around line 31ish?

BCerki (commented):
I'll finish this review Monday; just submitting now so github doesn't eat my changes over the weekend

schema/test/unit/computed_columns/form_change_holdback_amount_to_date_test.sql
```diff
@@ -46,44 +50,103 @@ values (
...
+    with record as (
+    select row(form_change.*)::cif.form_change
+    from cif.form_change where id=4
+    ) select cif.form_change_holdback_amount_to_date((select * from record))
+  ),
+  (
+  30000::numeric
+  ),
+  'Returns the correct value when adjustedHoldbackAmount or calculatedHoldbackAmount exists'
 );
```
BCerki, line 114: This seems to be almost the same test as the first one

schema/test/unit/computed_columns/form_change_holdback_amount_to_date_test.sql
```diff
@@ -46,44 +50,103 @@ values (
...
+    ) select cif.form_change_holdback_amount_to_date((select * from record))
+  ),
+  (
+    70000::numeric
+  ),
+  'Returns the correct sum of all milestones with expenses'
+);
+
+
+-- adjustedHoldbackAmount or calculatedHoldbackAmount are NULL
```
BCerki, line 72: I think this comment is in the wrong place

schema/test/unit/computed_columns/form_change_holdback_amount_to_date_test.sql
```diff
@@ -1,13 +1,17 @@
...
 
 insert into cif.project_revision(id, change_status)
   overriding system value
   values (1, 'pending');
 
+insert into cif.project_revision(id, change_status)
+  overriding system value
+  values (2, 'pending');
+
 insert into cif.form_change(
```
BCerki, line 15: suggest adding an archived value in here somewhere because filtering out archived is new in this test

schema/deploy/computed_columns/form_change_holdback_amount_to_date.sql
```diff
@@ -7,15 +7,26 @@ create or replace function cif.form_change_holdback_amount_to_date(parameter_fc
...
     where fc.project_revision_id = $1.project_revision_id
     and json_schema_name = 'milestone'
-    and (fc.new_form_data->>'hasExpenses')::boolean = true;
+    and (fc.new_form_data->>'hasExpenses')::boolean = true
+    and fc.operation <> 'archive'
+  )
+
+  select
+    case
+      when count(*) filter (where adjustedHoldbackAmount is null and calculatedHoldbackAmount is null) > 0 then null
```
BCerki, line 23: is `and` the correct condition here? If someone's filled out the adjusted holdback amount then I think they do have enough info to do the calculation

BCerki (commented):
Total project value is still calculated when I first add an additional funding source. Since status is the last field in that form, I'll have typed the number before the status, and it looks like the calculation doesn't update: https://www.loom.com/share/8767dd3aa64d4b6aa50220c3c0515c40

schema/test/unit/computed_columns/form_change_holdback_amount_to_date_test.sql
dleard: The test is similar, but the data being tested is different since one milestone is missing a calculated value and the other is missing an adjusted value

schema/deploy/computed_columns/form_change_holdback_amount_to_date.sql
dleard: This filter creates a condition where if any milestone is missing both values then the amount_to_date cannot be calculated

dleard:
> Total project value is still calculated when I first add an additional funding source. Since status is the last field in that form, I'll have typed the number before the status, and it looks like the calculation doesn't update: https://www.loom.com/share/8767dd3aa64d4b6aa50220c3c0515c40

This sounds like something to check with Nan on?

app/cypress/e2e/cif/project-revision/index.cy.js
BCerki: Should this TODO still be in here, vs in a card?

BCerki: Is the TODO meant to be in here, vs in a card?

BCerki (approved):
I asked Nan in Teams about the Loom video

PR #1860, opened 2023-08-31
Fix milestone commit handler updating archived records

Sepehr-Sobhani:


dleard (approved):
nice! Just needs a rebase

PR #1882, opened 2023-11-09
Chore/1876 rebase pending revisions

mikevespi:
Adding the backend work for handling concurrent revisions.


pbastia:
Needs rebase!

schema/test/unit/mutations/commit_form_change_internal_test.sql
```diff
@@ -77,7 +77,382 @@ select is(
...
+select cif.add_project_attachment_to_revision(3,2);
+
+select cif.commit_project_revision(3);
+
+-- emission_intensity
+-- project_contact
+-- project_manager  projectManagerLabelId 1 is update and 2 is created in pending
+-- reporting_requirement should be 3 total with 1,2,3 for reportingRequirementIndex
+-- milestone should be 3 total with 1,2,3 for reportingRequirementIndex
+-- attachment
```
pbastia, line 455: It looks like we're missing some assertions?

schema/test/unit/mutations/commit_form_change_test.sql
```diff
@@ -68,7 +68,7 @@ select results_eq(
       null::int,
       'pending'::varchar,
       'reporting_requirement'::varchar,
-      '["hazErrors"]'::jsonb
+      '[]'::jsonb
```
pbastia, line 71: any reason why we removed the errors?

pbastia (changes requested):
Great work!
There's a few things we need to adjust I think, but the logic is sound and it seems like it'll work. Also it looks like we're missing some tests at the end?

A note about postgres `count()` function, not that it matters in this case since we have at most a few rows to count, but `count(expression)` is less performant than `count(*)` since the postgres query planner recognizes this as "count everything without evaluating the expression every row"

schema/deploy/mutations/commit_form_change.sql
```diff
@@ -12,7 +12,12 @@ begin
     validation_errors = coalesce(form_change_patch.validation_errors, validation_errors)
   where id=row_id;
 
-  return (select cif_private.commit_form_change_internal((select row(form_change.*)::cif.form_change from cif.form_change where id = row_id)));
+  return (select cif_private.commit_form_change_internal(
+    (select row(form_change.*)::cif.form_change from cif.form_change where id = row_id),
+    (select id  from cif.project_revision
+      where project_id=(select project_id from cif.project_revision where id = form_change_patch.project_revision_id)
+      and change_status = 'pending' and id != form_change_patch.project_revision_id limit 1)
```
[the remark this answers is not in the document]
mikevespi: [These constraints](https://github.com/bcgov/cas-cif/blob/develop/schema/deploy/tables/project_revision_006_unique_pending_per_revision_type.sql) should prevent multiple from being possible, but what would we do when we log? Prevent the revision from being commit?

schema/deploy/functions/jsonb_minus.sql
```diff
@@ -0,0 +1,31 @@
...
+-- A note on the functionality:
+-- If a key is present in the subtrahend but not the minuend, it will not appear in the result set.
+-- {"a": 1, "b": 3} - {"a": 1, "b": 2, "c": 3} =  {"b": 3}
+-- If however a key is present in the minuend but not the subtrahend, it will appear in the result set with its value.
+-- {"a": 1, "b": 3, "c": 3} - {"a": 1, "b": 2} =  {"b": 3, "c": 3}
+
+-- This behaviour fits our needs at the time of writing this, so the additional complexity of handling the other cases is not needed.
+
+
+create or replace function cif.jsonb_minus(minuend jsonb, subtrahend jsonb)
```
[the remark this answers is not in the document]
mikevespi: It would! This is cool I didn't know that you could add them like this

schema/deploy/mutations/commit_form_change_internal.sql
```diff
@@ -1,10 +1,17 @@
 -- Deploy cif:mutations/commit_form_change to pg
 begin;
 
-create or replace function cif_private.commit_form_change_internal(fc cif.form_change)
+-- We need to explicitly drop the old function here since we're changing the signature.
```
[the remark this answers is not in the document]
mikevespi: Done!

schema/deploy/mutations/commit_form_change.sql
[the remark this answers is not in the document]
pbastia: Since we have those constraints I don't think we even need to say `limit 1` then. Maybe just a comment that we have unique indices would be good?

schema/deploy/mutations/commit_form_change_internal.sql
```diff
@@ -26,12 +38,238 @@ begin
...
+
+      else
+/*
+  If pending has not made changes to the form data, the pending record can be deleted as it never would have been made
+*/
+        delete from cif.form_change
+        where project_revision_id = pending_project_revision_id
+          and form_data_table_name = fc.form_data_table_name
+          and form_data_record_id = fc.form_data_record_id;
+      end if;
```
pbastia, line 264: 👌 👌

schema/test/unit/concurrent_revisions/conflicting_creates_test.sql
```diff
@@ -0,0 +1,335 @@
...
+  (select count(*) from cif.form_change where project_revision_id = 2 and json_schema_name = 'funding_parameter_EP'),
+  1::bigint,
+  'When committing and pending have both created a funding parameter form, only one exists in pending after the first is commit'
+);
+
+select is (
+  (select (new_form_data ->> 'provinceSharePercentage')::int from cif.form_change where project_revision_id = 2 and json_schema_name = 'funding_parameter_EP'),
+  1,
+  'When committing and pending have both created a funding parameter form, pending maintains its form values'
+);
```
pbastia, line 219: This is a tricky one. If a value is not set by the amendment in the funding parameter form, but set in the general revision, would a user expect that to be reflected in the amendment?
This is probably a rare edge case where there might be the impression that data was lost. This can probably be resolved with some training.

pbastia (approved):
Fantastic job with the testing especially!! :shipit:

schema/test/unit/concurrent_revisions/conflicting_creates_test.sql
mikevespi: It's a very good question. I leaned towards "the amendment is right" when I wasn't sure but I'd love to know where Business Area sit on the specific cases.

pbastia: Probably careful examination of the changes before submitting an amendment or a revision would do the trick

Issue #1333, opened 2023-01-11
Tech Debt: Remove dependency on Terraform Cloud and cas-shelf


pbastia:
Following a call with Junmin, here's what we need to do:

- use Terraform (not cloud) in the bucket provisioning job 
- store the terraform state in GCS

Refactoring opportunities will pop up from there:
- the bucket provisioning job should live in its own helm chart so it can be reused across projects

AC: 
- [ ] Terraform state for CIFF created in GCS bucket
- [ ] Relavent Terraform modules in CIFF Helm charts with Terraform job
- [ ] Existing resources in Terraform Cloud state _moved_ to GCS bucket state

pbastia:
@dleard @Sepehr-Sobhani @sam-warren :eyes:

pbastia:
@mikevespi has experience with this :)

Sepehr-Sobhani:
I'm interested to join if/when someone wants to tackle this :)

joshgamache:
**Research**:	
- `cas-cif/chart/cas-cif/templates/bucket-provisioning-job.yaml` references `TFC`, which are variables for use with Terraform Cloud
   - This file provisions the GCP Storage containers, based on configuration from another repo: `cas-shelf/bcgov/`. That repo's folder contains terraform files, get applied from TFCloud.
   - This [Dockerfile](https://github.com/bcgov/cas-shelf/blob/master/docker/add-app/Dockerfile) is what gets run in the above helm template
	- Pulls in `cas-shelf/tfe-scripts/helpers` and `cas-shelf/tfe-scripts/tf-add-app-run.sh`
	- [tf-add-app-run.sh](https://github.com/bcgov/cas-shelf/blob/master/tfe-scripts/tf-add-app-run.sh) is the script that runs in the container
   - In `cas-shelf`, these commands might be needed [Package cas-shelf/cas-shelf-tfe-add-app (github.com)](https://github.com/bcgov/cas-shelf/pkgs/container/cas-shelf%2Fcas-shelf-tfe-add-app#other-commands) to recover existing TF Cloud state?
- `c53ff1` is the namespace prefix for CIF 

**Goals**:
- Establish Remote State in GCP buckets (ie. use our OpenShift's secret with the tfbackend, but locally).
- Configure Terraform scripts that will be run, *but do not run them*
- **Back up existing state!**
[...]

PR #1887, opened 2024-01-26
Chore/1333 replace terraform cloud

joshgamache:
Addresses #1333 - [Project link](https://github.com/orgs/bcgov/projects/123/views/2?filterQuery=1333&pane=issue&itemId=49863829)
Adds Terraform scripts and Helm charts to run them. Resource states migrated from Terraform Cloud to a Google Cloud Storage bucket.

## Changes 🚧

- Added Terraform module, imported from cas-shelf
- Added Helm templates to handle:
  - Terraform module loading to configMaps
  - OpenShift service account for use by Terraform
  - Terraform apply job

## To test 🔬

- The Helm charts are built into the automation around deploys via ShipIt. Once a deploy has been run, you can check the job `Terraform-apply` in OpenShift to see if the run was successful.

## Notes📝 
- 🔒 Terraform Cloud state has been locked to enable migration of state to state buckets
- The script has been run and resources have been transitioned from Terraform Cloud to GCP buckets holding state for each namespace (cif-dev, cif-test, cif-prod).  
- Resources have been migrated rather than recreated, only the state "owning" the resource changes. 
- The resources `kubernetes_secret.secret_sa` have had Terraform changes, but only the addition of a default value (consistent with our use of them) due to a new provider version. 
- The script used for migration (/chart/cas-cif/terraform/tf-migration.sh) is kept here for history. It will be removed by https://github.com/bcgov/cas-pipeline/issues/83


chart/cas-cif/templates/jobs/terraform-service-account.yaml
pbastia, line 1: We typically use only one file per resource in our helm chart, this makes it more readable especially for people who don't know how things are setup. Unless there is a reason for having them together?

chart/cas-cif/templates/jobs/terraform-apply.yaml
```diff
@@ -0,0 +1,73 @@
...
+              value: "https://api.silver.devops.gov.bc.ca:6443"
+            - name: GOOGLE_APPLICATION_CREDENTIALS
+              value: "/etc/gcp/credentials.json"
+          # Terraform was having an issue pulling kubernetes_host in as a TF_VAR, so we add it as a attribute to the command
+          command:
+            [
+              "/bin/sh",
+              "-c",
+              "cp -r /terraform/. /working && cd working && export TF_VAR_kubernetes_token=$( cat /var/run/secrets/kubernetes.io/serviceaccount/token ) && terraform init -backend-config=/etc/tf/gcs.tfbackend && terraform apply -var=\"kubernetes_host=$kubernetes_host\" -auto-approve",
+            ]
```
pbastia, line 54: We might want to add a `set -euxo pipefail` to help debugging if the command fails
Helm sucks, and there's a slightly easier syntax to pass in bash scripts:

```suggestion
          command:
              - /bin/sh
              - -c
              - |
                  set -euo pipefail;
                  cp -r /terraform/. /working;
                  cd working;
                  export TF_VAR_kubernetes_token=$( cat /var/run/secrets/kubernetes.io/serviceaccount/token );
                  terraform init -backend-config=/etc/tf/gcs.tfbackend;
                  terraform apply -var=\"kubernetes_host=$kubernetes_host\" -auto-approve";
```

chart/cas-cif/templates/jobs/terraform-apply.yaml
```diff
@@ -0,0 +1,73 @@
...
+          imagePullPolicy: {{ .Values.devops.image.pullPolicy }}
+          volumeMounts:
+            - mountPath: /etc/gcp
+              name: service-account-credentials-volume
+              readOnly: True
+            - mountPath: /etc/tf
+              name: terraform-backend-config-volume
+              readOnly: True
+            - name: tf-cache
+              mountPath: /working
```
pbastia, line 29: Because of `emptyDir` the working directory will be deleted as soon as the job pod completes. Should we name this mount `tf-working-dir` or something like that? Since it doesn't cache anything

pbastia (changes requested):
Nice work! A few comments about bash formatting, looks good otherwise.
I'd suggest also:
- moving `tf-migration.sh` and `migration.example.tfvars` out of the helm chart, I don't think there is a to have these in there?
- move all the terraform yaml templates to `chart/cas-cif/templates/terraform` instead of the `jobs` folder

chart/cas-cif/templates/jobs/terraform-service-account.yaml
joshgamache: Since they were all directly related, I stitched them together for convenience. But there's no reason they need to be that way. I've addressed that in 8a55f15b.

chart/cas-cif/templates/jobs/terraform-apply.yaml
joshgamache: Thanks for the suggestion. I knew the long syntax wasn't the easiest to read, this makes it _much_ easier to understand! Changed with d068385a.

joshgamache:
> * moving `tf-migration.sh` and `migration.example.tfvars` out of the helm chart, I don't think there is a to have these in there?

I cleaned up the directories with the suggestions you made. Especially with the service account related file split, it keeps them consistently together. 

I removed the migration script entirely, as it is unneeded now that the migrations have been run.

pbastia (approved):
Slick!

mikevespi (approved):
Awesome stuff! Leaving it Approved but not merged as per discussion today about merging with eyes on. Aside from that let's make sure we get a tech-debt card to create custom GCP role instead of their generic admin, they over permit.

Great stuff!

chart/cas-cif/.helmignore
```diff
@@ -21,3 +21,11 @@
 .idea/
 *.tmproj
 .vscode/
+
```
mikevespi, line 24: Can we add the below to the .gitignore as well :)

joshgamache: Added in 81e56976e87a31cf74ca2dd3c1d6d820efd2295e

PR #1891, opened 2024-02-15
Chore: fix terraform helm hooks

pbastia:


chart/cas-cif/templates/terraform/terraform-apply.yaml
```diff
@@ -4,9 +4,10 @@ metadata:
   name: terraform-apply
   labels:
     component: backend
-  namespace: {{ .Release.Namespace }}
```
pbastia, line 7: These are automatically added by helm, we don't need them in the template

joshgamache (approved):
Fantastic! Great to see we replaced ~300 lines with 8!

pbastia:
> Fantastic! Great to see we replaced ~300 lines with 8!

all the credit is yours!

PR #1908, opened 2024-05-07
chore: update zaproxy action version and image source

mikevespi:
The image being used for Zaproxy can't be found anymore so we've changed the image source and bumped the action version.


joshgamache (approved):
Two lines both check out. Version exists and the ghcr has that image as expected.

PR #1910, opened 2024-05-10
Feat/clearer diffs

mikevespi:
This PR updates the way the custom diff fields represent the changes. The fields will now show the latest committed state (if any), and what will be changed/added/removed in the committing project revision. The intent is that it will be easier for users to discern what they will be changing about a project when applying a revision by avoiding showing them any intermittent states the project may have been in between the revision being created, and now.


pbastia (approved):
Slick!! This is a bunch of fantastic work.
Just nitpicking about a couple of comments, we don't have to fix them

Issue #1917, opened 2024-05-28
CIF: Add 2024 to "RFP Year" Drop-down List


patriciarussellCAS:
#### Describe the task

The CIF team will soon be filing new 2024 applications so need to have an option to select 2024 from the "RFP YEAR" drop-down list. 

#### Acceptance Criteria

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

patriciarussellCAS:
Thanks @pbastia I'm good to remove the backlog refinement tag.

PR #1918, opened 2024-05-28
Add 2024 to the funding stream rfp table

BCerki:
card: https://github.com/bcgov/cas-cif/issues/1917#issuecomment-2135841962

This PR:
- add 2024 via sqitch
- changes the e2e test for creating a project to use 2024 instead of 2020

### DIFF
--- a/app/cypress/e2e/cif/project-revision/undo-changes.cy.js
+++ b/app/cypress/e2e/cif/project-revision/undo-changes.cy.js
@@ -92,7 +92,7 @@ describe("when undoing, the project revision page", () => {
     }).click();
     cy.addOrEditEmissionIntensityReport(
       "2022-01-01",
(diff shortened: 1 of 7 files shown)


app/cypress/e2e/cif/project-revision/undo-changes.cy.js
```diff
@@ -92,7 +92,7 @@ describe("when undoing, the project revision page", () => {
     }).click();
     cy.addOrEditEmissionIntensityReport(
       "2022-01-01",
-      "2022-02-02",
+      "2022-04-02",
```
BCerki, line 95: The datepicker was ending up with March instead of February both locally and in CI, so I tried a different date and that seemed to make it happier

pbastia (approved):
💪 🏅

PR #1919, opened 2024-05-31
fix: updating sentry packages

pbastia:


app/sentry.client.config.js
```diff
@@ -16,4 +16,5 @@ Sentry.init({
   dsn: SENTRY_DSN,
   environment: SENTRY_ENVIRONMENT,
   release: SENTRY_RELEASE,
+  tracesSampleRate: 1.0,
```
Sepehr-Sobhani, line 19: This will send all transactions to the sentry and will exceed our sentry quota in a day. I had that experience in Registration sentry :) Suggest changing this to `0.1`

Sepehr-Sobhani (approved):
🥇

